Issue #51, opened 2015-01-15
Debug engine for Cosmos use not unique GUID


kant2002:
GUID which is used for the AD7Engine class if shared with Python Tools for VS and other various samples across internet.
See:
https://github.com/dtretyakov/node-tools/blob/master/src/DebugEngine/Engine/AD7Engine.cs#L29
https://github.com/LiXizhi/NPL/blob/master/Microsoft.VisualStudio.Debugger.SampleEngineWorker/SampleEngine.rgs#L17

On my PC personally It is shared with PythonTools for VS.

That have to be changed, since we Cosmos could not be used side-by-side with Python Tools

mterwoord:
Good catch. Be verify careful though, with things like our setup and stuff. Also, try to figure out what happens with current devkit/userkit users, ie upgrading should remove old registrations..

kant2002:
Not sure what to do with existing users.I assume that nobody affected since nobody complains.
Even if we have some users affected, I don't see any place in the existing installer which handle upgrade path, just plain install/uninstall.

mterwoord:
Are you able to make a check in the setup, that removes the old keys/values when the python tools are not installed?

PR #53, opened 2015-01-15
Changed GUID for debug engine

kant2002:
Current GUID was conflicting with existing tools and a lot of examples on the Internet
Relates to #51


mterwoord:
Can you create inno entries for removing the previously created entries?

kant2002:
Will try.

PR #105, opened 2015-05-17
Geramys changes

Geramy:
Added Bochs support.


Geramy:
Finished adding bochs support, now we can use bochs as a vm and a debugger.

PR #124, opened 2015-05-29
Documentation update

ARMmaster17:
Grammer fixes. Reordering and rewording confusing phrases.


ARMmaster17:
I'm not sure about cbbed3f0c7b6d5e2614410158f275f41663252e7, I deleted it because the objective of the article wasn't clear.

ARMmaster17:
It's also pointless now that Visual Studio 2013 Community is out and free.

mterwoord:
il2cpu: 
- please highlight that, although currently only x86 is supported, we have other platforms planned in the future.

Devkit: 
- Please remove all occurrences of VS 2010, and replace with 2013
- Please say the user needs to clone our repository. redirecting to a tutoral on how to do so with github (there's github docs for that)
- You still need the vs sdk (but the 2013 one)

Memory manager:
- compare/exchange is done using System.Threading.Interlocked. No manual assembler code needed. That's handled by the plugs for System.Threading.Interlocked.
- when using parenthesis (like this), add a space before the opening: "bla (bla" instead of "bla(bla" 

Plugs:
- We wouldn't be emulating the WINE api, but the windows api

Express: please keep it for now..

While community edition is free, not everyone can use it..

ARMmaster17:
Ok, I added some updates. It should be good to go now.

mterwoord:
Userkit:
- Docs/Express/Compile still referneces VS 2010 instead of VS 2013..

Memory manager:
- "Allocating a block happens by finding the first free block and split it (if necessary).

Plugs: 
- WINE API is still referenced. Doesn't make sense to do so..

ARMmaster17:
Ok, should be fixed now. I could fix /Doc/Express/Compile.md if I could figure out what it was trying to achieve.

mterwoord:
The process for express compilation is actually explained in the doc ok. Merging now.

PR #133, opened 2015-06-17
Rewrote Keyboard class mostly from scratch

zdimension:
Now, the num lock, caps lock etc works. I have also changed the
ConsoleImpl class to make that the no-char keys (like Esc, Win, F1-10)
don't print blank spaces.


mterwoord:
Instead of modifying DefaultKeyboard. Please make a new keyboard class, which inherits the base class, and which does your own magic...

PR #141, opened 2015-06-20
Fixed some empty-char-thingy bugs in the keyboard implementation and …

zdimension:
…now if you press a no-char key (like Esc, F1-12, PrtScr), it does nothing instead of printing a blank space.

### DIFF
--- a/source/Cosmos.HAL/DefaultKeyboard.cs
+++ b/source/Cosmos.HAL/DefaultKeyboard.cs
@@ -7,7 +7,7 @@ namespace Cosmos.HAL
 {
     //public delegate void HandleKeyboardDelegate(byte aScanCode, bool aReleased);
 
-    public class DefaultKeyboard: Keyboard
+    public class DefaultKeyboard : Keyboard
     {
 
         protected override void Initialize()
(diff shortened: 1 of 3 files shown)


mterwoord:
- Please don't include ignore file-level ignore rules for resharper.
- What do you think, The Numlock/ScrolLock/Capslock properties should be in the baseclass I think..
- In HandleScanCode, you don't check the result of the GetKey method. Please do so.
- I think, whenever a key doesn't have a character (like the F1 key), it shuoldn't be registered with char \0 (which is actually a char)...

source/Cosmos.HAL/DefaultKeyboard.cs
```diff
@@ -260,28 +232,25 @@ public void ChangeKeyMap(List<KeyMapping> aKeys)
...
-                        else key = map.Value;
-
-                        aValue = key;
-                        return true;
-                    }
-                    break;
+                    var map = t;
+                    var key = '\0';
+
+                    if (ShiftPressed && CapsLock) key = map.ShiftCaps;
```
mterwoord: Please always add { and } for codeblocks..

mterwoord:
Besides the line note, everything looks ok.

PR #149, opened 2015-07-02
VGAScreen fix

CSharpLover:
VGAScreen is now working.
Note: You must use Cosmos.System.VGAScreen class, instead of Cosmos.HAL.VGAScreen class.


mterwoord:
Can you please include a default: label in the switch statements? (Ie, have it throw an exception that the situation isn't implemented)

CSharpLover:
"default:" labels added in the switch statements.

Issue #143, opened 2015-06-23
Make a simple UI for our automated kernel tester


mterwoord:
.

mterwoord:
Also make a better console version..

PR #161, opened 2015-07-23
#143 has been finished ui for the test runner is finished

Geramy:
I have finished making the test runner ui hope you like it! :)


mterwoord:
Well, We're getting there..

- I'd like to not see all messages by default (See the difference between the OutputHandlerConsole and OutputHandlerFullConsole). After a click, you'd see the full details
- Make some kind of counter where it lists how many succeeded and how many didn't.
- Somehow make the list autoscroll, so you see the last message without problem.

Geramy:
Okay next ill do a log level selection, then I will add a counter and make
it auto scroll! :)

mterwoord:
Also, if you close the window, it should just stop running, instead of continuing..

mterwoord:
And bonus points if you could add the same kind of nesting the console does. Please try to put more context visible in general (like the run configurations,etc)

Geramy:
Okay so nest by level and yah I just have ro exit the thread in close

PR #163, opened 2015-07-24
Completed the VBE driver

MyvarHD:
all that remains now is nice to have features but this can come at later stage.


MyvarHD:
This commit includes the getpixel

PR #177, opened 2015-08-01
Use a more Github-looking badge for Gitter

zdimension:


mterwoord:
What's wrong with the official one?
I mean, I like the current one more actually..

zdimension:
See [https://github.com/gitterHQ/gitter/issues/636](this)
I mean as we use shields.io for the release badge, why not use it for gitter so the two use the same style... and the official one ~~is ugly~~ *doesn't have those beautiful gradients*.

mterwoord:
Now i see what you mean.

PR #179, opened 2015-08-05
Remove VS 2013 in files as it's not supported anymore

zdimension:


mterwoord:
Code doesn't compile.

PR #186, opened 2015-08-12
Added a Power class containing a WORKING reboot function

zdimension:


zdimension:
Done.

mterwoord:
I think interrupts need to be disabled?

zdimension:
Well, it works perfectly without disabling interrupts...

mterwoord:
OK, but then if it somehow doesn't reboot, the OS will just continue running (reacting to interrupts). Can you change this?

PR #193, opened 2015-08-16
Remove Cosmos.VS.Wizards Project

FreddyD-GH:
Removing the Cosmos.VS.Wizards project from the main solution and deleting directory since it is no longer used.


mterwoord:
Sorry, I remember again: We use the wizard to create 2 projects with 1 template entry ("Cosmos Project").

PR #196, opened 2015-08-19
Plug attribute

sgetaz:
New plug attribute, which causes an exception during compilation if method with attribute is not plugged


mterwoord:
@sgetaz Could you update this pull request please?

lukaandjelkovic:
Why is this still open?

mterwoord:
@lukaandjelkovic Because this PR needs updating, and testing.

PR #204, opened 2015-08-31
Add TestRunner Docs

ImaBrokeDude:
Super Basic TestRunner Docs

-ToDo Add more info on config


mterwoord:
Can you add a little pointer to the configuration file and where the projects are located?

PR #207, opened 2015-09-09
StringFunctions

Cyber4:
dStartsWith(input, comparator)
returns a boolean value.
dContains(input, comparator)
returns a boolean value
dEndsWith(input, comparator)
returns a boolean value


mterwoord:
Please fix the existing plugs rather than make new methods..

PR #226, opened 2015-10-25
Moved from Text Screen.cs to here

lukaandjelkovic:
80% of this is wrong, please repair


mterwoord:
You didn't actually put the members in a different class, but you moved the class to a different file. Please move the members into the VGA class.

mterwoord:
Please merge this PR and PR #227

PR #229, opened 2015-10-29
This should be better

lukaandjelkovic:


mterwoord:
Why are you changing the code this way?

mterwoord:
For simple stuff like tests, I like code duplicated like the original.

PR #233, opened 2015-11-03
nosound() is now noSound()

lukaandjelkovic:
just a uppercase letter.


mterwoord:
Please use normal uppercase camelcasing for identifiers. Thanks.

PR #252, opened 2015-11-12
FAT Work (New stuff)

Cyber4:
Added ReadAllBytes, WriteAllBytes, and Copy to the FileImpl plug.


mterwoord:
Didn't do any testing at all yet, but can you dispose both streams in Copy?

Also, when doing a full stream copy, I think it's better to use a smaller buffer than the complete file (ie, copy in chunks)

Cyber4:
By doing a lot of testing on the yes runner, it turns out that the amount of data being written has a lot less of an impact on speed than opening streams, so by going piece by piece copy would be around 2x slower.

Also, if the read and write spread put in place if it disposes it should do it copy too.  If not I can add it

mterwoord:
You wouldn't be reopening streams all the time..

Can you have File.Copy use Stream.CopyTo for the stream handling?

Cyber4:
I think I recall that not working, but I'll have to doublecheck.  I think we should use this for now and worry about optimizing it later.  Slow copy is much better than no copy.  Also readallbytes and writeallbytes is important too.

Cyber4:
UPDATE:

Here are the results of the benchmark testing between the copy used in this pr versus Stream.CopyTo:

Benchmark 1:
The results were tested on Bochs emulator with remote debugging off.
Current Copy:  Time = 7.88 seconds, successful copy.
File.CopyTo: Time = 18.89 seconds, failed copy.

Benchmark 2:
The results were tested on VMware with remote debugging off.
Current Copy: 3 miliseconds, success
Copy2: 4 miliseconds, failure.

Cyber4:
Hm?  Everything except the FAT related stuff has been removed, Im just bad at prs so it shows up there

PR #256, opened 2015-11-14
chg: fix pre requisit for setup program

rebizu:
Fixed the check for inno setup.


charlesbetros:
Why change is null or whitespace? It calls is null or empty.

rebizu:
hey @charlesbetros I was testing and forgot to change it back. 

i will change it and resubmit a pull request

PR #259, opened 2015-11-17
fixing #254

rebizu:
fixing #254


lukaandjelkovic:
ok, lets try does this one works

PR #283, opened 2015-12-07
Implemented Console's CursorSize and CursorVisible properties …

fanoI:
- Console's CursorSize and CursorVisible properties are implemented
- Console Cursor default size is 25 as on Windows (before was 1)
- Console Cursor is visible by default


charlesbetros:
I think "throw new NotimplementedException()" works. Instead of "new Exception("Not implemented.")".
Can you create a debug helper like https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.System/FileSystem/FileSystemHelpers.cs instead for your debug output? That way we can enable or disable it with a define.

fanoI:
Yes I could surely change "new Exception("Not implemented.") to the default existing one.
Regarding the Debugger class I tought that it was already under a conditional compilation is instead trying to write a not existing pipe slowing down the system? Ops!
So I need to create a new ConsoleHelper.cs class file near the Console.cs file is and using [Conditional("COSMOSDEBUG")] to decorate my methods but then what happens if COSMOSDEBUG (in release for example) is not defined and I call Debug.xxx?

mterwoord:
If the conditional is not defined, the calls won't be included, which is the whole point. At some point we'll add runtime switches for that, but for now, this is ok.

fanoI:
Fantastic! I'm a little newbie with C# so it is surprising how it seems to be so well tough (in C I had to use horrendous macro to obtain the same result).
I'll proceed with this modifications in the week end then...

PR #287, opened 2015-12-12
Updating the .editorconfig

Redall:
Updating *.cs > indent_size to 5
Removed some pointless comments

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,13 +1,10 @@
-# EditorConfig is awesome:
-http://EditorConfig.org
-
 # top-most EditorConfig file
 root = true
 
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -15,7 +12,7 @@ indent_size = 2
 trim_trailing_whitespace = true
 
 [*.cs]
-indent_size = 4
+indent_size = 5
```
charlesbetros: Why 5?

Redall: I triend tabs and i can get one lite more space. But if you don't want, you can, it's your project, not mine.

PR #290, opened 2015-12-12
Updated the install-VS2015.bat file

Redall:
> Added two messages :
   * For the build completed successfuly
   * For the Cosmos Builder Tool does not exists


mterwoord:
If I read the changes correctly, the batch file now also pauses when everything goes ok. This is not desirable... Can you please remove that?

Redall:
@mterwoord Ok, i will fix that

PR #354, opened 2016-04-09
Update Install-VS2015.bat

nithinswarrier:
Deleted del command to stop error file not found and set %ProgramFiles% if Windows is 32bit for msbuild directory

### DIFF
--- a/install-VS2015.bat
+++ b/install-VS2015.bat
@@ -1,8 +1,10 @@
 @REM This file is used to build the Cosmos sources.
 
-cls
-del source\Cosmos.Build.Builder\bin\Debug\Cosmos.Build.Builder.*
 
-"%ProgramFiles(x86)%\MSBuild\14.0\Bin\msbuild.exe" "source\Builder.sln" /nologo /maxcpucount /p:Configuration=Debug /p:Platform=x86
+cls
+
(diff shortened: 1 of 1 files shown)


mterwoord:
The delete command is intentional: It makes sure an old version is not getting installed.

I'm pretty sure there's something else going wrong. Are you on 32-bit windows?

nithinswarrier:
Yes, I'm using Windows 10 32 bit. I think there is no need for del command as msbuild replaces files present in the directory.

mterwoord:
MSBuild doesn't remove the old version if it didnt succeed, which is what the del command is for. 

Someone could fix it though by making things work both on 32-bit and 64-bit windows.....

nithinswarrier:
Yes it works on 64 and 32 bit if msbuild is installed

charlesbetros:
Why not use something like
```reg.exe query "HKLM\SOFTWARE\Microsoft\MSBuild\ToolsVersions\14.0" /v MSBuildToolsPath```

mterwoord:
@nithinswarrier So if you change how MSBuild.exe is found, it works, even with the delete command?

mterwoord:
@nithinswarrier Can you verify with latest master? I think we got this fixed now.

mterwoord:
This is merged now.

PR #362, opened 2016-05-06
add some Kernal test for Collections (Dictionary and Hashtable)

zarlo:


mterwoord:
Great work!

Could you add tests for testing with a custom comparer? Also, please rename all occurrences of Kernal to Kernel (ie, a -> e)

PR #383, opened 2016-06-17
add some kernel test for Collections (Dictionary and Hashtable) 45% Done

zarlo:


mterwoord:
Can you change all occurrences of Kernal to Kernel? (ie, a to e)

PR #390, opened 2016-06-19
Replace PXE interface textbox with a combobox, filled with local machine interfaces

### LINKED ISSUE #92: Dropdown box of interfaces on settings page
When the PXE profile is selected, there should be a drop-down box instead of a textbox that contains a list of interfaces on the machine.

jp2masa:
resolves #92


mterwoord:
Why close it?

jp2masa:
It has a problem. I reopen it today, I'll fix it.

PR #393, opened 2016-06-20
Fix String.EndsWith and add reference to PlugsInspector project

jp2masa:


jp2masa:
Now also updates PlugsInspector so it lists Cosmos.Core.Plugs plugs too.

PR #394, opened 2016-06-21
Adds plug for System.IO.File.Delete(), and adds deletion capabilities to FAT.

MichaelTheShifter:
Only thing wrong is it's wrongly looking for a directory. Still have to fix that.


jp2masa:
Change list from my commit:
- CosmosVFS.GetDirectoryListing(DirectoryEntry): "Argument is null or empty" error - Fixed

- File.Delete and Directory.Delete implementations - Fully Working

- Change Guess demo references to Project References - Done

jp2masa:
The test is failing for simple reasons I'll fix them in the next 12 hours I think

fanoI:
What do you mean by "Only thing wrong is it's wrongly looking for a directory"? You cannot delete directories or it pick the wrong path?

jp2masa:
I worked on FAT, with @MichaelTheShifter work as base and I think he meant that File.Delete, used with directories, would delete them instead of throwing an exception.

fanoI:
@MichaelTheShifter  / @jp2masa I'd say this is yet not ready to be merged could you please find a way to fix the last problem so to have a proper deletion? Thank you!

jp2masa:
There are tests for the deletion and they are passing in this PR, what do you think is missing, so I can fix it?

fanoI:
@charlesbetros said that while the file was apparently deleted correctly the entry was not properly cancelled from the FAT table so the space was not really freed but better to ask to him for details / help.

jp2masa:
I added a method to Clear a FAT entry(set it to 0), and when a file is deleted, it's FAT entries are cleared. I'll test manually to see if space is freed or not.

charlesbetros:
There should be a method in place already to set an entries value. In addition 0 is not the correct value.

jp2masa:
The method to set entries was private in FAT class so I created a public method to clear a FAT entry that calls the method to set the entry. I read on a website that it was 0 on FAT entries and 0xE5 on directort entry, so I would thank you if you could help me on this

charlesbetros:
You're right about the free entry being 0x00000000 sorry about that. :)

jp2masa:
No problem. :)

PR #407, opened 2016-07-12
Keyboard Update

### LINKED ISSUE #405: How to switch to AZERTY keyboard ?
Hello, i use Cosmos but i'm french, so my keyboard is AZERTY.

How do i switch to AZERTY ? By default, it's in QWERTY.

jp2masa:
Keyboard Update: now it's possible to change keyboard layout(it wasn't due to Rings); Fix typo in Cosmos.System.Plugs.RuntimeTypeImpl.cs; Implement op_Inequality(implemented as op_Equality and not sure if it's needed with the changes but it's needed by many system methods)

Fixes #405


charlesbetros:
Why add ScanMap class when we already have ScanMapBase?

jp2masa:
Because ScanMapBase is in HAL abd cannot be accessed from User(Kernel.cs)

charlesbetros:
Yeah, but what I mean is you should put an abstraction layer on top of ScanMapBase rather than just creating another class as a pass-through with no functionality.

mterwoord:
@jp2masa The scanmap should go in Cosmos.System. If you add a keyboard layout there, you can specify it in your kernel.

fanoI:
I think that all scan maps things should remain at HAL what is needed a function (adding it to the Console class? Or something regarding Culture?) to change keyboard layout **after** the kernel is started.

For example I'd like for my Kernel to add a command equivalent of the DOS "keyb" now I cannot do this. I'm Italian and I have an Italian keyboard so it is an issue for me too!

charlesbetros:
That's what I was thinking. A keyboard class or Culture setting or something.

jp2masa:
So is it supposed to create a new project on ring System?

jp2masa:
In order to change keyboard layout, the only way on the current state would be overriding the GetKeyboardLayout method but it is of type ScanMapBase which is in HAL, and Kernel needs to be in User ring(else it gives an error)

mterwoord:
I still feel things are made more complex than necessary at this point..
Just move the scanmap implementations to the System ring, and you can specify them on the kernel.

jp2masa:
It's strange that the test failed, and I don't really know why it failed(I can't relate the symbol redefined error with my changes). Also, as I already said, moving the ScanMapBase to Cosmos.System doesn't work, only creating an empty ScanMap class(on System) that inherits ScanMapBase(on HAL)

mterwoord:
Don't move the ScanMapBase, just move the implementations.

MichaelTheShifter:
Just wondering, wouldn't it be possible to use the current ScanMapBase, and when switching keyboard layouts from user ring, pass an enum that's sitting in System for the type of layout you want, for example `Keyboard.SetKeyboardLayout(KeyboardLayouts.Azerty)` and have an if statement to determine what ScanMapBase to use based on the enum value?

mterwoord:
Possible, but more work to adjust laster on.

jp2masa:
I used the TestRunner and investigated the assembly file for the kernel. In fact the symbols that nasm says that are redefined they are defined 3 times, which is a big problem. I don't know correctly, but the issue can be in IL2CPU or is it in my code?

PR #434, opened 2016-08-14
Added ASM instructions and basic DateTime-implementation

Unknown6656:
Added the x86-instructions 'LEA', 'RDTSC' and 'RDMSR'.
Added a simplistic plug for 'system::datetime', which now contains the
method for fetching the current date/time


Unknown6656:
I shall append, that the added assembler instructions are not needed in this commit, but in future ones, as I am currently working on methods involving thread timing and time measurement.

PR #435, opened 2016-08-14
Added Stopwatch, Thread.Sleep and some CPUID implementation

Unknown6656:
I implemented the plug for `system::diagnostics::stopwatch` and `system::threading::thread::sleeep(int)` to enable some time measurement.
I also impelemented the basic function for tick and frequency count and some basic CPUID-implementations (e.g. retrieving the vendor id string)


mterwoord:
1. Naming: Can you please make the internal methods use proper casing (CanReadCpuID instead of canreadcpuid, etc)
2. Could you use x# (XS class) for the assembler plugs?
3. What are the private static fields for in ProcessInformationImpl?

mterwoord:
Besides that: great work!
:)

Unknown6656:
@mterwoord: Sorry -- I have been away yesterday ... I will fix this asap

Unknown6656:
@mterwoord : Concerning your three points:

1. Yes, I have renamed it (but forgotten to rename one occurrence, this was however fixed in https://github.com/CosmosOS/Cosmos/pull/435/commits/593aefad6c5e02208046df4fbfc424f36d39f259)
2. I have switched to the `xsharp::compiler::xs`-class, I did however add a few instructions for consistency reasons.
3. Ah, yes ... I have described this in the comment in the lines 11..16 in the following file: [`Cosmos/source/Cosmos.Core.Plugs/ProcessorInformationImpl.cs:L11-L16`](https://github.com/Unknown6656/Cosmos/blob/743b6c9a6f80376103ec089f764dfe37fc58dbb0/source/Cosmos.Core.Plugs/ProcessorInformationImpl.cs#L11-L16)

Unknown6656:
I know, that kbd-layouts have nothing to do with the original pull request ... but I did not want to open a separate request, as I would have to split the branch

fanoI:
So @Unknown6656 this PR is ready to be merged?
I imagine we have not automatic test for these changes: how to auto test that a stopwatch counted the expected time? Maybe something could be done for CPUID hardcoding the fact that the emulated CPU is a Pentium III for Bochs but this would not have sense on VmWare in which your real CPU will appear...

Unknown6656:
@fanol: Yes, this PR is ready to be merged.
I could maybe make a real test on a small computer, where I could fetch the clocking signal (maybe a raspberry pi?) and solder an counter to it to compare the software and hardware results....

fanoI:
No I concord with you these things you have implemented cannot be auto-tested. 
I'll proceed with the merge, excellent work!

PR #364, opened 2016-05-10
Added EqualityComparer Plug + Change In Plug Resolution

Seeker1437:
This Pull Request accomplished the following

1: Added DictionaryTests
2. Changed plug resolution in regards to resolving generic types

PR #437, opened 2016-08-18
add some kernel test for Collections (Hashtable, Dictionary)

zarlo:


fanoI:
@zarlo but the added test are not passing right? Dictionary are not yet implemented. This PR https://github.com/CosmosOS/Cosmos/pull/364 should implemented them (I hope) and as you can see it has similar test.
Probably this and https://github.com/CosmosOS/Cosmos/pull/364 could be in some way "mixed" in one thing...

Issue #359, opened 2016-04-23
Float instructions are a mix of X87 and SSE


fanoI:
The majority of float operations are implemented using SSE but some are using the old X87 FP co-processor of ancient 486. While this guarantee and old retrocompatibility (all X86 has X87 instruction set implemented) forces Cosmos to that lowest common denominator.

This is the list of  opcodes that are implemented using X87 instead of SSE:

- [x] Add
- [x] Add_Ovf
- [x] Add_Ovf_Un
- [x] Ceq
- [x] Cgt
- [x] Cgt_Un
- [x] Clt
- [x] Clt_Un
- [x] Conv_I8
- [x] Conv_R4
- [x] Conv_R8
- [x] Conv_R_Un
- [x] Conv_U8
- [x] Div.cs
- [x] Div_Un.cs
- [x] Mul  
- [x] Neg
- [x] Sub

fanoI:
Please note that in the end because you are using the X86 architecture not all OpCodes have been translated as SSE instructions for example it is not possible in IA32 to do double to long (64 bit integer) using SSE so in those few cases we are forced to use the legacy FPU.

PR #445, opened 2016-08-22
Continuation of Float work

fanoI:
- Fixed Single.ToString() on special cases (infinities, NaN and 0) and aumented the range of printable values
- Fixed Double.ToString(): it always printed "Double Overrange" for a bug in opcode ldarga
- Fixed opcode ldarga: the displacement of the argument variable was off of 4 bytes
- Fixed opcodes shr, shr_un and shl when the shift was more that 32 bytes, added to BCL relative tests
- Added BLC tests regarding BitConverter and unsafe code
- Moved the meat of the code of Single.ToString() and Double.ToString() to the class StringHelper together with the analogous methods for numer types
- Re-added _floatsignbit to CosmosAssembler.cs so the neg test should not fail anymore
- Removed all code relative to x87 that I had left under #if false
- Clean up


fanoI:
This closes https://github.com/CosmosOS/Cosmos/issues/359

PR #448, opened 2016-08-27
Tests Update

jp2masa:
## Changes:

- Added "Save Log" button to the Test Runner UI.
- Added tests for StreamReader, StreamWriter, BinaryReader and BinaryWriter for FAT.
- Added MemoryStream.
- Added autoscroll to the ListView.
- Fixed ListView column autosize.
- Added Copy Message option.

The tests are currently commented as they fail.


fanoI:
As you have seen recently the "Test Runner UI" have you noticed that text inside the psuedo "commandline" window is cut? If you will find a way to fix this too would be really fantastic.

jp2masa:
I'm sorry but I didn't understand, can you explain the issue better, so I can work on it?

fanoI:
If you try to run a test the text in the GUI window is cut right? It is a if it not 80 column but a little less...

https://postimg.org/image/3matak4k3/

do you see? The vertical scrollbar is not following the text (should be at the end).

jp2masa:
I usually resize the column, but I agree that it's better that it resizes automatically as well as the vertical scrollbar always in the bottom. I'll work on it.

fanoI:
Thank you!

PR #460, opened 2016-09-10
VBE Driver work

zarlo:
Renamed VGAMemoryBlock to VBEMemoryBlock


MyvarHD:
When i wrote the vbe driver, i did not think about the naming technically its a linear frame buffer, but vga stands for Video Graphic Array, witch is technically also correct, but vbe stands for VESA Bios Extension, VBEMemoryBlock is very wrong, VGAMemoryBlock is sort of correct but LinearFrameBuffer is the official name

mihail-rotmg:
I only care about one thing.
Will it run faster?

zarlo:
its not faster

MyvarHD:
with a smart doublebuffer, and virtual box it runs at 30 fps

mihail-rotmg:
How do i do that?

Issue #93, opened 2015-05-10
Improve Null Reference Exception handling


mterwoord:
Right now, visual studio shows a message like:

NullReferenceException occurred at address 0x020B9FBB! Halting now.

whenever a NullReferenceException occurs. This could be improved to show, at least the method name and instruction location it occurs.

cjhannah:
This would be greatly appreciated.

jp2masa:
It annoys me too. I'll try to do this.

MichaelTheShifter:
This is why I'd have a try/catch in my Run() that's a catch-all, and if an
exception occurs and isn't handled before it hits Run(), it prints
exception message to the console, then your OS just resumes and keeps on
going.

JuliusCaesar12:
I have found out that when a Null Reference Exception occurs anywhere, it causes an invalid opcode exception and halts the CPU.

Kiirx:
Still happening and still very annoying

ADev531:
Let's use map file, like other debuggers...

zarlo:
we do but its on the debugger (its off by default)

valentinbreiz:
Cosmos Gen3 has replaced the IL2CPU-based Gen2 codebase, and `gen3` is now the default branch of this repository. This issue was filed against Gen2, which is no longer developed, so it is being closed as part of the Gen3 cutover (#3088).

If you still hit this on Gen3, please open a fresh issue. The compiler, the build pipeline and the debugging story are all different now, so a new report will get a far more useful answer than reviving this one. Start here: https://cosmosos.github.io/

Gen2 remains on the `gen2` branch, and this issue stays readable.

PR #472, opened 2016-09-17
Add Labels description for NullReferenceException

jp2masa:
This PR helps on #93
Now labels are shown in the NullReferenceException
I don't know if it's useful to have labels on other exceptions, like StackOverflowException, but I can do it if it's useful.


jp2masa:
- Minor change in VS DebugEngine NullReferenceException message.
- Added NullReferenceException labels option (which is probably slower, so it is disabled by default, it doesn't work but shouldn't be hard to fix).
- Plugged ValueType.Equals(Object) but it's not tested, it allows to see that the next error in Zork is StreamWriter.

jp2masa:
I think I found out what I needed to do, I thought .map files were all the same...

PR #477, opened 2016-09-29
Fix Newobj for string and minor changes

### LINKED ISSUE #476: Possible RAT Bug
BitConverter.ToString() is not working properly, the last bytes converted to string look right but the strings converted from bytes before get modified and look strange.

For example, in the test code above, commenting lines 32 and 39 will cause variable "e" to be the only one looking right, while commenting only line 39 will result on variable "f" looking right, but it doesn't get printed. This allows me to say the problem is not on Console.WriteLine, but somewhere in BitConverter (which I'm not sure if it's plugged because I couldn't find a plug for it).

Test Code: http://pastebin.com/hrAEwXG3

Also, on AppVeyor a Path test is failing, and I suspect there's some problem with strings implemention, but I'm not sure what it can be or if it exists.

jp2masa:
## Changes:
- Fixed Newobj for string with length parameter.
- Minor changes in debugging and FAT test messages.
- Removed useless "if" in FAT, previously added by me.

EDIT: I couldn't fix the bug in DirectoryEntry creation.
EDIT: Fixes #476

### DIFF
--- a/source/Cosmos.System.Plugs/System/IO/DirectoryInfoImpl.cs
+++ b/source/Cosmos.System.Plugs/System/IO/DirectoryInfoImpl.cs
@@ -42,8 +42,8 @@ namespace Cosmos.System.Plugs.System.IO
                 throw new NullReferenceException("Failed to create DirectoryInfo: Directory '" + aPath + "' doesn't exist");
             }
 
-            aFullPath = aStorage.mFullPath;
-            aName = aStorage.mName;
+            aFullPath = aPath;
+            aName = Path.GetFileName(aPath);
         }
 
         public static string get_Name(DirectoryInfo aThis)
--- a/source/Cosmos.System.Plugs/System/IO/DirectoryInfoImpl.cs
+++ b/source/Cosmos.System.Plugs/System/IO/DirectoryInfoImpl.cs
@@ -36,12 +36,6 @@ namespace Cosmos.System.Plugs.System.IO
             Global.mFileSystemDebugger.SendInternal(aPath);
 
             aStorage = VFSManager.GetDirectory(aPath);
-
-            if (aStorage == null)
-            {
-                throw new NullReferenceException("Failed to create DirectoryInfo: Directory '" + aPath + "' doesn't exist");
-            }
-
             aFullPath = aPath;
             aName = Path.GetFileName(aPath);
         }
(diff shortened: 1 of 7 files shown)


source/Cosmos.System.Plugs/System/IO/DirectoryInfoImpl.cs
```diff
@@ -36,13 +36,19 @@ public static class DirectoryInfoImpl
             Global.mFileSystemDebugger.SendInternal(aPath);
 
             aStorage = VFSManager.GetDirectory(aPath);
+
+            if (aStorage == null)
```
charlesbetros: Don't throw an exception here.

MSDN remarks:
This constructor does not check if a directory exists. This constructor is a placeholder for a string that is used to access the disk in subsequent operations.

source/Cosmos.System/FileSystem/FAT/Listing/FatDiretoryEntry.cs
```diff
@@ -298,22 +316,21 @@ public List<FatDirectoryEntry> ReadDirectoryContents()
                 uint xFirstCluster = (uint)(xData.ToUInt16(i + 20) << 16 | xData.ToUInt16(i + 26));
 
                 int xTest = xAttrib & (FatDirectoryEntryAttributeConsts.Directory | FatDirectoryEntryAttributeConsts.VolumeID);
-                if (xStatus == FatDirectoryEntryAttributeConsts.UnusedOrDeletedEntry)
```
charlesbetros: We don't want to list deleted and unused entries. Don't we need this?

jp2masa: It is already skipped earlier with "continue;"

source/Cosmos.System.Plugs/System/IO/DirectoryInfoImpl.cs
jp2masa: So should I replace this with a debugger message?

charlesbetros: DirectoryInfo doesn't care if the directory exists.

jp2masa: But it throws a NullReferenceException if it doesn't, because the DirectoryEntry is null and it can't get the path

PR #486, opened 2016-10-13
Changes suggestion

jp2masa:
## Changes:
- Changes to IL2CPU folder organization and namespace, so it's easier to implement newer CPU platforms.
- Changed Cosmos.Core.Memory implementation slightly, doesn't change anything currently, as Cosmos.Core.Memory.Old is the currently used namespace for memory operations.
- Fixed warning in Cosmos.VS.Package about a reference that wasn't found, and it wasn't needed.
- Fixed a typo.

**This Pull Request is just a suggestion, as it changes an important namespace and part of the (I guess for future, as it's currently only tested with a Unit Test and not used by any part of Cosmos) Cosmos.Core.Memory implementation.**


jp2masa:
This PR makes no sense.

PR #487, opened 2016-10-14
Environment.NewLine Plug and tests

zarlo:


jp2masa:
I think the VBE changes shouldn't be applied, as @MyvarHD said in #460 and also because it's not the purpose of this Pull Request.

charlesbetros:
I agree. Let's try to stick to one issue one pr.

fanoI:
This other PR https://github.com/CosmosOS/Cosmos/pull/480 contains the test for this? If so could you add to this PR so AppVoyer properly check that the plug is working.

Thank you!

fanoI:
The test is failing but I'd say that is replace_test that has not the expected value is ""That is a test string." instead should be "this a test string" (or if you prefer could change the replace test with:
test = @"That is a test strin


g";)

jp2masa:
Try replacing `string new_expected = "this is a test string..";` with `string new_expected = "this is a test strin..g";`

fanoI:
OK now the test is passing but when you use a real newline it not worked right?

That is in the string:
"this is a test string.
"
after the '.' there was not the same thing as Environment.NewLine? While in code is more correct how you have written if you had read it from a file you get a string similar to that probably...

Tests/Cosmos.Compiler.Tests.Bcl/System/StringTest.cs
```diff
@@ -57,6 +57,18 @@ public static void Execute()
...
             Assert.IsTrue((test.Replace("This", "That") == replace_test), "string.Replace(string, string) does not work.");
 
+            test = "this is a test string." + Environment.NewLine;
+            string new_expected = "this is a test string..";
+            Assert.IsTrue((test.Replace(Environment.NewLine, ".") == new_expected), "string.Replace(Environment.NewLine, string) does not work.(Environment.NewLine)");
+
+			
+            test = @"this is a test string.
+			";
+            string new_expected = "this is a test string..";
```
jp2masa: Replace "string new_expected" with "new_expected"

zarlo: o how did i not see that

zarlo: note to self do not code in notepad

zarlo:
it has stopped crashing YAY you would not get an error i would just crash when you used 
```
Environment.NewLine
```
plug not neded

PR #495, opened 2016-10-28
ARMv7 Basic Support

jp2masa:
## Changes
- Added ARMv7 instructions to Cosmos.Assembler
- (**REMOVED**)Created basic ASharp.Compiler similar to XSharp.Compiler but a bit more limited in functionality
- Created AssemblerStyle in Cosmos.Assembler to allow different assemblers to be used with Cosmos, currently NAsm and GNU, and only the comment char depends on the compiler style for now.

EDIT: This is just a base, my objective is to run Cosmos on Raspberry Pi, but I'll focus on helping to fix bugs in Cosmos, just did this for "fun". I just want to help on adding ARMv7 support to IL2CPU much later.


charlesbetros:
I think we will wait for @mterwoord to review this one so it may be some time until he gets to it. Getting Cosmos running on ARM will be great. Thanks for the PR.

VortexIncAdmin:
Yeah, I was thinking about COSMOS on a Raspberry Pi. What Raspberry Pis are supported for COSMOS though? This could give more power with creating electronics on a Raspberry Pi, because YOU get to make  the OS, AND the electronic components.

jp2masa:
If we add support for ARMv7, at least Raspberry Pi 2 and 3 should work, but it's not a priority. I'd like to work on it, but there are other important things that need to be done first.

PR #508, opened 2016-11-11
Some chanes could be better...

MohammadSharper:
There's missing plugs for managing threads, we can't abort or getting IsAlive and... 
Please use the real Common Language Runtime, Use Mono, not IL2CPU.
This is not possible to load library in runtime from file system.
This is not possible to use "Reflection" for using classes types and their members.
Is there anyway to download complete plugs and add them to project? You need a social network to share exprience and complete templates and codes.
Also you can use linux as kernel and run c# codes on mono on linux kernel and compile it as a linux distro.
This is not possible for me because my knowledge and experience isn't enough.


MohammadSharper:
Also using custom fonts and using mouse pointer in Text Mode isn't implemnted.
A framework required to make programmimg easier for example ready to use drivers and gui components.

fanoI:
I think youhave not understood what Cosmos is... we cannot use the Net Framework or Mono inside Cosmos because they depend from an OS and Cosmos is an OS (if you want you could see it as a kernel + a net run time in the same binary image), we need our IL2CPU as any IL opcode has to be converted in assembler to be executed.

We are trying to replace the Linux kernel so using their kernel as base would not make sense.

I don't understand what this PR changes in any way...

fanoI:
While I can concord that some files as problem of "style" we prefer to do this "aestical" modifications when we are changing those files... merge a PR of 44 files in which there is really no new content makes no much sense sorry.

I recommend if you have other questions to join our chat.
Thank you!

MohammadSharper:
You can provide some types of project:
1. Native OS (I know that you convert c# to assembly language and compile it to the binary using nasm)
2. Linux distro (a linux kernel+mono and c# codes in initramfs)
3. Native binary (Your OS isn't boot directly, it boots from isolinux)

Can I use this to boot a linux kernel? Like grub and isolinux and ...
How to execute a binary? An OS has utilites and external programs.
An OS isn't a singel binary file.

PR #509, opened 2016-11-13
My speak isn't ended

MohammadSharper:
You have closed my pull request but my speaks isn't ended.
I have plug problem with threading. I can't create multi-thread OS.

You can provide some types of project:
1. Native OS (I know that you convert c# to assembly language and compile it to the binary using nasm)
2. Linux distro (a linux kernel+mono and c# codes in initramfs)
3. Native binary (Your OS isn't boot directly, it boots from isolinux)

Can I use this to boot a linux kernel? Like grub and isolinux and ...
How to execute a binary? An OS has utilites and external programs.
An OS isn't a singel binary file.


jp2masa:
2. Cosmos isn't related with Linux in any way and I think it won't be, so if you want to create a Linux + Mono OS you can simply compile them and make a simple Linux distro.

I didn't understand the difference between 1. and 3. is it in the bootloader?

fanoI:
As I've already said this is where the questions / discussion should be:
https://gitter.im/CosmosOS/Cosmos

There is a FAQ of Cosmos here that tries to explain what Cosmos really is / will be:
https://github.com/CosmosOS/Cosmos/wiki/FAQ

It is obvious that in future Cosmos will have multithreading, networking and so on but all of this will came when we will be ready to implement it. Any help is obviously accepted but we never throw way our kernel to use the Linux one... Cosmos is a kernel!

charlesbetros:
Please create a github issue and not a pull request, or join our Gitter chat for questions and suggestions. Pull requests are not meant to be used questions and suggestions. They are for submitting code changes. Once again, we are more than  happy to answer questions and chat about your suggestions in our Gitter chat or through a GitHub issue. Thanks!

fanoI:
Regarding point 3: no OS boot directly they always need a bootloader Linux has them (a lot of them! The most famous is probably Grub), Windows has it (if it is the only one OS you don't see it but it exists  and does the boot!), BeOS / Haiku OS too.

So how Cosmos could boot directly? Please don't confound yourself with the bootloader name: Syslinux is not Linux based!

zarlo:
what is this for???

MohammadSharper:
> So how Cosmos could boot directly? 

A MBR Code, An EFI code boots directly. forexample, bootloaders boot directly from the frimware (BIOS)
and there's no bootloaders.
We only can write basic OSes and Simple Bootable applications, It's good for learners. Also, I want use it as an Bootloader. for example, Make a Bootable disk with a menu that written in C# with Cosmos.
Did you undrestand?

fanoI:
For now we have decided to not write our bootloader (that while technically could be seen as part of the OS is different code with no link with the OS itself... indeed BeOS "Bootman" could make Linux start as Grub could make start Windows for example).
In UEFI there is a bootloader it is simply integrated into the BIOS itself!

Yes the actual version of Cosmos is limited to have the kernel and "applications" in the same image but in future as I already said we will add processes / threads so it will not only useful to write "toy OS".
Please note that in some particular embedded work having OS + application "integrated" could be desiderable.

MohammadSharper:
You don't undrestand.
I want to develop a Bootloader, Not OS.
Do you know grub how boots? It has a MBR code that searchs and load the Grub image (grldr) from disk.
I want to write this grldr and this MBR code with C#.
You have developed IL2CPU that easily can convert any C# code to binary, so this is not impossible for you.
A Bootloader, don't need to be boot from another bootloader like ISOLinux or Grub, It boots **directly**.
I have described, so I Hope you undrestand.
Sorry, I am only a basic English learner and I am not too professional in translating.
You are my teacher! (Shoma Ostad Maa Hastid) and I'm only a basic C# Programmer.

fanoI:
Well doing a Boot loader in C# will be really interesting yes IL2CPU and what you need for this purpose but you need your run time too... the problem is that a boot loader probably would be better as a separate project from Cosmos but actually could be some "interdependence" between IL2CPU and Cosmos that should not be present in theory.
Before to start this work I conseul to wait that @charlesbetros ends and merge the Core RT work so you will have a more "stable" point to start to work (a lot of plugs hopefully) will be not needed.

You probably need X# too as - I suspect - a boot loader is more assembler code that anything else, right?

MohammadSharper:
I just give you my idea.
This is too hard for me. I just following your project for fun.
I'm not serious with cosmos.
Booting my own OS is very interesting. Not?
Anything you want do with it. I just give the key.

MohammadSharper:
And Sorry!
I didn't read your notices about pull request.
I think we use this for requesting new features for projects.
I'm too sorry.

PR #536, opened 2016-12-18
Preliminary version of CGS [Not merge this yet!]

fanoI:
This is the first version of CGS:

1. It works only with VBE (that is only with Bochs)
2. It permit to create a new Screen with the desidered mode
3. And to Draw a point on Screen

In reality crashes spectacularly with a glorious "Stack Overflow" 👎


fanoI:
@cjhannah The DrawPoint() inside Canvas is a really generic implementation that no driver should really use. Indeed VBEScreen overwrites it doing a sort of memset() into the LinearFrameBuffer, with deviceces that on in hardware this will be yet more faster.

Issue #26, opened 2014-12-18
Add HyperV support


mterwoord:
Add support for hyper-v. (Client/local HyperV, bonus points for remote support)
This involves:
- Adding settings pages to the project for configuring the HyperV mode 
- Adding support for creating and starting the HyperV virtual machine
- Adding support for debugging HyperV machines
- Possibly, fix any bugs in the kernel which are uncovered in the process..

PR #550, opened 2016-12-27
Added basic Hyper-V support

Suchiman:
**What works**
- Added HyperV DebugEngine that creates a default virtual machine and starts it up / stops it.
- Added DebugConnectorClient since Hyper-V only supports named pipe server mode
    - Unfortunately this uses a new named pipe `CosmosSerial` instead of `Cosmos\Serial` since Hyper-V throws an InvalidArgument error when trying to configure `Cosmos\Serial` as named pipe.
- Added testrunner for Hyper-V
- Added project configuration integration and hyper-v preset
- Added `Filesystem.vhdx` harddrive with contents of `VMWare\Workstation\Filesystem.vmdk` (unfortunately i couldn't get the dynamic expanding disk smaller than ~36MB)
- Creating a new project, choosing HyperV preset and starting debugging
- All stable test runner tests pass
- Refs #26 

**Todo**
- Getting debug connection stable
    - Sending commands from VS to the VM stops working after a few commands, looping in here [DebugConnector.Sending.cs#L151-L168](https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.Debug.Common/DebugConnector.Sending.cs#L151-L168)


Suchiman:
@jp2masa It shouldn't break anything AFAIK, it has only additions + 2~4 parts where i've changed existing code in the style of `if (hyper) { ... } else { previous }`

MihaMarkic:
Since it has been merged, is Hyper-V supposed to work now?

Suchiman:
@MihaMarkic it's been some time since i've played with it but last time, i was able to configure my project for Hyper-V and start the kernel. The biggest issue was, that evaluating variables and expressions during debugging would break the debugging sooner or later.

MihaMarkic:
@Suchiman Oh, well, thanks. I wish hyper-v support will come rather sooner than later. I don't mind using VMWare but if you have hyper-v enabled than you have to use it, unless you are willing to restart the OS to switch between the two.

MihaMarkic:
For anybody interested, I tried Hyper-V today and these are the results:
1. Debugging doesn't work - it'd say "waiting for debugger"
2. With debugging disabled, it loads kernel (one from the template) accepts few keystrokes and then freeze.
3. It mostly works only first time I run the app. If I run it next time it complaints that .cdb file is locked. I have to close visual studio and kill cosmos.hyperv background process to make it work again.

PR #552, opened 2016-12-29
Add check for BGA

cjhannah:
so they don't use it on systems where it can't be used


fanoI:
Thank you for your contribution!

My idea for CGS is that the static method GetFullScreenCanvas() of FullScreenCanvas will return the right instance of the "canvas" driver to use (a fake plug & play if you want), the check that you have added could in any case have sense but it would be better to do it in the VBEDrinver constuctor not in one of its method. As actually there is no explicit constructor I'd say add one and throw there if the BGA is not found.

cjhannah:
added to the constructor

PR #571, opened 2017-02-03
Console Optimize 

zarlo:


fanoI:
I'm unsure doing 4 calls instead than one could be more faster sincerely... how you have benchmarked it?
I think the real speed up would to do buffering stdout instead that write any single char on the Console...

Instead as you are making see the code I'm mistaking or '\r' does nothing?
https://github.com/zarlo/Cosmos/blob/4ff5b1835c7f42e037b72eca562292883bcbef2a/source/Cosmos.System/Console.cs#L95

is that an empty statement?
`Console.Write(\'r')` should move the cursor back at the start of the current line, to be more "portable" the check on '\n' should use the predefined Console.NewLine (that could be '\n' or '\r\n' depending on how one prefer).

zarlo:
I see it as one less for statement 4 less else if statements but yes i think a buffering stdout would help

fanoI:
Ah I see it now the handling of '\t' is done inside Write(String) and doing Write("    ") effectively we are calling Write(String) inside Write(String)!
A doubt while we are "hardcoding" that NewLine is '\n' historically on Windows it was '\r\n' (technically more correct as you move to the right the cursor and then go to the next line) while on the Unix world it is simply '\n' maybe it was for this that '\r' was ignored?
Probably the right thing would be to check for Enviroment.NewLine (that can be 2 char so makes things more complex!) and do NewLine while do move the cursor to the right only when you see '\r' alone...

Sadly it is not rare to see stricly not portable code like this in C#:

`Console.Write("Text\nText\n\n")`
instead of
`Console.Write("Text "+  Enviroment.NewLine +"Text" + Enviroment.NewLine + Enviroment.NewLine);

PR #573, opened 2017-02-06
First implementation of processor info

ERamaM:
<strike>Here is the first implementation of ProcessorInformation using SMBIOS.

The first advantage to use SMBIOS is that you can get information on every device in the PC "easily".

I provided an interface for all the processors in the system provided by the SMBIOS table (you can get basic info this way).

As is expected, this is full of bugs, needs heavy refactoring and the convention of the variables is not "good".

However, I decided to submit an early implementation to discuss it.

To use it you have two options:

- Call the static method ProcCpuInfo() in the class Cosmos.System.PCInfo.ProcessorInfo which gives a raw string similar to linux /proc/cpuinfo. NOT RECOMMENDED. **Only for debugging purposes**.
- Use the attributes (public variables) for each processor in ListProcessors to gather specific information. RECOMMENDED.</strike>

Seems that SMBIOS is not intended to be applied to this issue (it's designed to be used by the end-user, no OS developers) (source:  http://f.osdev.org/viewtopic.php?f=1&t=18280&start=0)  However, an implementation of the parser stays if needed in the future (if we find an indispensable parameter unparseable or we want to "port" some sort of DMI decode). 

**Note**: if the parser is needed, will need to reattach it to HAL and system.

We take the old route: x86 cpuinfo calls (for now, perhaps we will need something more in the future). We assume that only a single processor exists so we make the parsing simple. Relevant links if needed:

https://en.wikipedia.org/wiki/CPUID
http://wiki.osdev.org/Detecting_CPU_Topology_(80x86)
http://wiki.osdev.org/CPUID

# CPUID

There is an instruction in x86 that allows retrieving the "flags" for a specific x86 processor: **cpuid**. Acs depending on the previous value of EAX. May have different meanings depending on the processor.

In SMBIOS, the table "processor information (type 4)" has a field that has this same information. What's more: this field is independent of the architecture (x86 and arm only) but needs a different parsing (since the bits have a different meaning). **UPDATE**: this allows reusing part of the SMBIOS code. 

#  Known issues

- <strike> Doesn't work well with Bochs (due to a Bochs "unexpected" bug: https://sourceforge.net/p/bochs/mailman/message/20126548/) (http://forum.osdev.org/viewtopic.php?f=1&t=17818&sid=4069e9efc19bcb4ee2e716%2056d693b6eb). Seems that the "bug" is not going to be repaired soon (since its not a "bug")</strike> **UPDATE:** the bug referred to an old version of Bochs so it "seems" fixed... **UPDATE 2**: Bochs doesn't complete the SMBIOS tables at all.

# TODO

- [x] Get processor signature and parse it.
- [ ] Get additional feature info and parse it.
- [ ] Get frequency of CPU in a reliable way (with interrupts if needed).
- [ ] Get cache and TLB information.
- [x] Get vendor string of the CPU.
- [ ] Get extended features (eax=7, ecx=0)
- [x] Get the processor's brand.
- [ ] Get ACPI information (if possible)
- [x] Provide an interface to the system to use the information classes. **NOTE**: this interface provides unmeaningful numbers as /proc/cpuinfo (such as "CPU family", "model", "stepping". Would be nice to provide a meaning of these numbers (as an enum for example) if possible.
- [ ] Make the processor interface independent of the architecture.
- [ ] Detect the processor topology.
- [ ] Clean code (move code to functions when needed).
- [ ] Change the workarounds with .NET functions when possible.


ERamaM:
So far:

- I moved the [Inline] methods to separate classes (seems logical).
- The method GetCPUID was the second method you requested: a facade. (Note that in HAL the classes are not unsafe).
- I didn't implement the argument as a int in the second CPUID (which is an overload). I do not understand clearly the purpose of this since using an enum it is more clear. Even if you want to use the parameter as an argument to the unsafe CPUID you could do:

```csharp
//This is the facade CPUID
public int[] CPUID(CPUIDOperation.GetWhatever cpuidop){
    int arg = (int)cpuidop;
    uint eax, ebx, ecx, edx;
    //Call the plugged CPUID
    CPUID(arg, &eax, &ebx, &ecx, &edx);
}
```
Which is "arguably correct" (if you adjust the enum values) but some functions require more than 1 call to the CPUID instruction (namely the call to obtain the processor brand which requires 3 calls). Furthermore, its clearer to the end user of the method (since the parameter is "autodocumented").

fanoI:
I'll try to explain my idea better I think this function should not be needed:
https://github.com/ERamaM/Cosmos/blob/c1fb69edd7eb7078243531cbd40bdc0d4b8b2d75/source/Cosmos.Core/PCInformation/ProcessorInformation.cs#L26

what the scope of the enum if I've to do a switch on any element (they are more than 30 possible commands if I remember correctly) and than convert manually from CPUIDOperation.GetVendorID (0) to 0?
I don't understand this comment too:

> <remarks>There is not a 1:1 correspondence between the enum and the cpuid call.</remarks>

that probably explain why that kludge is needed... that is calling it with 1 the first return GetProcessorInformation  while the second time GetFlags? Or is that an error and there should be a "2" (that is the same thing GetFlags is / should be as enum value)?

If there is correspondence between this list and the operations you have to do well it is all really weird:
https://en.wikipedia.org/wiki/CPUID

my idea was the following in Core should exist only the "engine" that is CPUID and the other methods needed to get CPU frequency, timestamp and so on...

it was for this that I've tough to have a more unsafe version easier to set from Assembler (with the 4 int pointers) and one safe in which an int array is returned, better to "hide" in C# the fact that these was CPU registers (in a ideal world they should return an uint128 and all will be more simple!) so one could:

``cs
// plugged (notice the "private")
private static void CPUID(int command, int *result1, int *result2, int *result3, int *result4) {

}

https://en.wikipedia.org/wiki/CPUID#EAX.3D1:_Processor_Info_and_Feature_Bits

ERamaM:
So, here is the catch.

The infamous Thread.Sleep is ALREADY IMPLEMENTED!. But what is the problem, you may think? That the classes that implement that seem to be like ProcessorInformation, i.e, probably they doesn't work (they do calls to ProcessorInformation of "unplugged plugs").

Here is how to implement sleep it: http://referencesource.microsoft.com/#mscorlib/system/threading/thread.cs,6a577476abf2f437,references
Which is implemented is SleepInternal.

Fun fact: i deleted the classes and nothing happened. Like the classes were never used.

As far as the changes requested is concerned: i managed to reduce the switch to one case: brand string (which requires three calls to CPUID) and changed the accessibility of the plugged methods (i didn´t know that the plugged methods could be private, lol). Also implemented RDMSR (but its is not very useful since sometimes gives 0s but that is something i need to work out)

More things. I have been thinking about "abstracting the processor". I came up with a hierarchy that could be done on Cosmos.Core (replacing ProcessorInformation). The hierarchy is as follows: 

macedonios

Next, in hal, we would some "abstract factory" instantiation like follows:

````csharp
public void getProcessor(){
    IProcessorFactory factory;
    Processor proc; //Abstract processor

    #if intel
        factory = new IntelProcessorFactory();
    #endif
    #if amd
        factory = new AMDProcessorFactory();
    #endif
    #if ARM
         factory = new ARMProcessorFactory();
    #endif
    proc = factory.getProcessor();
}
```

Where get processor would instantiate a concrete processor (Intel, AMD, ARM...).

I know it is a bit early to think about this but it is a first approach.

fanoI:
Regarding Thread.Sleep() we have already seen that in Net Core is done differently (no more SleepInternal()) so the idea is to retry to plug it when Net Core will have been merged in master branch (really soon I think).

The changes you have done are OK, I like your idea to use polimorphism to resolve the fact that Intel and AMD have some CPUID commands with different meaning the only thing is that a good part of these classes should be done in HAL not in Core as I've already said in Core should be only the less possible (code that needs assembler as CPUID itself or access to I/O ports), for now let's not think of ARM it will be totally different but it is not a problem as code in Core and HAL could be architecture specific.
But how to instantiate an IntelProcessorFactory() or AMDProcessorFactory()? Using "#if intel" presupposes that this is known at compilation time (that you cannot known) the only way to know the CPU manufacturer is to use CPUID? We have a sort of chicken and egg problem...

Unknown6656:
Glad to see that there is some progress concerning CPUID etc.
I worked on some of these files before, but I did not have enough time for quite a while to continue working on it...

fanoI:
Yes you and @ERamaM had done really a good work with this!

For me could be already sufficient to be merged as at least the basic thing of "cpuinfo" are present the only things is that the conflicts should be resolved.

Ideally this should be before Net Core is merged so the operation is more simple (a lot of things have been changed / moved around in the Net Core branch).

charlesbetros:
@quajak is this work already in mater?

PR #611, opened 2017-03-31
Update OpNone.cs for Math.Round()

teokitan:
OS says Math.Round() is not implemented without the changes. It should work now!

### DIFF
--- a/source/Cosmos.IL2CPU/ILOpCodes/OpNone.cs
+++ b/source/Cosmos.IL2CPU/ILOpCodes/OpNone.cs
@@ -141,8 +141,9 @@ namespace Cosmos.IL2CPU.ILOpCodes
         case Code.Ldelem_U1:
         case Code.Ldelem_U2:
         case Code.Ldelem_U4:
-          return 2;
         case Code.Ldelem_R4:
+        case Code.Ldelem_R8:
+          return 2;
         case Code.Ldnull:
           return 0;
         case Code.Dup:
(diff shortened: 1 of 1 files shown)


source/Cosmos.IL2CPU/ILOpCodes/OpNone.cs
```diff
@@ -142,6 +142,7 @@ public override int GetNumberOfStackPops(MethodBase aMethod)
         case Code.Ldelem_U2:
         case Code.Ldelem_U4:
           return 2;
+        case Code.Ldelem_R4:
```
jp2masa: `case Code.Ldelem_R8:` should be here too.

source/Cosmos.IL2CPU/ILOpCodes/OpNone.cs
```diff
@@ -273,7 +274,8 @@ public override int GetNumberOfStackPushes(MethodBase aMethod)
         case Code.Ldelem_U1:
         case Code.Ldelem_U2:
         case Code.Ldelem_U4:
-          return 1;
+        case Code.Ldelem_R8:
```
jp2masa: `case Code.Ldelem_R4:` should be here too.

source/Cosmos.IL2CPU/ILOpCodes/OpNone.cs
jp2masa: `case Code.Ldelem_R4:` as well as `case Code.Ldelem_R8:` should be in the line above `return 2;`

Issue #603, opened 2017-03-12
[CGS] Add to the Canvas abstract Class methods to draw more shapes


fanoI:
Continuing to take inspiration from System.Drawing...

- [x] DrawFilledRectangle(Pen pen, int x_start, int y_start,int width, int height)
~~- [ ] DrawSquare(Pen pen,  int x_start, int y_start, int int height)~~
- [x] DrawTriangle(Pen pen, int vertex1, int vertex2, int int vertex3)
- [x] DrawPolygon(Pen pen, int[] vertexes)
- [x] DrawCircle((Pen pen, int x_center, int y_center)
- [x] DrawEllipse(Pen pen, int x, int y, int width, int height)
- [x] Add to the CGS Demo examples of drawing of these shapes

fanoI:
#601

miyu:
I can take this for assignment if it's still available.

fanoI:
Sure that it is available!

Keep us posted on the progresses.

fanoI:
To help you to start these methods should be implemented in the abstract Canvas class taking as example how is done DrawRectangle():
https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.System/Graphics/Canvas.cs#L193

in Canvas do exist the software implementation a more evolved GPU than Bochs should override them if it has an hardware way to do all (or some) of these... I notice now that I've forgotten to add the virtual attribute to DrawLine() and DrawRectangle() please add it if not they could not overridden :-)

DjAlEx234:
I can take this assignment

nuhash:
All of these functions have been implemented. This issue can be closed now.

fanoI:
Yes thank you for the help!

PR #628, opened 2017-04-28
Canvas Update

DjAlEx234:
Add DrawFilledRectangle()


fanoI:
Now I see that the code not compiled but do you have tried it on your machine before creating the PR?

I suggest you a more simpler approach instead to draw rectangles one inside the other it will be better to use DrawLine you can draw a simple draw horizontal lines until the rectangle has been filled.
If you want to test your modifications there is a Canvas TestRunner that you can use.

Thank you for the help!

valentinbreiz:
I tried to compil cosmos with https://github.com/CosmosOS/Cosmos/pull/628 and and Cosmos does not build.

fanoI:
There is obviously no hurry :-)
Let @DjAlEx234 correct its code when he wants...

DjAlEx234:
where is the test runner?

fanoI:
The graphic TestRunner is here: https://github.com/CosmosOS/Cosmos/blob/master/Tests/GraphicTest/Kernel.cs

to enable it comment the other test and leave only it from here: https://github.com/CosmosOS/Cosmos/blob/master/Tests/Cosmos.TestRunner.Core/TestKernelSets.cs

you simply execute tests use the TestRunner project (the text mode one, the graphic based does not show Cosmos'  console so no GUI seen!).

source/Cosmos.System/Graphics/Canvas.cs
```diff
@@ -230,6 +230,29 @@ public void DrawRectangle(Pen pen, int x, int y, int width, int height)
             DrawLine(pen, xc, yc, xd, yd);
         }
 
+        public void DrawFilledRectangle(Pen pen, int x_start, int y_start, int width, int height)
+        {
+            for (int i; i != width; i++)
```
jp2masa: It should be `int i = 0`

source/Cosmos.System/Graphics/Canvas.cs
```diff
@@ -230,6 +230,29 @@ public void DrawRectangle(Pen pen, int x, int y, int width, int height)
...
+            {
+                DrawLine(pen, x_start, y_start, x_start + height, y_start);
+                y_start++;
+            }
+        }
+
+        public void DrawSquare(Pen pen, int x, int y, int sideslength)
```
jp2masa: It doesn't make sense to have DrawSquare nor DrawFillesSquare, as it can be done using DrawRectangle and DrawFilledRectangle.

fanoI:
Another thing the method that could be overriden by a driver that has these "shapes" in hardware should be done virtual so DrawLine(), DrawRectangle() and DrawFilledRectangle() must be all virtual so as all the methods requested in https://github.com/CosmosOS/Cosmos/issues/603.

Thank you!

DjAlEx234:
What do you mean by virtual?

fanoI:
Add virtual modifier to the method for example:

```cs
 public virtual void DrawLine(Pen pen, int x1, int y1, int x2, int y2)
```
Bochs not having the capacity to draw lines in hardware will use this software implementation but a more powerful GPU (VmWare has already some emulated "hardware" capabilities) will override this method to implement its version of DrawLine() but it cannot do it if DrawLine() is not declared virtual in the Canvas abstract class.

DjAlEx234:
So can I make a copy of every drawX void and make it virtual?

fanoI:
No you not have to copy anything the methods: DrawLine(), DrawRectangle() and DrawFilledRectangle() must have the virtual modifier nothing else have to be changed.
The method Clear() is one example, in that case the Bochs driver ovverides it too because there is a more performant way to do it: https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.System/Graphics/VBEScreen.cs#L99

The other methods in the more "shapes" issue should be virtual too maybe some very advanced GPU gives the possibility to draw triangles in hardware so it can be overridden too.

valentinbreiz:
I'm writing the .DrawString method, here an exemple for the letter A:
canvas.DrawLine(pen, x, y + 1, x, y + 7);
canvas.DrawLine(pen, x + 4, y + 1, x + 4, y + 7);
canvas.DrawLine(pen, x + 1, y, x + 4, y);
canvas.DrawLine(pen, x + 1, y + 4, x + 4, y + 4);
There is a way to write this method well... faster ? x)

jp2masa:
@valentinbreiz I don't know if we should have a DrawString method, it should be implemented by the graphics library, because fonts can be vectorial or use images for letters. The example you provided is vectorial, but it should be handled by the graphics library. We should provide methods for drawing vector images, like drawing lines, bézier curves... I don't know if we should handle drawing images, I think it should be implemented by a graphics library too.
@fanoI What do you think?

fanoI:
DrawString() and DrawImage() was in the list of "CGS New Features" but I've put them there I didn't tough well where they should be... probably as you say they deserve intermediate libraries that will use CGS as building block.
Probably to DrawString() efficiently the solution is to have before DrawImage() and to create the string as bitmap / vectorial imagine in memory, so to Draw a String we need:

1. Font support, makes sense in this age to support only TrueType that seems they have a sort of VM to be implemented to do correct rendering
2. Create the "image" of the String in Memory
3. C# DrawString class has Brushes, Styles a lot of things...
4. Draw the String as an Image

So in the end I'd say for now let's continue with what is in https://github.com/CosmosOS/Cosmos/issues/603 that are really more primitive graphic types we think of "how to DrawImages" and "how to DrawStrings" after.

DjAlEx234:
can we add a project that writes letters from a library?

fanoI:
First a premise:
Text rendering is much complex issue then "pasting" some glyphs... Not just much complex, it is very complex: kerning, ligatures, spacing, bidirectional text, vowels, and much more...

Microsoft itself divides the problem in two parts DirectWrite and Direct2D / GDI+ (legacy), Direct2D is in some way the equivalent of our Canvas class.
https://en.wikipedia.org/wiki/DirectWrite

The OpenType format (successor of TrueType):
https://en.wikipedia.org/wiki/OpenType

In conclusion how to Draw Text in Cosmos needs more study and sure should affronted in a separated PR.

source/Cosmos.System/Graphics/Canvas.cs
```diff
@@ -230,6 +230,22 @@ public void DrawRectangle(Pen pen, int x, int y, int width, int height)
             DrawLine(pen, xc, yc, xd, yd);
         }
 
+        public void DrawFilledRectangle(Pen pen, int x_start, int y_start, int width, int height)
```
jp2masa: This method should be virtual.

source/Cosmos.System/Graphics/Canvas.cs
```diff
@@ -230,6 +230,22 @@ public void DrawRectangle(Pen pen, int x, int y, int width, int height)
...
+        {
+            for (int i = 0; i != width; i++)
+            {
+                DrawLine(pen, x_start, y_start, x_start + height, y_start);
+                y_start++;
+            }
+        }
+
+        public void DrawTriangle(Pen pen, int v1x, int v1y, int v2x, int v2y, int v3x, int v3y)
```
jp2masa: This method should be virtual.

prepare:
Hello @fanoI,

I want to introduce a pure C# framework that can

1.  Read OpenType font (.ttf, .otf)  

2. Layout a string the glyph-spans and layout each glyph. (with OpenType glyph layout)

3. Render it to a Bitmap surface with grey-scale technique or with LCD-effect technique
    (subpixel rendering)

Here =>https://github.com/LayoutFarm/Typography

---

I think my lib can help your graphics system.

:)

prepare:
It is not only advertisement.  I want to help you too. 

But I don't known much about Cosmos.

I need a time to learn your system first.

fanoI:
@prepare thank you for your help offer!
We have already bitmap supported so drawing text is really the next step of CGS.

PM me if you have requests on how Cosmos works.

PR #630, opened 2017-04-29
VFS changes

fomenok:
Made possible the creation user VFS (not dos-like) and create user file systems


fanoI:
I have not a lot of experience on VFS so better to expect @charlesbetros on this.
The modifications you have done broke our tests (as AppVoyer clearly shows), it seems that Bool.GetHashCode() needs a plug.

fomenok:
I fixed

charlesbetros:
Thanks!

Issue #542, opened 2016-12-23
Implement ACPI HALT command


fanoI:
Cosmos can not shutdown the machine but only boot it and do reboot.
To implement a proper shutdown Cosmos ACPI support should be enanched to support the HALT command; at the user level the expected interface is simply kernel.Stop().

A starting point:
http://wiki.osdev.org/Shutdown

asm2620:
I have this problem with MonoOS, that based on linux.
When I run /sbin/poweroff, It stops working and screen will freeze.

fanoI:
On Linux HALT is surely implemented I usually do: 'shutdown -h now' not poweroff...

Please left this issue for the actual Cosmos implementation of the HALT command.
Thank you.

DarrellBailey:
It appears as though the current implementation doesn't even implement reboot.

The `Kernel.Reboot` method is empty?

fanoI:
Yes Kernel is an abstract class with apparently all empty methods but are plugged somewhere else I think.

However I've read that there were version of Halt command implemented in user kernel so why don't port this implementation where it should be inside the Cosmos source tree?

DarrellBailey:
@fanoI 

That code seems like it will do the trick. I can implement this in the core and bubble it up to the kernel layer but this wont be a full implementation of ACPI. In fact, to fully implement acpi would be a huge undertaking by itself which is why intel releases the acpica (https://github.com/acpica/acpica)

This will, however, be good enough for now.

fanoI:
Do you mean the difficulties are to implement the standby / resume and so on? For now having the possibility to really shutdown the machine will be sufficient for us.

The code I've linked I'd say needs refactoring as too much unsafe code is needed probably it will need to splitted between Core (were unsafe can be used) and HAL rings.

jp2masa:
@fanoI No, I found that code in a tutorial, in the start of the file it says:
```
/// <summary>
/// ACPI and Power Management Functions
/// 
/// Provided by Grunt to app-builder on the Cosmos discussion boards
/// </summary>
```

fanoI:
Closed by https://github.com/CosmosOS/Cosmos/pull/671

PR #633, opened 2017-05-04
Implementation of Sys.Power.Shutdown();

valentinbreiz:
That works and should replace the Stop(); method ! (https://github.com/CosmosOS/Cosmos/issues/542 fixed)
PS: That fixes "Stack overflow occured at address 0x020AA29D! Halting now." after using Stop(); with canvas.
Thanks to @TheCool1James for his ACPI.cs :)


source/Cosmos.Core/CPU.cs
```diff
@@ -62,5 +63,12 @@ public static bool DisableInterrupts()
             mInterruptsEnabled = false;
             return xResult;
         }
+
+        public void Shutdown()
```
jp2masa: I think that CPU shouldn't have the Shutdown method, it's not related to the CPU, but to ACPI.

valentinbreiz: Ok I will try tomorrow.

fanoI:
I'm asking myself if it was better to retain the legacy Kernel.Stop() that maybe user OS are already using or not... on the other hand a Power class makes more clear the intent...

And probably Power class it will make yet more sense when we add Standby and Hibernate.

valentinbreiz:
@fanol I think it's better to use Power class, as you said it, that makes more clear the intent.
And for more future features it is "cleaner" to do like that.

source/Cosmos.Core/CPU.cs
valentinbreiz: @jp2masa why Reboot method is also in Cosmos.Core/CPU.cs ?
Because Cosmos can be rebooted from ACPI too.

jp2masa:
The current reboot method is done on the CPU, we should have ACPI restart too.

valentinbreiz:
@jp2masa there is a reboot method in the ACPI.cs : https://github.com/valentinbreiz/Cosmos/blob/7708d564c2b03dc94c08882156c25d60b3b67eec/source/Cosmos.Core/ACPI.cs#L317
It just have to be implemented to Power.cs.

fanoI:
The actual kernel class: https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.System/Kernel.cs had a AfterRun() "callback" that permitted to do something before the OS died this could be useful when we have a true OS in future and - for example - when the user request shutdown we can send a message to all the open application to close gracefully (and give them the possibility to save open documents too).
Retain the actual behavior using ACPI to do a true shutdown seems easy:

https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.System/Kernel.cs#L98
when AfterRun() returns simply call Power.Shutdown().

Regarding reboot should for the user point of view should be a call to Kernel.Restart(): https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.System/Kernel.cs#L140 that - in turn - will call Power.Reboot().

@jp2masa what do you think?

jp2masa:
Shutdown should be handled in AfterRun(). The correct way to shutdown the OS would be Kernel.Stop(), so that the OS loop stops and AfterRun() gets called. Power.Shutdown() would be the last method to be called in AfterRun().
I think that we should check if ACPI is present, and if not, use CPU reboot. This check would be in the Power class.

fanoI:
Well in any case ACPI is not present it Power.Shutdown could simply return and we will be in the same situation of now the Run loop ends and I don't know if the caller of "Kernel" (does it exist?) issued something similar to ASM instruction CLI, HALT... or simply CPU was yet on but our code was not existing anymore?

jp2masa:
CPU will work while it is being powered, only ACPI can really stop it. It just doesn't execute any instruction I think.

valentinbreiz:
@jp2masa I moved ACPI.cs to Cosmos.HAL and deleted the old Reboot method from Cosmos.Core.CPU.

jp2masa:
The Reboot method in CPU should be kept. Try moving the ACPI IOPort to Cosmos.Core.IOGroups.

valentinbreiz:
Why it should be kept ?
And with ACPI IOPort you say the ACPI.cs I added ?

jp2masa:
It should be kept in case the system doesn't have ACPI. There is an IOPort declaration in ACPI.cs that should be moved to Cosmos.Core.IOGroup.ACPI, and added to Cosmos.Core.BaseIOGroups.

TheCool1James:
Hello. I would like to point out that _that_ ACPI implementation is a port from `http://forum.osdev.org/viewtopic.php?t=16990`

fanoI:
The HAL code has the problem to be practically written in C not in C# (all static methods, totally unsafe, return codes instead of exception in case of errors).
If there was a bug there I sincerely do not know where to start :-(

Maybe a second pass of this after having Kernel.Stop() and Kernel.Restart() using ACPI, could be to refactor that code.

valentinbreiz:
@fanol or halt the system if ACPI is unavailable ?

fanoI:
I'd say you can use the Power class you have written as a "bridge" between the world pre and post ACPI, the static class ACPI should give you have the possibility to know if ACPI is supported.

Then in your Power.Shutdwon you will do:

```cs
        public static void Shutdown()
        {
            if (ACPI.Supported) {
                ACPI.Shutdown();
                ACPI.Disable();
                // calling Core.Global.CPU.Halt() should not be needed the CPU will be already powered off here
            }
            else
                   Core.Global.CPU.Shutdown(); 
        }
```

Reboot will be similar:

```cs
        public static void Reboot()
        {
             if (ACPI.Supported) {
                     ACPI.Reboot();
            }
            else {
                    Core.Global.CPU.Reboot(); 
           }
        }
```

I'm a little unsure if this Power class should be in System making use of ACPI / x86 instructions or if this would be better in HAL, so one rewrite it with the specif modes one does Shutdown / Reboot in other architectures. In this way the only things the user can use to Shutdown / Reboot the kernel are into the Kernel class.

What do you thing @jp2masa?

valentinbreiz:
@fanoI in Core.Global.CPU I must use the same methods as in the ACPI class ?

jp2masa:
I think that the CPU class shouldn't have a Shutdown methods, as a CPU can't shutdown. When ACPI isn't supported, shutdown would be writing on the screen some message saying the system can be powered off and disable interrupts and halt or simply do a Kernel.Stop(). I also think that the Power methods shouldn't be in the Kernel, they should be called from the Power class. The Power implementation should be implemented in HAL.

KL4932:
@jp2masa i need you in gitter, something really important

valentinbreiz:
@jp2masa "in the Kernel" you said in Cosmos.System ?

jp2masa:
Cosmos.System.Kernel class.

fanoI:
@jp2masa in reality was what I meant:

1. In the Kernel class in System there are the methods Stop() and Restart() as now
2. Stop calls Power.Shutdown
3. Reboot calls Power.Reboot

The Power class will be in HAL and will use ACPI or not to implement its methods.

jp2masa:
I think that we should remove Kernel.Restart and make Kernel.Stop internal, and provide a Power class in Cosmos.System that calls the methods from HAL.

fanoI:
I don't understand well... it ones needs to do Power.Shutdown() to stop the system how could be AfterRun() called? I think it could be useful not now for simple kernels but in future.

In a previous discussion with @mterwoord he said that when possible the "interface" with the user level should be in Kernel it is for this that - for example - how set keyboard layout is done and there no "special" keyboard class in System to do this.

jp2masa:
For now, better keep Kernel.Stop() at least, but I think it's not the best way to do this.

valentinbreiz:
Now that I have canceled all the old changes I resume! xD

valentinbreiz:
Too many commits lol, I close this and start with a new branch !

PR #646, opened 2017-05-21
Implemented additional drawing function as per issue #603

nuhash:
Fulfills #603 

Implemented the following functions:
```
public void DrawLine(Pen pen, Point p1, Point p2)
public virtual void DrawCircle(Pen pen, int x_center, int y_center, int radius)
public virtual void DrawEllipse(Pen pen, int x_center, int y_center, int x_radius, int y_radius)
public virtual void DrawPolygon(Pen pen, PointsCollection points)
```
The `DrawLine()` function is just a wrapper on the existing one, it doesn't need to be virtual.

Implemented the following classes:
```
public class Cosmos.System.Graphics.Point
public class Cosmos.System.Graphics.PointsCollection
```

bochs_2017-05-21_23-02-33

### DIFF
--- a/source/Cosmos.System/Graphics/Canvas.cs
+++ b/source/Cosmos.System/Graphics/Canvas.cs
@@ -5,6 +5,8 @@ using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using System.Drawing;
+using System.Collections;
+using System.Collections.ObjectModel;
 
 namespace Cosmos.System.Graphics
 {
@@ -187,6 +189,11 @@ namespace Cosmos.System.Graphics
             DrawDiagonalLine(pen, dx, dy, x1, y1);
         }
 
+        public void DrawLine(Pen pen, Point p1, Point p2)
+        {
+            DrawLine(pen, p1.X, p1.Y, p2.X, p2.Y);
+        }
+
         public void DrawLine(Pen pen, float x1, float y1, float x2, float y2)
         {
             throw new NotImplementedException();
@@ -254,6 +261,15 @@ namespace Cosmos.System.Graphics
                 if (e2 <= dy) { y++; err += dy += a; }
                 if (e2 >= dx || 2 * err > dy) { x--; err += dx += b1; }
             }
+        }
+
+        public void DrawPolygon(Pen pen, PointsCollection points)
+        {
+            for (int i = 0; i < points.Count-1; i++)
+            {
+                DrawLine(pen, points[i], points[i + 1]);
+            }
+            DrawLine(pen, points[0], points[points.Count - 1]);
         }
 
         public void DrawRectangle(Pen pen, int x, int y, int width, int height)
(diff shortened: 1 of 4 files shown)


jp2masa:
If we use Point in the future, we should use System.Drawing.Point, but structs with Sequential layout currently don't work. 
@fanoI What do you think? Should we use Point or keep it as int X and int Y?

fanoI:
Initially the API I add written mimed totally the System.Drawing one and so any method had for implementations for example DrawLine():

```cs
public void DrawLine(Pen pen, int x1, int y1, int x2, int x2);
public void DrawLine(Pen pen, Point p1, Point p2);
public void DrawLine(Pen pen, float x1, float y1, float x2, float x2);
public void DrawLine(Pen pen, Pointf p1, Pointf p2);
``` 
The versions that takes float do exist yet in the code (but not Pointf a point done with float coordinates) but are all not implemented functions... it is not clear to me how to draw a thing in position 42.15 can I draw at sub-pixel positions? Maybe with more "serious" GPU than Bochs one is possible...

We had decided that Point() (and the Rect() type) was really unuseful and so we had removed them from the CGS API.

I think @nuhash has been forced to introduce Point not for DrawLine() but for DrawPolygon() which prototype would have been "impossible" to write otherwise.

In the end if we re-introduce Point I'd say overload for DrawPixel(), DrawRectangle and so on should be added for consinstency.
I've some doubt on the necessity to create a "fake" type as PointsCollection will not be more simple to use an array directly for DrawPoligon as in System.Drawing?

```cs
public virtual void DrawPolygon(Pen pen, Point[] points)
```

jp2masa:
Maybe we can use Tuple<int, int> or int[] pointsX, int[] pointsY, but it would be a bit confusing and I don't know if Tuple works.

fanoI:
A more correct replacement of the Point struct could be a C# 7 ValueTuple that at least permits to give a name to the elements so one could do:

```cs
var point = (X: 42, Y: 42);
```

The DrawPolygon() method should be something strange as:

```cs
public virtual void DrawPolygon(Pen pen, (int X, int Y)[] points)
```
well I prefer Point in the end it seems more clear to me...

jp2masa:
@fanoI In fact, Point[] would be better than PointsCollection.

nuhash:
I added exceptions to the `DrawCircle` and `DrawEllipse` function, it just checks if the 4 axial extremes are on screen. I haven't done it for `DrawPolygon` as its handled by `DrawLine`.

I actually think that it shouldn't throw an exception, there are good reasons to want to draw offscreen (moving an object out of view, hiding the taskbar). It would be better if the offscreen draw instructions were ignored.

fanoI:
Regarding the exceptions when out of screen my idea was that was responsibility of the higher level compositor to do this reasoning that is when the object was partly out of screen only a part of the item should be drawn.

We can rethink this later in any case.

source/Cosmos.System/Graphics/Canvas.cs
```diff
@@ -303,4 +396,30 @@ protected void ThrowIfCoordNotValid(int x, int y)
...
+        }
+        int y;
+        public int Y
+        {
+            get { return y; }
+            set { y = value; }
+        }
+    }
+
+    public class PointsCollection : List<Point>
```
jp2masa: PointsCollection should be removed.

PR #670, opened 2017-06-19
Guid Impl and test

zarlo:


fanoI:
As at any call the GUID generated should be unique (a different number) there is not so much that you can test in the TestRunner I can suggest maybe this two things:

* The returned value is guaranteed to be not equal Guid.Empty so you can test for this
* The returned value should be a byte value (you can convert the returned GUID to a byte Array with ToByteArray() and check that value returned is 16)
* Do another call to NewGuid() and check that is different from the previous one you have got

Tests/Cosmos.Compiler.Tests.Bcl/System/GuidTest.cs
```diff
@@ -0,0 +1,47 @@
+﻿using Sys = System;
```
jp2masa: This isn't needed. There's no Guid type in Cosmos.System, so it can simply be `using System;`.

source/Cosmos.System.Plugs/System/GuidImpl.cs
```diff
@@ -7,9 +7,35 @@ namespace Cosmos.System.Plugs.System
...
+        private static String encode(byte[] byteArray)
+        {
+            string hexBuffer = "";
+            for (int i = 0; i < byteArray.Length; i++)
+                for (int j = 1; j >= 0; j--)
+                    hexBuffer += HEX[(byteArray[i] >> (j * 4)) & 0xF];
+            return hexBuffer;
+        }
+
+        public static Guid Empty = new Guid(0,0,0,0,0,0,0,0,0,0,0);
```
jp2masa: I think that Guid.Empty doesn't need to be plugged.

Tests/Cosmos.Compiler.Tests.Bcl/Kernel.cs
```diff
@@ -44,6 +44,8 @@ protected override void Run()
                 UnsafeCodeTest.Execute();
                 DelegatesTest.Execute();
 
+                //GuidTest.Execute();
```
jp2masa: Why is the test commented? When a test doesn't pass, you shouldn't comment it, but read the error logs and try to fix it.

source/Cosmos.System.Plugs/System/GuidImpl.cs
```diff
@@ -7,9 +7,35 @@ namespace Cosmos.System.Plugs.System
...
         {
-            return new Guid(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0);
+            
+            Random rnd = new Random();
+            byte[] guid = new byte[16];
+            rnd.NextBytes(guid);
+            guid[6] = (byte) (0x40 | ((int) guid[6] & 0xf));
+            guid[8] = (byte) (0x80 | ((int) guid[8] & 0x3f));
+
+            return new Guid(encode(guid));
```
jp2masa: Why are you encoding the byte array to a string? There's a Guid constructor that takes a byte array as parameter.

zarlo:
most of the test is to see if is compliant with RFC 4122

jp2masa:
The problem with the tests is a stack corruption. Guid is a struct so probably it's related to how we handle structs. There are some known issues, and I'm thinking about working on it soon, so better leave this for now.

zarlo:
should i add the RFC4122 test

jp2masa:
Yes.

zarlo:
now it needs a plug -_-

`

04:26:06.568906         Msg: Error: Exception: System.Exception: 
04:26:06.568935         Msg: Native code encountered, plug required.
04:26:06.568936         Msg:   DO NOT REPORT THIS AS A BUG.
04:26:06.568937         Msg:   Please see http://www.gocosmos.org/docs/plugs/missing/
04:26:06.568939         Msg:   Need plug for: System.Int32  Interop+BCrypt.BCryptGenRandom(System.IntPtr, System.Byte*, System.Int32, System.Int32).
04:26:06.568940         Msg:   Called from :
04:26:06.568941         Msg: Interop+BCrypt::Int32 BCryptGenRandom(Byte*, Int32)
04:26:06.568943         Msg: Interop::Void GetRandomBytes(Byte*, Int32)
04:26:06.568944         Msg: System.Random::Int32 GenerateGlobalSeed()
04:26:06.568945         Msg: System.Random::Void .cctor()
04:26:06.568951         Msg: System.Random
04:26:06.568953         Msg: 
04:26:06.568954         Msg: 

`

Arawn-Davies:
I'm guessing System.Random is used?

zarlo:
System.Random is pluged

Arawn-Davies:
Moved to Pull #942

PR #671, opened 2017-06-19
ACPI Implementation (Restart, Shutdown)

valentinbreiz:
This pull request follows the first closed! (#633)


source/Cosmos.HAL/Cosmos.HAL.csproj
```diff
@@ -8,6 +8,10 @@
         <Company>Cosmos</Company>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+      <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
```
jp2masa: Code in HAL should never be unsafe. The ACPI implementation should be separated between Core and HAL.

Arawn-Davies:
This ACPI class does work with Cosmos operating systems, no errors or anything, plus many Cosmos OSes use this same class so thanks for implementing it :)

If AllowUnsafeBlocks is set to false, wouldn't the unsafe class not work? or could one just remove the unsafe bit from the class? so it's just public class ACPI instead of public unsafe class ACPI

Arawn-Davies:
on mine I have unsafe code allowed and keep the class unsafe, but dunno if it wouldn't work in this case

valentinbreiz:
Hello @Arawn-Davies, as @jp2masa said it, Code in HAL should never be unsafe.  So I moved the class to Cosmos.Core and let the AllowUnsafeBlocks to false in Cosmos.HAL

Arawn-Davies:
Ah didn't see jp2masa's comment, mobile github is funky xD

fanoI:
No moving all the class to Core is not the "solution", the idea that idea is that in Core should be only low level stuffs as for example Memory Operations, ACPI is not so much low level to be all in Core!
It should be divided in a HAL part (the majority of the code) and Core port for the real low level stuff that could be not ACPICore but maybe some support class.

For example a mechanism is needed to find in a random block of memory an ASCII string that indicates the start of a struct / table this seems a "trait" of x86 low level code, SMBIOS needed this and if remember correctly VBE need this too, so a specif class could exist or maybe new methods could be added in MemoryBlock.
What do you think @jp2masa?

jp2masa:
I agree. We should have an easy way of reading those tables, but probably the best option would be to use structs. Some low level code (or even unsafe code, I don't know if this is possible in C#) would just assign the table pointer to a struct instance and it could be used safely.
Maybe we should have a method like `ParseTable<T>(IntPtr aTablePointer, ref T aStruct)`.

fanoI:
C# 7 Span<T> has a lot of interesting functionalities to which "interpret" a big bag of bytes in practically anything: https://github.com/dotnet/corefxlab/blob/master/docs/specs/span.md. Probably our MemoryBlock could be a Span<T> too...

I don't think Span<T> could work in Cosmos probably they will require CLR changes.

jp2masa:
I think that Span can be really useful for Cosmos. When we move to .NET Core 2.0, we should make sure it works and use it when we need it. It seems to be specially useful for reading unmanaged arrays (including strings). Maybe our MemoryBlock could be replaced by Span, and we can add extension methods for Span, but for now let's wait for the .NET Core release.

fanoI:
A possible solution before having Span<T> could be to use Marshal for the struct so fixed (that implies unsafe) is not needed anymore for example the RSDPtr could be changed to:

```cs
        public unsafe struct RSDPtr
        {
           [MarshalAs(UnmanagedType.ByValTStr, SizeConst =  8)]
            public String Signature;
            public byte CheckSum;
            [MarshalAs(UnmanagedType.ByValTStr, SizeConst =  6)]
            public String OemID[6];
            public byte Revision;
            public int RsdtAddress;
        };
```

then to do the conversion from "unmanaged memory" (better to use IntPtr in C# that byte *) one must use Marshal.PtrToStructure().

The problem to have a safe method to search this so called "Tables" remains, in reality I think is simply to decide what to do: a new class? This could be part of MemoryBlock?

valentinbreiz:
@jp2masa this pull request can be merged now.

jp2masa:
I'll review it later today.

source/Cosmos.HAL/Cosmos.HAL.csproj
```diff
@@ -8,6 +8,10 @@
         <Company>Cosmos</Company>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
jp2masa: This change should be reverted.

fanoI:
The code to do reboot is not really using ACPI but the keyboard (in some way is the same as if you did CTRL+ALT+CANC!) OsDev says a little on how to do with ACPI sadly:

http://wiki.osdev.org/Reboot

valentinbreiz:
@fanoI yes you're right.. But there is no such resource how to do ACPI reset :/

jp2masa:
ACPI is not only for power management and we will need to implement the full specification later.

fanoI:
The code to do this should be:

```c
outportb(FADT->ResetReg->Address, FADT->ResetValue);
```c

valentinbreiz:
@fanoI, in fact this is the "same thing". According to this site https://clover-wiki.zetam.org/Configuration/ACPI#acpi_resetaddress-and-resetvalue, Address and ResetValue are the same than the PS2 controller reset (0x64/0xFE). But we can do a PCI reset too with 0x0CF9/0x06...

fanoI:
If I've understood well ACPI reboot should "abstract" the fact that PS2 (that could not exist as in a x86 Machintosh) or PCI or whatever is used to do "reboot".
IMHO the legacy reboot using PS2 CTRL+ALT+DEL should be used only as fallback if ACPI is not working (and that should not be in the ACPI code).

The "real" ACPI (that will require a VM to be used) will give us other important information to us for example:
1. Address of legacy peripherals (Floppy, serial port ecc...)
2. The root of PCI bus
3. Power Management features

but it is "future" for now limit yourself to shutdown and reboot.

PR #672, opened 2017-06-21
Improve the NullReferenceException message

jp2masa:
The NullReferenceException message now includes the label name, the document and the line where it occurred, when that information is available.


valentinbreiz:
Why don't merge this @jp2masa ?

PR #674, opened 2017-06-23
Compiler Fixes

jp2masa:
## Changes
- Added tests for number types.
- Fixed convert opcodes.
- Implemented Not and Xor for long and ulong.
- Fixed StringHelper.GetNumberString(int).
- Moved MoveD from XS.SSE2 to XS.


Tests/Cosmos.Compiler.Tests.Bcl/System/ByteTest.cs
```diff
@@ -44,6 +44,115 @@ public static void Execute()
...
+            Assert.IsTrue(val2 == -0x0D, "Byte bitwise not doesn't work got: " + val2);
+
+            val2 = value & 0x06; // low-order bits: val2 = value & 0b0000_0110 = 0b0000_0100
+            Assert.IsTrue(val2 == 0x04, "Byte bitwise and doesn't work got: " + val2);
+
+            val2 = value | 0x06; // low-order bits: val2 = value | 0b0000_0110 = 0b0000_1110
+            Assert.IsTrue(val2 == 0x0E, "Byte bitwise or doesn't work got: " + val2);
+
+            val2 = value ^ 0x06; // low-order bits: val2 = value ^ 0b0000_0110 = 0b0000_1010
+            Assert.IsTrue(val2 == 0x0A, "Byte bitwise xor doesn't work got: " + val2);
```
fanoI: I'm unsure if these test should be part of "Byte" while yes are Byte operations in C# are done all with integers indeed 'val2' is an int variable not a Byte.

source/Cosmos.IL2CPU/IL/Conv_I2.cs
```diff
@@ -24,48 +21,45 @@ public override void Execute(_MethodInfo aMethod, ILOpCode aOpCode)
...
+            }
+            else if (xSourceSize <= 8)
+            {
+                if (xSourceIsFloat)
                 {
                     XS.SSE2.MoveSD(XMM0, ESP, sourceIsIndirect: true);
+                    XS.Add(ESP, 4);
                     XS.SSE2.ConvertSD2SIAndTruncate(EAX, XMM0);
                     XS.Set(ESP, EAX, destinationIsIndirect: true);
                 }
```
fanoI: I preferred the switch here sincerely if the intent of 'xSourceSize <= 4' was that sizes 1 (byte) and 2 (short) should have been comprised I preferred this:

```
case 1:
case 2:
case 4:
   [...]
```
So you avoid to accept size 3 by accident...

 else if (xSourceSize <= 8) is yet less clear we want accept really size 5, 6 and 7? I think not...

IMHO the ideal as now C# 7 support will be to use type pattern matching in all these case so we avoid the annoying xSourceIsFloat tests too:

https://visualstudiomagazine.com/articles/2017/02/01/pattern-matching.aspx

Tests/Cosmos.Compiler.Tests.Bcl/System/ByteTest.cs
jp2masa: The compiler uses stack size 1 for bytes.

source/Cosmos.IL2CPU/IL/Conv_I2.cs
jp2masa: I think that we should optimize code.

fanoI: Well but switch could be more efficient that if / else if / else (in assembler a switch could become a jump table) and I think it was more clear before what was the intent.

jp2masa: Example of Conv_I1 switch with pattern matching: https://gist.github.com/jp2masa/2a4bd6ddf12cf03848eed83688371c50
Is this your suggestion?

fanoI: If we had a sort of enum on the internal types it will be yet more better and clear the intent:

```cs
switch (xSourceType)
{
    case xSourceType.Byte:
        XS.Pop(EAX);
        XS.MoveSignExtend(EAX, AL);
        XS.Push(EAX);
        break;
    case 4 when xSourceIsFloat:
        XS.SSE.MoveSS(XMM0, ESP, sourceIsIndirect: true);
        XS.SSE.ConvertSS2SIAndTruncate(EAX, XMM0);
        XS.MoveSignExtend(EAX, AL);
        XS.Set(ESP, EAX, destinationIsIndirect: true);
        break;
    case 8 when !xSourceIsFloat:
        XS.Pop(EAX);
        XS.Add(ESP, 4);
        XS.MoveSignExtend(EAX, AL);
        XS.Push(EAX);
        break;
    case 8 when xSourceIsFloat:
        XS.SSE2.MoveSD(XMM0, ESP, sourceIsIndirect: true);
        XS.Add(ESP, 4);
        XS.SSE2.ConvertSD2SIAndTruncate(EAX, XMM0);
        XS.Set(ESP, EAX, destinationIsIndirect: true);
        break;
    default:
        throw new NotImplementedException($"Cosmos.IL2CPU.x86->IL->Conv_I1.cs->Error: StackSize ${xSourceSize} for ${(xSourceIsFloat ? "float" : "non float")} not supported");
}
```

jp2masa: The problem is that stack size 1, 2 and 4 are different, as they need to be handled differently in load instructions (stack sizes 1 and 2 need to be extended), so I can't think of a simple way of matching "stack size 1, 2 or 4 and not float".

jp2masa: I updated the gist. I think that having an enum with the internal types would be a bit hard to maintain and it would affect the compiler performance.

fanoI: A thing that has never been clear to me: when we say xSourceSize we are in some way guaranteed that in reality the value is Int32 / UInt32 / IntPtr / UIntPtr? Or a struct as this will pass giving not the expected results:

```cs
stuct ImNotAnInt {
    byte a;
    byte b;
    byte c;
    byte d;
}
```

I'm pretty sure that the .Net Jitter has an Enum corresponding with the types defined in ECMA (that is not always are the same of the ones exposed by C#).
For example they are talking of TYP_CHAR vs TYP_USHORT here: https://github.com/dotnet/coreclr/issues/12983

jp2masa: The stack types are verified by the compiler. For example, for OpNone: https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.IL2CPU/ILOpCodes/OpNone.cs#L519.

fanoI: I've tried the pattern matching code with TryRoslyn it seems the second check "when xSourceIsFloat" is not needed as it does a sort of if (with return) and "implicit" else pretty clever... it uses goto but well is generated code and maybe thing makes thing faster so...
https://goo.gl/b785Vx

jp2masa: As the compiler checks for the types, I think that using if..else if will be much faster and it won't fail. The number of IL opcodes for if..else if is about less 50 than for switch.

PR #690, opened 2017-07-03
Just a little modification

valentinbreiz:
I just change the installation icon for better rendering.


jp2masa:
I don't know if the quality of the image is better or worse.
Original: https://raw.githubusercontent.com/CosmosOS/Cosmos/master/Setup/images/cosmos_small.bmp
Modified: https://raw.githubusercontent.com/valentinbreiz/Cosmos/77c36d84e3020eb8dc9105498f558e554fa67fcc/Setup/images/cosmos_small.bmp
In the original image the background seems to be slightly darker and the border of the letter C is more visible.

valentinbreiz:
@jp2masa it changes the background color. Now in the setup there is just the Cosmos logo and not the grey background..

jp2masa:
In fact, the background should be white, but I'm not sure about the border of the letter C, it shouldn't be changed.

valentinbreiz:
It is not important, it is pretty good, I tried !

jp2masa:
Actually, I think that bmp images don't have transparency, so the background should really be white.

PR #692, opened 2017-07-06
Newobj fix

jp2masa:
When the size of the constructor arguments was bigger than the size of the struct, the copy of the arguments that occurs in newobj would overlap the original arguments.


fanoI:
This should solve the struct boxing issue?

jp2masa:
This fixes structs initialization when the size of the constructor args is bigger than the struct size (i.e. DateTime and TimeSpan). This is part of #553, which I'm dividing into smaller PRs.

PR #700, opened 2017-07-15
PS/2 Driver Work

jp2masa:
## Changes
- Improved the PS/2 driver implementation.
- Added MouseManager.


mterwoord:
Great work, but can you split the PR into a compiler specific one and a feature specific one? Also, please include tests for the compiler changes (with explanation..)

jp2masa:
The compiler change is needed for the whole driver to work. It's just a small change on sign extension, which was being applied on unsigned values, but it shouldn't.

fanoI:
Regarding how KeyboardManager and MouseManager handle the interrupt I'd say that the delegate is better using static variables would not be thread safe.

I've seen that finally there is a Mouse class in System so it is not needed anymore to have a kernel in HAL to use the mouse... but how this class should be used? That is I have to check continually if the variables x, y, MouseState are changed? On the other hand making a user level delegate for this would mean that de facto we will have 2 delegates that call each other.

Could be acceptable in a GUI application to "block" on a function called Mouse.ReadMouse() as in a Console application you will do on Console.ReadKey() / ReadLine()?

jp2masa:
The current implementation of MouseManager and KeyboardManager assume that PS/2 first device is a keyboard and it's connected as well as PS/2 second device is a mouse and it's connected (which is true on virtual machines). We need to change the way that devices are registered, maybe merge part of HAL and System, or we'll need to have classes just to bypass rings, which is not the best option.

fanoI:
Maybe at System level should exist a sort of HIDManager? It could generates messages with a sort of header that indicates what type of device generated the event (keyboard, mouse, touch screen, remote control, ...) and a generic body to indicate what happened. Console.ReadLine() / Console.ReadKey() at that point will use that instead of a List<Char> in which "block".

Probably to do this things correctly thread support is needed.

Arawn-Davies:
IL2CPU has the fix added to it now in the separate repo. The Cosmos.IL2CPU folder in source/ can be deleted as it was an LDarg issue.

@jp2masa  My own fork has the conflicts sorted now in Apollo-OS/Cosmos#2, with project folders updated and all that stuff done. Do you want me to leave it in my fork and see what happens with gen 3 or add it to the ps2-driver branch directly so its ready to merge when the time comes? Although I understand it won't be until the checklist is discussed/done amongst other things.

My fork has the tests and (for my purposes) unnecessary stuff removed so I wouldn't open a PR from Apollo-OS/Cosmos to upstream, as it will affect stuff badly. A simple copy, paste and triple check will do to ensure it goes smoothly.

jp2masa:
I prefer to update `ps2-driver` myself, probably later.

sayaPintar:
mouse manager cant work i try it with tutorial in Wiki cosmos .

PR #704, opened 2017-07-16
ACPI Improvements.

valentinbreiz:
- ACPI Enable and Disable removed from System and HAL
- ACPI Reboot deleted (because it is not a ACPI Reset...)

Now the user has two choice : Sys.Power.Shutdown (that uses ACPI) and Sys.Power.Reboot (that uses CPU).


fanoI:
To make thing more clear ACPI is automatically enabled?

valentinbreiz:
Yes

fanoI:
Excellent!

PR #705, opened 2017-07-16
List Fixes

jp2masa:
Fixed `List<T>.Insert` and `List<T>.RemoveAt`.
## Changes
- Array.Copy fix (it's still not a complete implementation, but at least it checks for array overlapping).
- Added List tests.


fanoI:
Regarding Memcpy() / Memove() I think in future we have to find a way to have a common implementation in Core similar to the one I've done for MemoryOperations.Fill() all version of "memcpy" that are in the C# runtime should call this implementation.
(using SSE for do memcpy() / memove() is surely faster that movs)

I think will be better to add movs to XS so one can XS.movs(...).

PR #713, opened 2017-07-20
Improve Math.Sqrt() implementation.

invisal:
Using Newton Method and a good square root approximation to improve the Math.Sqrt() speed.

### DIFF
--- a/Tests/Cosmos.Compiler.Tests.Bcl/Helper/EqualityHelper.cs
+++ b/Tests/Cosmos.Compiler.Tests.Bcl/Helper/EqualityHelper.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Cosmos.Compiler.Tests.Bcl.Helper
+{
+    class EqualityHelper
+    {
+        public static bool DoublesAreEqual(double left, double right)
+        {
+            // Define the tolerance for variation in their values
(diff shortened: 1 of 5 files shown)


fanoI:
Thanks for you contribution!

To test double "equality" I've created an utility method when I tested Double:
https://github.com/CosmosOS/Cosmos/blob/master/Tests/Cosmos.Compiler.Tests.Bcl/System/DoubleTest.cs#L12

probably this could be put in some common static utility class in the BCL test sources...

invisal:
I have added the all the edge cases.

fanoI:
OK can you use DoublesAreEqual() moving it in a static class (something as EqualityHelper... in other tests there is an ArrayAreEqual() too if I remember well)?

invisal:
Okay. I will try. Right now, I am struggling to run the TestRunner. It always show "error: symbol `DebugStub_SendStackOverflowOccurred' undefined" error message. I will move once I solve the TestRunner first ^^

fanoI:
Ah so you ever executed TestRunner in your machine then?

fanoI:
Minor thing but here should say "Newton" no "Network":

https://github.com/invisal/Cosmos/blob/0cee95b9480b807144bb73383e661428c310150f/source/Cosmos.System_Plugs/System/MathImpl.cs#L412

fanoI:
Thank you for you contribution!

PR #717, opened 2017-07-24
Extended ASCII characters.

valentinbreiz:
I think this is a bad way to do it, but it works ... What do you think @fanoI ?

### DIFF
--- a/source/Cosmos.System/Console.cs
+++ b/source/Cosmos.System/Console.cs
@@ -100,7 +100,522 @@ namespace Cosmos.System {
                     WriteChar(' ');
                     WriteChar(' ');
                     WriteChar(' ');
-                } else {
+                }
+
+                //Extended ASCII Support
+                else if (aText[i] == 'Ç')
+                {
+                    WriteChar((char)128);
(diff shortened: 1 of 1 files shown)


zarlo:
dont we use unicode

fanoI:
C# String are Unicode (UTF-16) yes but the terminal in text mode is limited to the legacy 8 bit code pages so we need to Encode / Decode the text for the terminal.

czhower:
How can I see the changes? I opened it in git desktop but it shows nothing different in that branch. What are the basic changes?

Yes eventually full Unicode is the goal, but its a LOT of work. Extended ASCII gets us Western European with only little work and serves as a good mid term solution.

czhower:
Changes look fine to me for now.

czhower:
Merged

czhower:
A suggestion - an array with lookups would be shorter code and also faster. A byte array of 255 elements.

czhower:
Actually ignore that - as the incoming values are Unicode, it would make the array very large if the code points are scattered.

PR #718, opened 2017-07-27
might make console faster

zarlo:


valentinbreiz:
@zarlo why this will be faster ?

zarlo:
UpdateCursor(); gets called alot less

mterwoord:
Can you:
- rename iWriteChar to something like InternalWriteChar or something like that
- Have WriteChar use InternalWriteChar

zarlo:
will fix the RP when i have the time

fanoI:
A doubt but Console.Write/WriteLine with for example 'è' it is working? I think that the conversion from UFT-16 to CP437 should be done in Write(Char)... what do you think?

jp2masa:
Yes, the conversion should be moved to InternalWriteChar.

PR #734, opened 2017-08-12
mSize in GetVolumes fix.

valentinbreiz:
Now mSize in GetVolumes works and won't return 0.

### DIFF
--- a/source/Cosmos.System2/FileSystem/CosmosVFS.cs
+++ b/source/Cosmos.System2/FileSystem/CosmosVFS.cs
@@ -360,7 +360,7 @@ namespace Cosmos.System.FileSystem
                 switch (FileSystem.GetFileSystemType(mPartitions[i]))
                 {
                     case FileSystemType.FAT:
-                        mFileSystems.Add(new FatFileSystem(mPartitions[i], xRootPath, xSize));
+                        mFileSystems.Add(new FatFileSystem(mPartitions[i], xRootPath, (ulong)xSize));
                         break;
                     default:
                         global::System.Console.WriteLine("Unknown filesystem type!");
--- a/source/Cosmos.System2/FileSystem/CosmosVFS.cs
+++ b/source/Cosmos.System2/FileSystem/CosmosVFS.cs
@@ -360,7 +360,7 @@ namespace Cosmos.System.FileSystem
                 switch (FileSystem.GetFileSystemType(mPartitions[i]))
                 {
                     case FileSystemType.FAT:
-                        mFileSystems.Add(new FatFileSystem(mPartitions[i], xRootPath, (ulong)xSize));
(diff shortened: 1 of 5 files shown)


source/Cosmos.System2/FileSystem/FAT/Listing/FatDiretoryEntry.cs
```diff
@@ -44,7 +44,7 @@ internal class FatDirectoryEntry : DirectoryEntry
             string aFullPath,
             string aName,
             uint aFirstCluster)
-            : base(aFileSystem, aParent, aFullPath, aName, 0, DirectoryEntryTypeEnum.Directory)
+            : base(aFileSystem, aParent, aFullPath, aName, aSize, DirectoryEntryTypeEnum.Directory)
```
jp2masa: `aSize` doesn't exist here.

valentinbreiz: Done.

source/Cosmos.System2/FileSystem/CosmosVFS.cs
```diff
@@ -355,10 +355,12 @@ protected virtual void InitializeFileSystems()
             for (int i = 0; i < mPartitions.Count; i++)
             {
                 string xRootPath = string.Concat(i, VolumeSeparatorChar, DirectorySeparatorChar);
+                ulong Size = mPartitions[i].BlockCount * mPartitions[i].BlockSize / 1024 / 1024;
```
jp2masa: You can make the cast here:
```cs
ulong Size = (long)(mPartitions[i].BlockCount * mPartitions[i].BlockSize / 1024 / 1024);
```

fanoI:
In reality I'm not sure if make a public property return an unsigned type is so much correct... it will not be anymore CLS compliant (so J#, Javascript.net and IronPython for example could not "consume" it).
In the end long (Int64) is really big and well if a partion is more bigger that 2^64 Byte you have problems in any case.

jp2masa:
2^63 - 1, right?

fanoI:
Yes that is the number: 9,223,372,036,854,775,807 that should be sufficient at least until we reach the hexabyte HDD drives :-)

valentinbreiz:
So is it ready to merge ?

jp2masa:
See my revision. `ulong` should be changed to `long`, I hadn't noticed that DirectoryEntry uses long for size.

valentinbreiz:
I changed that.

jp2masa:
You changed `long` to `ulong`, but you need to revert that change.

valentinbreiz:
Done !

jp2masa:
Can you do this small change too, please? https://github.com/CosmosOS/Cosmos/pull/734/files#r132822920

valentinbreiz:
Done! xD

PR #737, opened 2017-08-14
Added DrawFilledRectangle into CGS demo

Arawn-Davies:
Added DrawFilledRectangle into CGS demo, basically a 715 with required changes so the issue can be closed


Arawn-Davies:
The fix is thanks to ThomasBeHappy, not my code
I simply added the demo which he probably already had done just not committed

@ThomasBeHappy thanks for the fix, I should've merged your PR then added the demo

PR #755, opened 2017-09-01
Fixing WriteAllLines issue

KingLuigi4932:
Since IL2CPU doesn't support arrays as `IEnumerable<T>`
I changed WriteAllLines code
and convert string[] into a string ends with '\r\n' based on windows for creating New Line

### DIFF
--- a/source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
+++ b/source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
@@ -139,6 +139,8 @@ namespace Cosmos.System_Plugs.System.IO
         public static void WriteAllLines(string aFile, string[] contents)
         {
             string text = String.Join(Environment.NewLine, contents);
+            Global.mFileSystemDebugger.SendInternal("Writing contents");
+            Global.mFileSystemDebugger.SendInternal(text);
             WriteAllText(aFile, text);
         }
 
--- a/source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
+++ b/source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
@@ -120,18 +120,25 @@ namespace Cosmos.System_Plugs.System.IO
             return result;
         }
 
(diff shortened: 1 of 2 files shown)


source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
```diff
@@ -119,19 +119,14 @@ public static string[] ReadAllLines(string aFile)
 
             return result;
         }
-
-        public static void WriteAllLines(string aFile, IEnumerable<string> contents)
+
```
jp2masa: Nit: trim whitespace.

source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
```diff
@@ -119,19 +119,14 @@ public static string[] ReadAllLines(string aFile)
...
-            {
-                text = string.Concat(text, line, Environment.NewLine);
-            }
-
+            string text = String.Join(Environment.NewLine, contents);
+            
             Global.mFileSystemDebugger.SendInternal("Writing contents");
             Global.mFileSystemDebugger.SendInternal(text);
-
+
```
jp2masa: Nit: trim whitespace.

KingLuigi4932: Working when trying to use the codes in the kernel :)

KingLuigi4932: Means it works with Writing Test and Reading Test

fanoI: You need to add a test in the FatTestRunner so AppVoyer will compile it in any PR and we known automatically that it did not broke.

TestRunner is easier for you too as now you need to have an external kernel, compile Cosmos anytime, Compile the Kernel... right?

KingLuigi4932: Yes, You mean I Need to Create A Tester in this pull request

PR #756, opened 2017-09-02
Copy Impl Fix

geomtech:
This pull request is ready to be merged.

All works, and I made a method to overwrite file like it already did in .NET


source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
```diff
@@ -162,11 +162,31 @@ public static void WriteAllBytes(string aFile, byte[] aBytes)
 
         public static void Copy(string srcFile, string destFile)
         {
-            using (var xFS = new FileStream(srcFile, FileMode.Open))
+            try
```
jp2masa: The `FileStream` code didn't work? I think it's faster to use `FileStream`.

source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
```diff
@@ -162,11 +162,31 @@ public static void WriteAllBytes(string aFile, byte[] aBytes)
...
+            }
+            catch (IOException ioEx)
+            {
+                throw new IOException("File Copy", ioEx);
+            }
+        }
+
+        public static void Copy(string srcFile, string destFile, bool overwriting)
+        {
+            if (overwriting)
```
jp2masa: If it's not overwriting, and the file doesn't exist, the copy has to be done.

source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
geomtech: No, it doesn't work.

source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
geomtech: Oh right ! I will modify this.

source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
geomtech: And the file created is empty

geomtech:
I made a mistake, I open this pull with my Cosmos master branch lol, i recreate one

PR #759, opened 2017-09-03
Fixed Random Issues

KingLuigi4932:
Fix Random issues
- [x] Fix not generating random numbers
- [x] Fix not generating specified random number from user seeds;
- [ ] Fix generate random seeds for random values on `.ctor(aThis);`


KingLuigi4932:
Fixed an issue that leads to not generate random seed ;(

PR #767, opened 2017-09-14
[CGS] VmWare Support

### LINKED ISSUE #602: [CGS] VmWare support
Luckily it is not so difficult as it could seem as Cosmos has a VMWare driver already:
https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.HAL/Drivers/PCI/Video/VMWareSVGAII.cs

it simply needs a VmWareScreenClass in System to make it usable from the user level!
The one done for Bochs should be used as template:

https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.System/Graphics/VBEScreen.cs

the first objectives should be to overwrite the "stub" Clear method in Canvas to use the hardware Clear implemented in the HAL VmWare driver, then overwrite DrawPoint() with what is needed to make it work with VmWare.

Done this the other methods of CGS should already work as they are all based on DrawPoint() but thing can be done better in VmWare as it has hardware accelerated line and rectangle drawing, to take advantage of them you simply need to overload the methods of the abstract canvas class.

Points to indagate:

1. There is a method in VmWare to get the list of supported modes? If it exists the List returned by getAviableModes() should be filled with real values not with hard coded ones as done for the Bochs driver
2. The VmWare driver permits to return in Text Mode?

KingLuigi4932:
Adding VMWare SVGAII Support for CGS (Cosmos Graphics Subsystem)
Implementing the VMWareSVGAII Screen Class and applying some TODOs

Fix #602

### DIFF
--- a/source/Cosmos.System2/Graphics/FullScreenCanvas.cs
+++ b/source/Cosmos.System2/Graphics/FullScreenCanvas.cs
@@ -42,19 +42,9 @@ namespace Cosmos.System.Graphics
         public static Canvas GetFullScreenCanvas()
         {
             Global.mDebugger.SendInternal($"GetFullScreenCanvas() with default mode");
-            
-            /* Use SVGAII When Exists in PCI */
-            if(SVGAIIExists)
-                videoDevice = VideoDriver.VMWareSVGAIIDriver;
-            
-            if (videoDevice == VideoDriver.VMWareSVGAIIDriver)
-                return MyVideoDriver = new SVGAIIScreen();
-            else if (videoDevice == VideoDriver.VBEDriver)
-                return MyVideoDriver = new VBEScreen();
-
+ 
             /* We have already got a VideoDriver istance simple reset its mode to DefaultGraphicMode */
-            MyVideoDriver.Mode = MyVideoDriver.DefaultGraphicMode;
-            return MyVideoDriver;
+            return GetFullScreenCanvas(MyVideoDriver.DefaultGraphicMode);
         }
     }
 }


KingLuigi4932:
@fanol Finished Requested things +
I didn't understand this

> In future we must start the Video driver during boot and he will do here only the passage in graphic mode.

KingLuigi4932:
Results - Running the GraphicTest
[
image
](url)

Problems: 1366x768 doesn't work fine with Canvas

fanoI:
> I mean our Video Driver already start at booting
but if you're talking about from Booting (After the PCI Device Checks and FAT)
Then why to do that?

Bochs and VmWare Video driver are not started at boot but in the FullScreenCanvas static class, the thing is made started is a sort of VGA text only driver that is used for have the minimal possible to implement the Console class.

In future we should make start the correct VGA driver at boot time probably directly in graphic mode "emulating" the Console not to only to show a penguin at boot :-) but to forget forever all the issues that legacy text mode has (no support for UTF-8, no custom fonts, no graphic mode and so on...).
At that point FullScreenCanvas will need to have simply a method to get "an handle" to running VGA driver but this the future (we rethink of this when we will have G3 kernel working).

Your modification is OK for now a part for a thing: my idea was to avoid that the user could inadvertently create more instances of the VGA driver it was for this that I had a static variable and I did a check for null and only in that case I did "new xxx()", you need to do restore this check again, I've not seen if you have reverted the changes you did in the CGS demo class.

Please ensure that Bochs continues to work too.

KingLuigi4932:
Should i close this?

KingLuigi4932:
@fanoI Now i knew what you said, you mean to start VGA at the boot and run the console by the VGA Driver

fanoI:
No you should not close this.

I clarify what I wanted as change, the original version of FullScreenCanvas had this static variable and you have removed it:

```cs
static private Canvas MyVideoDriver = null;
```

the sense of this variable was that the user could really init Canvas (that is the underlying VGA device) only one time so you should do change GetFullScreenCanvas() in this way:

 ```cs
        public static Canvas GetFullScreenCanvas(Mode mode)
        {
            Global.mDebugger.SendInternal("GetFullScreenCanvas() with mode " + mode);
            
            /* Use SVGAII When Exists in PCI */
            if(SVGAIIExists)
                     videoDevice = VideoDriver.VMWareSVGAIIDriver;

            /* We have not an instance yet of VideoDriver, let's create it */
            if (MyVideoDriver == null) {
                 if (videoDevice == VideoDriver.VMWareSVGAIIDriver)
                      return MyVideoDriver = new SVGAIIScreen(mode);
                 else if (videoDevice == VideoDriver.VBEDriver)
                      return MyVideoDriver = new VBEScreen(mode);
           }

            /* We have already got a VideoDriver instance simple change its mode */
            MyVideoDriver.Mode = mode;
            return MyVideoDriver;
```

Issue #747, opened 2017-08-22
Dot in directories name issue


geomtech:
Hello,

_(We already discuss of that with @jp2masa and @fanoI on gitter.)_

I discover an issue that need to be investigate. 
When we create a directory with a dot in the directory, the dot is replaced by many spaces...
Ex: If I want to create the directory "folder.system", it will be "folder         system".

I discover this when I wanted to make a system to display only directories that are not beginning by a dot (like on Linux and Android by default to not display hidden directories and files.)

The problem is here only with directories, files can have dot in the name, and can begin by a dot. (tested).

This is supported normally by FAT. So this is a problem of Cosmos.
Thanks for support.

PS: Sorry for my english if I made mistakes, I'm french and my English level is medium.

fanoI:
In these days I'm trying to take advantage of the better work that is been done in Net Core regarding System.IO that will permit us to plug things at lower level.

For example this issue seems to be correct in my code using DirectoryInfo at least, Directory.Create() will use practically the same code in the end so I'm fiducious that this will work too in this case.

jp2masa:
This is what should happen, according to the FAT spec:
- If a file or directory name start with a dot, it's a dot entry. This special kind of entries can only be created by the FAT driver when a directory is created. The following cases have to be verified:
    - Currently dot entries are not created by the FAT implementation.
    - When trying to create a directory entry with a name starting with a dot, an exception should be thrown (not sure what's the current behaviour in this case).
- If a file or directory name contains a dot, it has to be a long filename, not sure how it works currently.

geomtech:
Okay, so do you want me to close the issue, because it's not really an issue if it's in the FAT specs ?

jp2masa:
Dots aren't allowed only at the start of the directory entry name. The cases I listed have to be verified/fixed.

fanoI:
Probably the easiest way is in a Windows PC with a FAT32 HDD formatted and a C# program that make create any file you want... and see where it fails. Maybe a VM is the best way...

PR #775, opened 2017-09-24
More work on FAT / System.IO

fanoI:
The objective of this PR is to leverage the fact that in Net Core a lot of code that we needed to plug for Net Framework is not needed anymore there is Managed Code or a low level class called "FileSystem" in which all platform dependent code is concentrated.

- [x]  Introduces new plug CosmosFileSystem that plugs any instance of the FileSystem class
- [x]  Add DirectoryInfo support
- [x]  Remove unneeded plugs from the Directory class (it should be all managed code calling FileSystem class)
- [x]  Add FileInfo support
- [x] Remove unneeded plugs from File class (it should be all managed code calling StreamWriter / StreamReader classes)
- [x] Plug StreamWriter
- [x] Plug StreamReader
  
Blocked by https://github.com/CosmosOS/IL2CPU/pull/32


source/Cosmos.System2_Plugs/System/Runtime/CompilerServicesImpl.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Cosmos.IL2CPU.API.Attribs;
+
+namespace Cosmos.System_Plugs.System.Runtime
+{
+    [Plug(Target = typeof(global::System.Runtime.CompilerServices.RuntimeHelpers))]
+    public static class RuntimeHelpersImpl
+    {
+        /* For RAT in Cosmos the size of Reference Types is 8 on x86 / 32 bit too...
+         * String object header has two reference types so the data part is at offset 16
```
jp2masa: This comment is not totally correct.
1. References are: `(currentAddress << sizeof(IntPtr)) | nextAddress`, where `currentAddress` and `nextAddress` are `IntPtr`.
2. The data part starts at offset 12, but the first char is at offset 16, which is what this property returns. From the comment in CoreCLR (notice that it skips over the string length, which is the first 4 bytes after the object header):
>// Number of bytes from the address pointed to by a reference to
// a String to the first 16-bit character in the String.  Skip 
// over the MethodTable pointer, & String 
// length.

geomtech:
You said that DirectoryInfo will fixe https://github.com/CosmosOS/Cosmos/issues/747, so it will be fixed with this pull request ?

jp2masa:
This pull request improves the support of Cosmos for file system methods in the BCL. #747 is a FAT bug.

fanoI:
If I remember well "folder.system" works correctly in this PR but not ".folder".

jp2masa:
That's because the current FAT implementation considers entries that start with a dot as dot entries. I'm not sure about the correct behaviour, but that needs to be reviewed.

fanoI:
It is in any case not object of this PR that is more use more of the Net Core managed code possible.

source/Cosmos.System2_Plugs/System/IO/PathImpl.cs
```diff
@@ -242,7 +242,7 @@ public static string GetTempFileName()
 
         public static string GetTempPath()
         {
-            return @"\Temp";
+            return @"0:\Temp";
```
jp2masa: Why is this change needed?

source/Cosmos.System2_Plugs/System/IO/FileInfoImpl.cs
```diff
@@ -0,0 +1,21 @@
...
+using IL2CPU.API.Attribs;
+using Cosmos.System;
+using Cosmos.System.FileSystem.VFS;
+
+namespace Cosmos.System2_Plugs.System.IO
+{
+    [Plug(Target = typeof(FileInfo))]
+    public static class FileInfoImpl
+    {
+        /* Optimize this: CosmosVFS should expose an attribute without the need to open the file for reading... */
```
jp2masa: I think that the FAT driver has support for entry sizes. The VFS has to provide a way of getting the size, but it doesn't need to be done in this PR.

source/Cosmos.System2_Plugs/System/IO/FileImpl.cs
```diff
@@ -16,205 +16,32 @@ namespace Cosmos.System_Plugs.System.IO
...
-            }
+            Global.mFileSystemDebugger.SendInternal("Writing contents");
 
-            var xEntry = VFSManager.CreateFile(aFile);
-            if (xEntry == null)
+            using (var xSW = new StreamWriter(aFile))
             {
-                return null;
+                foreach (var current in contents)
+                    xSW.WriteLine(current);
```
jp2masa: Nit: please use braces.

source/Cosmos.Core_Plugs/Cosmos.Core_Plugs.csproj
```diff
@@ -22,4 +22,8 @@
     <Folder Include="Properties\" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Include="System\Threading\TaskImpl.cs" />
```
jp2masa: Why is it included as None?

Tests/Cosmos.Compiler.Tests.Bcl/Kernel.cs
```diff
@@ -42,10 +42,11 @@ protected override void Run()
                 BitConverterTest.Execute();
                 UnsafeCodeTest.Execute();
                 DelegatesTest.Execute();
+                EncodingTest.Execute();
                 RandomTests.Execute();
 
-                System.Collections.Generic.ListTest.Execute();
-                System.Collections.Generic.QueueTest.Execute();
+                //System.Collections.Generic.ListTest.Execute();
+                //System.Collections.Generic.QueueTest.Execute();
```
jp2masa: Why are these tests commented?

source/Cosmos.System2_Plugs/System/IO/PathImpl.cs
fanoI: Testing on Windows the value returned is indeed "C:/temp", it made some low level thing not work... Path or Directory I cannot remember correctly.

Tests/Cosmos.Compiler.Tests.Bcl/Kernel.cs
fanoI: I've simply forgotten to re-enable them, I do this now.

source/Cosmos.Core_Plugs/Cosmos.Core_Plugs.csproj
fanoI: I think this has been done automatically by Visual Studio, where should it be?

source/Cosmos.System2_Plugs/System/IO/FileInfoImpl.cs
fanoI: Yes I think so, this can be a "starting point".

source/Cosmos.System2_Plugs/System/IO/PathImpl.cs
jp2masa: The root drive should represent the system drive I think. "0:\\" may not be the root drive, so this needs to be fixed in the future.

source/Cosmos.Core_Plugs/Cosmos.Core_Plugs.csproj
jp2masa: It shouldn't be added at all.

PR #779, opened 2017-09-28
Fixed Cosmos.targets for new user kit

Arawn-Davies:
This fixes the commit with the last userkit where a few lines needed to be added to Cosmos.targets
This pull request fixes this issue by adding it to the source before being compiled into a userkit for a release


Arawn-Davies:
Was originally built with VS/MSBuild version 15.1 and dotnet core 1, however AppVeyor uses 15.3 and netcore 2 which is why appveyor build fails

building with VS 15.1 and .NET Core 1 was successful

PR #788, opened 2017-10-05
Revert "might make console faster"

Arawn-Davies:
Reverts CosmosOS/Cosmos#718

Issues regarding writing to the Console, nothing printed out using Write or WriteLine appears, yet input is still visible, something to do with the Extended ASCII pr, tried fixing to no avail. For now, I think it needs some changes to ensure it works.

### DIFF
--- a/source/Cosmos.System2/Console.cs
+++ b/source/Cosmos.System2/Console.cs
@@ -1,4 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


LegitSoulja:
Don't revert this, as I am fixing the extended ascii pr, basically the entire Console class.

Arawn-Davies:
Ah okay sorry didn't know, please keep me updated on how it goes :)

PR #789, opened 2017-10-06
Implements ACPI.Reboot, removing the thrown NotImplemented exception

Arawn-Davies:
This commit implements ACPI.Reboot();, has been verified to work as I've used it in my Operating System for a long time now. just copied the void from the Power.cs in my repo (https://github.com/Arawn-Davies/Medli/blob/master/Medli.Hardware/Power.cs#L318) and updated the namespace changes in mine.

### DIFF
--- a/source/Cosmos.Core/ACPI.cs
+++ b/source/Cosmos.Core/ACPI.cs
@@ -113,13 +113,22 @@ namespace Cosmos.Core
         // Reboot
         public static void Reboot()
         {
+            //Old Reboot code
+            /*
             byte good = 0x02;
             while ((good & 0x02) != 0)
                 good = Inb(0x64);
             Outb(0x64, 0xFE);
 
             Global.CPU.Halt();
+            CPU.DisableInterrupts();
+            */
+            //Sets the value of the ACPI FADT reset register (0xcf9) to 6 (0x06)
+            Outb(0xcf9, 0x06);
         }
+
+
+        
         public static byte Inb(ushort port)
         {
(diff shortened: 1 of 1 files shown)


fanoI:
It will be interesting to have real ACPI reboot, what do you think?

Arawn-Davies:
oh I thought it was, my bad, I'm guessing this just resets the 8042 controller? 
how can I access the FADT? I think I've got what I need for ACPI reset, 
keeping the Outb void I can do 
```cs 
Outb(FADT->ResetReg->Address, FADT->ResetValue); 
```
although this means finding whether the ResetReg is in the IO bus, PCI bus or memory mapped

Arawn-Davies:
On some systems, the ACPI FACP RESET_REG and RESET_VALUE are set 0xcf9 and 0x06 so gonna give these ago using Outb(0xcf9, 0x06);

fanoI:
You should not hardcode these values but read them from the FADT structures as written on OSDEV.

Arawn-Davies:
Don't they need values assigned to them? Or are they read from the machine when compiled or something

fanoI:
Accessing FADT structures is as if you are reading from BIOS so they should have the correct values.
How to understand if "ResetReg is in the IO bus, PCI bus or memory mapped" is less clear... OsDev left it under a "TODO" :-)

If you want you can ask on their forum how to do it...

Arawn-Davies:
okay, ill also see if the questions already answered too

PR #802, opened 2017-11-07
Updating Syslinux to version 6.03

Arawn-Davies:
Updating the executable in the earlier commit wasn't nearly enough, so this PR includes the required modules needed (i.e. ldlinux.c32, libcom32.c32).

I also added a small snippet in the VMWare VMX to allow for the use of the Cosmos virtual machine  on older processors e.g. Intel Core 2 Duo. Unfortunately VMWare Worktation 14 Pro and Player have dropped support for these older CPUs so this line wouldn't have an effect on 14+, but I've kept it just in case it changes in the future and because it works on VMWare 12/13.

I'm going to add the updated Cosmos.iss and push it to this branch which reflects the added files for USB and ISO, also adding the required files for PXE which might not be working anyway.


Arawn-Davies:
Added the updated .iss!

PR #813, opened 2017-11-30
move wait for VS into inno setup

trivalik:
The builder should not wait until VS is closed, this is work of the Setup itself. Furthermore we check now for Addition applications which are checked by VSIXInstaller.exe before Installation.

### DIFF
--- a/Setup/Cosmos.iss
+++ b/Setup/Cosmos.iss
@@ -153,6 +153,79 @@ Filename: "{app}\Build\Tools\nuget.exe"; Parameters: "sources Remove -Name ""Cos
 Filename: "{app}\Build\Tools\VSIXBootstrapper.exe"; Parameters: "/q /u:Cosmos.VS.ProjectSystem"; StatusMsg: "Removing Visual Studio Extension: Cosmos Project System"
 
 [Code]
+function IsAppRunning(const FileName: string): Boolean;
+var
+  FWMIService: Variant;
+  FSWbemLocator: Variant;
+  FWbemObjectSet: Variant;
+begin
+  Result := false;
+  FSWbemLocator := CreateOleObject('WBEMScripting.SWBEMLocator');
+  FWMIService := FSWbemLocator.ConnectServer('', 'root\CIMV2', '', '');
+  FWbemObjectSet := FWMIService.ExecQuery(Format('SELECT Name FROM Win32_Process Where Name="%s"',[FileName]));
+  Result := (FWbemObjectSet.Count > 0);
+  FWbemObjectSet := Unassigned;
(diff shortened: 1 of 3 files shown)


source/Cosmos.Build.Installer/Paths.cs
```diff
@@ -85,7 +85,7 @@ public static void UpdateVSPath()
 
         if (xCurrentInstance == null)
         {
-          throw new Exception("The Visual Studio instance is invalid!");
+          throw new Exception("The found Visual Studio instances are not the expected Visual Studio 2017!");
```
jp2masa: This exception is thrown when there are more than one VS instances installed, but none of them match the `VSPath` (command line argument). So I think that a better message would be `$"There is more than one Visual Studio instance installed, but none of them matches the specified path '${VSPath}'"`.

source/Cosmos.Build.Builder/CosmosTask.cs
```diff
@@ -120,22 +120,38 @@ public class CosmosTask : Task {
...
-        if (IsRunning("devenv")) {
-          Log.WriteLine("--Visual Studio is running.");
-          Log.WriteLine("--Waiting " + xSeconds + " seconds to see if Visual Studio exits.");
-          // VS doesnt exit right away and user can try devkit again after VS window has closed but is still running.
-          // So we wait a few seconds first.
-          if (WaitForExit("devenv", xSeconds * 1000)) {
-            throw new Exception("Visual Studio is running. Please close it or kill it in task manager.");
-          }
-        }
+    protected void CheckIfVSandCoRunning() {
```
jp2masa: I think that MSBuild and VBCSCompiler prevent the VSIX installation too.

Setup/Cosmos.iss
```diff
@@ -153,6 +153,79 @@ Filename: "{app}\Build\Tools\nuget.exe"; Parameters: "sources Remove -Name ""Cos
...
+function IsAppRunningWithResponse(const FileName: string): Boolean;
+var
+  Retry: Integer;
+begin
+  Result := IsAppRunning(FileName);
+  if Result then
+    MsgBox(FileName  + ' is running. Please close the application before running the installer.', mbError, MB_OK);
+end;
+
+function InitializeSetup: boolean;
```
jp2masa: I think that MSBuild and VBCSCompiler prevent the VSIX installation too.

source/Cosmos.Build.Installer/Paths.cs
Arawn-Davies: Just reworded the exception to say the above and committed to the PR branch, this alright?

PR #823, opened 2017-12-10
Guess demo improvements

jp2masa:
## Changes
- Fixed `Random` plug.
- Guess demo generates a random number and can be played more than once.


KingLuigi4932:
Did you test it?
And shouldn't GenerateGlobalSeed in the System.Private.CoreLib Random?

jp2masa:
Yes, I tested it.

fanoI:
It is not easy probably to make TestRunner test Random(), maybe one could do these tests:

1. Get a Random number in a range
2. Verify that the generated number is in the selected Range
3. Get another Random number in the same range
4. Verify that is NOT the same number of the first Random

jp2masa:
That would cause random test failures.

fanoI:
You fear that 2 subsequent Random numbers could have the same value?

jp2masa:
It's possible, and tests should be deterministic. We can test generating a random number and checking if it's in the given range, which would at least prevent compiler bugs and missing plugs.

fanoI:
OK if tests are added I'd say this could be merged.

PR #826, opened 2017-12-13
Added code style rules

jp2masa:
Please discuss these rules, including their severity.

## Changes
- Added code style rules.


fanoI:
The remotion of "indent_size" what effect has? Visual Studio uses its default value for this?

jp2masa:
It's already specified for all files.

PR #827, opened 2017-12-16
[DONE] a bit of CGS work

zarlo:


zarlo:
dont merge yet there are some more thing i want to add

jp2masa:
When a PR is work in progress, you should add [WIP] before the title.

source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -277,7 +297,14 @@ public virtual void DrawPolygon(Pen pen, params Point[] points)
             DrawLine(pen, points[0], points[points.Length - 1]);
         }
 
-        public virtual void DrawRectangle(Pen pen, int x, int y, int width, int height)
+        public virtual void DrawRectangle(Pen pen, Point point, int width, int height = -1)
```
fanoI: The idea to have height with default value of -1 is to have the possibilty to do "DrawSquare()"? I'm unsure if it was not more clear to have an explicit DrawSquare() at that point.

source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -317,32 +349,97 @@ public virtual void DrawRectangle(Pen pen, int x, int y, int width, int height)
...
+
+        public virtual void DrawArray(Color[] colors, Point point, int width, int height = -1)
+        {
+
+            DrawArray(colors, point.X, point.Y, width, height);
+
         }
-         
+        
+        public virtual void DrawArray(Color[] colors, int x, int y, int width, int height = -1)
```
fanoI: I imagine the idea of this is for draw Bitmaps / Fonts, it is a "naive" implementation as done for Clear() in Canvas, I think will make sense to do something similar to what I've done for Clear() in VBEScreen (this time instead of Fill / Memset() will be something as Memcpy() but the ASM code would similar).

PR #830, opened 2017-12-19
Allows user to change install directory

quajak:
If the user is not admin, this allows the user to change the install place to his user's appdata. This fixes the problem I had.

### DIFF
--- a/Setup/Cosmos.iss
+++ b/Setup/Cosmos.iss
@@ -64,6 +64,7 @@ AlwaysShowComponentsList=False
 ShowComponentSizes=False
 LicenseFile=LICENSE.txt
 DisableDirPage=yes
+PrivilegesRequired=lowest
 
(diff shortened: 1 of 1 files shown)


quajak:
Can afterwards the branch installer-fix be merged with the master?

Arawn-Davies:
Probably via a PR on the CosmosOS/Cosmos repo from this branch to master :)
no issues, errors or unexpected results with this change? along with different deployment options? e.g. PXE, USB (not sure which ones are working atm however)

jp2masa:
This change will allow the user to select a custom path. There's a message pointing that the user should choose AppData\Roaming, so I think it's an acceptable solution for now.

fanoI:
This patch seems OK for me,  @quajak can you "move" it to the Master branch?

PR #833, opened 2017-12-20
Implementation of Thread.Sleep() using IOPort Techinque!

KingLuigi4932:
This is not the true implementation of Thread.Sleep(), but it will help for now :+1:

### DIFF
--- a/source/Cosmos.Core_Plugs/System/Threading/ThreadImpl.cs
+++ b/source/Cosmos.Core_Plugs/System/Threading/ThreadImpl.cs
@@ -1,5 +1,6 @@
+  using System;
 using System.Threading;
-using IL2CPU.API.Attribs;
+using Cosmos.IL2CPU.API.Attribs;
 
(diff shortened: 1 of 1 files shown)


Arawn-Davies:
This should be done really using the PIT method, as CPUs would all operate at different speeds, and we wouldn't want to limit the speed in any way.

There's already works of a Programmable Interval Timer sleep method in the following repo, don't think it's working however so if you could see and check out what's going wrong and use that as an alternative, that'd be great :)

https://github.com/Apollo-OS/AIC-Framework/blob/master/AIC.Core/PIT.cs#L62

jp2masa:
There's a PIT driver in Cosmos too, but it has to be fixed: https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.HAL2/PIT.cs.

KingLuigi4932:
@jp2masa I'll try my best to fix it!

PR #834, opened 2017-12-21
Add Buffer methods to MemoryBlock (Read8, Read16 etc..)

KingLuigi4932:


fanoI:
If the type to Read / Write is short you copy 2 bits at the time so you must divide Size by 2. Analogously for Int but by 4.

KingLuigi4932:
For performance?

fanoI:
it is not for performance but to avoid to read / write more that in the MemoryBlock is!
Think of this example you have a MemoryBlock that contains 8 byte how many read you need for any type to read it all?

1. For byte is really easy: 8
2. For short you cannot do the for using Size because anytime you do the cast you are really reading two bytes so you need half the iterations and - at the same time - an array of x[Size] is too big!
4. For int again any time that you read casting to int read 4 bytes at a time

KingLuigi4932:
@fanol that means it'll skip half of the array elements
every method contains a pointer with the type of the method so for example:
I want to read 2 shorts, in the method pointer will be incermented by (i) and the array will be filled by the short (2 bytes) in the memory,

The method's arguments are the same as the data type
void Read16(ushort[])
void Read32(uint[])

if there was something like Read16(byte[]) then you may be right or that i didn't understand you correctly

fanoI:
It seems OK, but while I was reviewing AHCI I've seen that you have used these method there... I'd say it is better to make this changes here and close this PR.

KingLuigi4932:
@fanol Ok, i'm waiting!

PR #835, opened 2017-12-22
Implementing parsing of strings to floats or doubles

quajak:


fanoI:
Can you add tests here for Double.Parse() / TryParse() and Float.Parse() / TryParse() in these files:

https://github.com/CosmosOS/Cosmos/blob/master/Tests/Cosmos.Compiler.Tests.Bcl/System/DoubleTest.cs
https://github.com/CosmosOS/Cosmos/blob/master/Tests/Cosmos.Compiler.Tests.Bcl/System/SingleTest.cs

source/Cosmos.System2_Plugs/System/MathImpl.cs
```diff
@@ -187,55 +202,163 @@ public static double Cos(double x)
...
+                u_threshold = -7.45133219101941108420e+02,
+                invln2 = 1.44269504088896338700e+00,
+                twom1000 = 9.33263618503218878990e-302,
+                P1 = 1.66666666666666019037e-01,
+                P2 = -2.77777777770155933842e-03,
+                P3 = 6.61375632143793436117e-05,
+                P4 = -1.65339022054652515390e-06,
+                P5 = 4.13813679705723846039e-08,
+                huge = 1.0e+300;
+
```
fanoI: Can you please declare any const double on his separate line?

source/Cosmos.System2_Plugs/System/MathImpl.cs
```diff
@@ -187,55 +202,163 @@ public static double Cos(double x)
...
+                if (hx >= 0x7ff00000)
+                {
+                    Console.WriteLine("Low word! Text");
+                    if (((hx & 0xfffff) | LowWord(x)) != 0) //Assume that __Lo(x) is lower word of x
+                        return x;       /* NaN */
+                    else return (xsb == 0) ? x : 0.0;   /* exp(+-inf)={inf,0} */
+                }
+                if (x > o_threshold) throw new OverflowException(); /* overflow */
+                if (x < u_threshold) return double.MinValue; /* underflow */
+            }
```
fanoI: For C# spec it seems that Math.Exp never throws: https://msdn.microsoft.com/en-us/library/system.math.exp(v=vs.110).aspx in these cases probably is OK to return double.Min and double.Max.

source/Cosmos.System2_Plugs/System/MathImpl.cs
```diff
@@ -187,55 +202,163 @@ public static double Cos(double x)
...
+
+            //Check if non-finite argument
+            if (hx >= 0x40862E42)
+            {           /* if |x|>=709.78... */
+                if (hx >= 0x7ff00000)
+                {
+                    Console.WriteLine("Low word! Text");
+                    if (((hx & 0xfffff) | LowWord(x)) != 0) //Assume that __Lo(x) is lower word of x
+                        return x;       /* NaN */
+                    else return (xsb == 0) ? x : 0.0;   /* exp(+-inf)={inf,0} */
```
fanoI: Can you changes this and other occurences to have the body of the if on the next line?

Tests/Cosmos.Compiler.Tests.Bcl/System/MathTest.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Cosmos.Compiler.Tests.Bcl.System
 {
-    class MathTest
+    internal class MathTest
     {
```
fanoI: "internal" qualifier not needed

Tests/Cosmos.Compiler.Tests.Bcl/System/MathTest.cs
```diff
@@ -35,6 +35,41 @@ public static void Execute()
...
+            //Test with integer
+            result = Math.Exp(2);
+            Assert.IsTrue((result == 7.38905609893065), "e^2 is equal to 7.38905609893065");
+
+            //Test with double exponent
+            result = Math.Exp(1.5);
+            Assert.IsTrue(EqualityHelper.DoublesAreEqual(result, 4.48168907033806), "e^1.5 returns correct result");
+
+            #endregion Math.Exp
+
```
fanoI: Please add tests for "corner" cases to exp is 0, 1, +/-Infinity and so on...

Tests/Cosmos.Compiler.Tests.Bcl/System/MathTest.cs
```diff
@@ -35,6 +35,41 @@ public static void Execute()
...
+            result = Math.Pow(-2, 2);
+            Assert.IsTrue(EqualityHelper.DoublesAreEqual(result, 4), "Math.Pow gives correct result when raising negative number to even power");
+
+            result = Math.Pow(-2, 3);
+            Assert.IsTrue(EqualityHelper.DoublesAreEqual(result, -8), "Math.Pow gives correct result when raising negative number to odd power");
+
+            //Test with negative power
+            result = Math.Pow(2, -1);
+            Assert.IsTrue(EqualityHelper.DoublesAreEqual(result, 0.5), "Pow gives correct results when handling negative powers");
+
```
fanoI: Add tests with double as base (I don't know 0.5^2) and all the corner cases: base / exponent of 0, 1, +/-Infinity and so on...

source/Cosmos.System2_Plugs/System/MathImpl.cs
quajak: I tried it on Windows, it is giving me Infinity for double.Max and 0 for double.Min. I have implemented that behaviour in the fix.

Tests/Cosmos.Compiler.Tests.Bcl/System/MathTest.cs
jp2masa: Even though `internal` is the default, I think that access modifiers should be explicitly specified.

source/Cosmos.System2_Plugs/System/MathImpl.cs
fanoI: OK we must adhere to C# spec.

PR #844, opened 2017-12-29
Fix Blinking Text on (VGA that can blink)

### LINKED ISSUE #559: Screen flickering
Constant screen flickering in bochs when running my kernel. Does anyone know how to fix this?

KingLuigi4932:
## Fixes:
 - Fixes #559
## Changes:
 - Clear the **seventh bit**(0) in `TextScreen.cs` at method `SetColors(ConsoleColor, ConsoleColor)`.

### Is it the true fix?
**No**, the real fix is to switch to real mode and clear in Mode Control Register (Index 10) the third bit to disable blinking and enable the background bright bit for dark gray and other brighten colors.

 ---
(0) = "Seventh bit is defaultly the blinking bit by the bios, if set to 1: the written text after the `SetColors` method will blink forever, if set to 0: none of the above will happen.

### DIFF
--- a/source/Cosmos.HAL2/TextScreen.cs
+++ b/source/Cosmos.HAL2/TextScreen.cs
@@ -78,7 +78,12 @@ namespace Cosmos.HAL
 
         public override void SetColors(ConsoleColor aForeground, ConsoleColor aBackground)
         {
-            Color = (byte)((byte)(aForeground) | ((byte)(aBackground) << 4));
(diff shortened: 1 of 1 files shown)


valentinbreiz:
Wow :D

fanoI:
The problem that I see is that then how you do enable cursor blinking again if you want it? By default Windows console has blinking cursor while the rest of the text is not blinking how can they do that?

I'm pretty sure that the original DOS console did this too....

PR #863, opened 2018-01-16
Fixed PIT

ewoudje:
I am not sure it is how it should work but it works (i am still learning)


ewoudje:
Oh it isn't working right, something wrong with the delay

PR #867, opened 2018-01-19
PIT Fix

ewoudje:
The PIT is working now!

I can be the 50th contributor 😄

### DIFF
--- a/source/Cosmos.HAL2/PIT.cs
+++ b/source/Cosmos.HAL2/PIT.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using Cosmos.Core;
 
 namespace Cosmos.HAL {
 	/// <summary>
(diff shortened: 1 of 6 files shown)


ewoudje:
Not working when debugging enabled!

- [ ] Fixed

source/Cosmos.HAL2/PIT.cs
```diff
@@ -57,6 +60,12 @@ public class PITTimer : IDisposable {
         public const uint PITDelayNS = 838;
         public bool T0RateGen = false;
 
+	    public PIT()
+	    {
+            INTs.SetIrqHandler(0x00, HandleIRQ);
+	        _T0Countdown = 65535;
```
charlesbetros: Need to set the property here instead of the field.

T0Countdown = 65535;

source/Cosmos.HAL2/PIT.cs
```diff
@@ -167,27 +178,36 @@ public class PITTimer : IDisposable {
...
-            PITTimer hndlr = null;
-            for (int i = ActiveHandlers.Count - 1; i >= 0; i--) {
-                hndlr = ActiveHandlers[i];
+	    private void HandleIRQ(ref INTs.IRQContext aContext)
+	    {
 
-                hndlr.NSRemaining -= T0Delay;
+	        int T0Delay = (int) T0DelyNS;
+	        PITTimer hndlr = null;
+	        for (int i = ActiveHandlers.Count - 1; i >= 0; i--)
```
charlesbetros: Add a line before the for loop to reset the countdown after the irq fires.

T0Countdown = 65535;

PR #875, opened 2018-01-26
Test runner improvements

jp2masa:
## Changes
- Test runner improvements.


fanoI:
Good idea to leverage Serilog instead to write our one.
Looks good to me!

PR #905, opened 2018-02-24
New Implementation for Math.Ceiling

quajak:
The new implementation should be able to be used for numbers larger than an int.


Arawn-Davies:
Have the latest tests been compiling and running fine for you lately?

PR #906, opened 2018-02-25
[CGS] VMWare SVGAII Support

Arawn-Davies:
Built upon previous work done, it's now at a usable state, with support for changing graphics mode whilst running.

image
An example of SVGAII running at 800x600@32, working on Bochs as well (as originally did)

Beforehand, the SVGA PCIDevice was returning null if it didn't exist, whereas now, it still returns null so it needs being worked on, but instead a DeviceExists boolean  is set to false in the event it doesn't exist.


zarlo:
YAY

source/Cosmos.System2/Graphics/SVGAIIScreen.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        public override List<Mode> getAvailableModes()
+        {
+            return new List<Mode>
+            {
+            /* 16-bit Depth Resolutions*/
+
+                /* SD Resolutions */
+                new Mode(320, 200, ColorDepth.ColorDepth16),
+                new Mode(320, 240, ColorDepth.ColorDepth16),
```
fanoI: Regarding getAvailableModes() would be interesting to see if VmWare has a mode to query for this instead of hardcode the values.
This could be object of a successive PR if you prefer.

Arawn-Davies: The VMWare driver doesn't have a hardcoded set of available modes which the user can use, instead it supports an arbitrary range of resolutions depending on the size of the VM window

What may be doable is having a check to tell what size the VM window is at, and resize accordingly, this might be better left to an enhancements PR

Demos/CosmosGraphicSubsystem/Kernel.cs
```diff
@@ -14,9 +14,14 @@ public class Kernel : Sys.Kernel
...
             Console.WriteLine("Cosmos booted successfully. Let's go in Graphic Mode");
+            Console.WriteLine("Specifying custom graphics mode as 800x600@32");
+            // Specify custom graphics mode (800x600 @ 16)
+            Mode start = new Mode(800, 600, ColorDepth.ColorDepth32);
+
+            Console.WriteLine("Here we go!");
+            // Create new instance of FullScreenCanvas, specifying graphics mode
+            canvas = FullScreenCanvas.GetFullScreenCanvas(start);
 
-            /* Get on istance of the Canvas that is all the Screen */
```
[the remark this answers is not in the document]
Arawn-Davies: I've reverted it so it now uses the default graphics mode.

Arawn-Davies:
Ready for merge, now using default graphics mode, but allowing the user to set it
If a new SVGAIIScreen() or VBEScreen() is initialised, then it uses default graphics mode, if the mode is specified using 
```
return new SVGAIIScreen(mode) 
```
or,
```
return new VBEScreen(mode);
```
then it uses the specified mode.

PR #908, opened 2018-02-25
[WIP] Created plugs to make Hashtable work (at least when an object is key).

fanoI:
There are issue however:
1. IL2CPU modifications are needed (see the PR for the list of the PR of IL2CPU to merge to make this usable)
2. Some fix in IL2CPU seems to give problems with Dictionary (VMT problem again?)
3. Hashtable with valuetype as key doesn't work (ContainsKey() always return false)

These IL2CPU PR are needed to test this PR:
https://github.com/CosmosOS/IL2CPU/pull/47 (OK merged)
https://github.com/CosmosOS/IL2CPU/pull/48  (OK merged)
https://github.com/CosmosOS/IL2CPU/issues/37 (OK merged)
https://github.com/CosmosOS/IL2CPU/pull/69 (OK merged)
https://github.com/CosmosOS/IL2CPU/pull/70 (OK merged)


jp2masa:
The folder structure in the BCL tests represents the namespaces, so `Tests/Cosmos.Compiler.Tests.Bcl/System/Collections/Not Generic/HashtableTest.cs` should be moved to `Tests/Cosmos.Compiler.Tests.Bcl/System/Collections/HashtableTest.cs`.

source/Cosmos.System2_Plugs/System/Threading/ThreadImpl.cs
```diff
@@ -1,13 +1,11 @@
...
-
 using Cosmos.HAL;
-
 using IL2CPU.API.Attribs;
 
 namespace Cosmos.System_Plugs.System.Threading
 {
-    [Plug(Target = typeof(Thread))]
+    //[Plug(Target = typeof(Thread))]
+    [Plug(TargetName = "System.Threading.Thread")]
```
jp2masa: Why doesn't `Target = typeof(Thread)` work? If there are different `Thread` classes in different assemblies, you should specify the assembly to make it clear which type is being plugged, like `System.Threading.Thread, System.Private.CoreLib`.

source/Cosmos.System2_Plugs/System/Threading/SpinWaitImpl.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Threading;
+using IL2CPU.API.Attribs;
+
+namespace Cosmos.System_Plugs.System.Threading
+{
+    [Plug(Target = typeof(global::System.Threading.SpinWait))]
```
jp2masa: `[Plug(Target = typeof(SpinWait))]` doesn't work here?

Tests/Cosmos.Compiler.Tests.Bcl/System/Collections/Not Generic/HashtableTest.cs
```diff
@@ -0,0 +1,57 @@
...
+            /* The indexer written in this way should be the same thing of Add("Two", "Two") */
+            h["Two"] = "Two";
+
+            Assert.IsTrue((string)h["Two"] == "Two", "Hashtable indexer failed: existing value (II) not found");
+
+            Assert.IsTrue(h.Count == 2, "Hashtable Count failed: value != 2");
+
+            /*
+             * Got Il2CPU exception:
+             * System.Exception: Original method argument $this is a reference type. Plug attribute first argument is not an argument type, nor was it marked with ObjectPointerAccessAttribute! Method: SystemObjectSystemArrayGetValueSystemInt32 Parameter: aThis
```
jp2masa: The fix for this is simple. Change [this line](https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.Core_Plugs/System/ArrayImpl.cs#L95) to:
```cs
 public static unsafe uint GetValue([ObjectPointerAccess]uint* aThis, int aIndex)
```
It's probably a good idea to apply that change to all methods in `ArrayImpl`.

fanoI:
The HashTable works for when key is an object but continues to not work when is a value type.

There is a very strange problem by the way that is really blocking if HashTable tests are executed Dictionary tests fails with a IL2CPU exception.

fanoI:
With this update the problems with ValueTypes as keys are resolved after https://github.com/CosmosOS/IL2CPU/pull/69 and https://github.com/CosmosOS/IL2CPU/pull/70 it should be ready to be merged.

Issue #601, opened 2017-03-12
[CGS] New Features


fanoI:
This a list of my ideas of what is yet needed to complete CGS:

- [x] CGS should work in VmWare too
- [ ] Add to the Bochs driver the support for 4, 8, 16 and 24 bit color depths
- [x] Add to the Canvas abstract Class methods to draw more shapes for example DrawFilledRectangle(), DrawSquare() and so on
- [x] Add a method to draw an image on CGS for example DrawImage()
- [x] Add a method to draw a string on CGS
- [x] Double buffering (unsure if this must be done better in the Window Manager)
- [x] How to return to text mode (and is this really needed?)

If you have other ideas for now features please let me know in this issue.

AnotherAltr:
>return to text mode (is this really needed?)
Sure.
Will not be superfluous~

fanoI:
Yes I tough this too but guys on OsDev made me think: one time that you have graphic mode working why returning to text mode? To show the Console? But the Console could be done better using GUI:

1.  You don't have the problem of hardcoded fonts (you can use true type fonts)
2.  You don't have the problem of codepages but you can use unicode directly
3.  You can show graphic elements directly (for example a PictureViewer application can show the image in the "Console" directly)
4.   You can have more than 16 colors for the text itself (but here the C# Console methods are limiting us)
5.   Easier to do bold, italic and so on...

If you run Linux an look well during the but phase what happens after the boot loader give the control to Linux? Ever you noticed that the fonts changes? Linux "text mode" itself is NOT text mode :-)

zarlo:
You might want to return to text mode to save RAM and CPU and just have the GUI running then the user needs it.

MichaelTheShifter:
Also I don't think "DrawSquare" would be useful as all a square is is a
rectangle with all sides an equal length.

You could just do DrawRectangle(4,4,4,4) to draw a square at (4,4) with a
size of (4x4) pixels, for example.

fanoI:
Yes DrawSquare() would be only an utility method and obviously would call DrawRectangle()!

@zarlo I don't that with modern PC being in GUI mode occupies so much CPU / RAM in particular if to emulate text mode you will use 640 x 480 @ 16 (or 256) colors.

Arawn-Davies:
[...]

PR #913, opened 2018-02-26
[CGS] CGS Enhancements

Arawn-Davies:
With VMWare support in CGS now here, there are still some things that need doing, 
built upon the list by @fanol in issue #601  

- [ ] Add to the Bochs driver the support for 4, 8, 16 and 24 bit color depths
- [x] [WIP] Add to the Canvas abstract Class methods to draw more shapes for example DrawFilledRectangle(), DrawSquare() and so on (Most shapes are done, some need to be implemented at a driver level (SVGAIIScreen))
- [x] Add a method to draw an image on CGS for example DrawImage() (#918 complete)
- [ ] Add a method to draw a string on CGS
- [ ] Double buffering (Should this be left to the user in a window manager?) - some of this is done at a hardware level
- [x] How to return to text mode
- [x] Update the old VGA code to support CGS
- [ ] Detected VBE through multiboot

If you have other ideas for new features please let me know in this pull request, and if you do some work, add it to a fork and open a PR, checking off an item on this check list


fanoI:
This version of Linux I'm forced to use at work does not permit me to do review anymore, sadly!
One thing the default graphic mode of VmWare was 1024x768@16Bit maybe was for this was not supported? Can you try 1024x768@32Bit?

Arawn-Davies:
Sadly neither 1024x768@32 was supported either, at least they're not in the list of supported modes. Does the list of modes actually list the supported modes or are they just hardcoded as part of Cosmos and need implementing?

fanoI:
The idea was that the device had a list in its memory and you queried it: the real VESA / VBE has provision for a specific command to do so, but Bochs when used in native mode does not support it and now seems neither VmWare. At this point yes you should remove from the list of VmWare SVGA all modes it does not support.

Demos/CosmosGraphicSubsystem/Kernel.cs
```diff
@@ -79,6 +84,11 @@ protected override void Run()
                 pen.Color = Color.Chocolate;
```
fanoI: I don't understand this change maybe VmWare supports 16 bit color, but surely not VBE at least until we do not do the conversion from RGB32 to RGB16.
In theory this mode change should throw for VBE as it is not supported.

Arawn-Davies: I've reverted it in the latest commits, I must've got confused with the deployment profiles
32-bit works fine in both VMWare and Bochs, but only 16-bit with VMWare, and throwing as necessary

Arawn-Davies: For 800x600 anyway - for 1024x768, that needs checking

fanoI:
As we have it already yes, so one could use CGS on real hardware albeit with all the limitation of legacy VGA (640x480@16 colors as max resolution!), simply in system a VGAScreen done similarly to VBEScreen needs to be added.

You would have to change FullScreenCanvas.cs to use VGA only if SVGAII or Bochs VBE cannot be used.

source/Cosmos.System2/Graphics/FullScreenCanvas.cs
```diff
@@ -1,77 +1,99 @@
...
         }
 
         private static Canvas GetVideoDriver(Mode mode)
         {
-            if (SVGAIIExist())
+            if (_SVGAIIDevice != null && PCI.Exists(_SVGAIIDevice))
+            {
+                return new SVGAIICanvas(mode);
+            }
+            else if (BGAExists() || PCI.GetDevice((VendorID)0x80EE, (DeviceID)0xBEEF) != null)
```
valentinbreiz, line 58: VBE is detected for bochs and VirtualBox now, @quajak do you think it has to be detected through Multiboot too? Because VBE VESA can be set by syslinux and the VBE informations are set into multiboot

EliaSulimanov:
@valentinbreiz This force-push deleted a lot documentation work.

valentinbreiz:
indeed. I take a look

valentinbreiz:
Can you show me what has been deleted @EliaSulimanov?

EliaSulimanov:
@valentinbreiz Sure,
Everything of this files:
source/Cosmos.System2/Graphics/Canvas.cs
source/Cosmos.System2/Graphics/FullScreenCanvas.cs
source/Cosmos.System2/Graphics/SVGAIIScreen.cs
source/Cosmos.System2/Graphics/VBEScreen.cs
source/Cosmos.System2/Graphics/VGAScreen.cs

You can see the specific code [here](https://github.com/CosmosOS/Cosmos/commit/dffe58862d61b3bb1f9e3d7228030910901fa039), check for the deletion.

EliaSulimanov:
@valentinbreiz I checked it further, actually Everything in Cosmos.System2/Graphics has been deleted, nothing left at all.

PR #918, opened 2018-02-28
[CGS] Bitmap implementation

quajak:
This PR aims to add a Bitmap and Image class. A bitmap will be able to be loaded and saved as a file. Further they can be drawn by the canvas.

This PR also fixes the bug in File.ReadAllBytes, which was actually a bug in FatStream.Read() and stopped files from correctly being read which were larger than one cluster.


fanoI:
Regarding the size:
> the image size. This is the size of the raw bitmap data; a dummy 0 can be given for BI_RGB bitmaps.

possible you has been so unlucky to get this corner case?
Do you can try your code on Windows?

quajak:
Bitmaps work until a size of about 10x10. After this point the file will exceed 512 bytes and part of the image will be filled with a random color. Therefor this can not be fixed until that bug is resolved.

valentinbreiz:
How do you load the bitmap? With File.ReadAllBytes?

quajak:
Yes

fanoI:
Probably nothing change but can you try to use FileStream?
I suggest two version of CreateBitmap one taking string and one stream, the first one does only open the FileStream, while the other operates really on the Bitmap.

fanoI:
It seems that we have a problem with long division that in x86 must be implemented all in software (!) this causes the problem with File.ReadAllBytes() (the problem in reality is at lower level in FatStream).

jp2masa:
@quajak Try these fixes: CosmosOS/IL2CPU#41.

quajak:
@jp2masa It is still not working

source/Cosmos.System2/Graphics/Bitmap.cs
```diff
@@ -0,0 +1,226 @@
...
+            }
+            Array.Copy(imageData, 0, file, position, imageData.Length);
+            stream.Write(file, 0, file.Length);
+        }
+
+        public static Bitmap CreateBitmap(string path)
+        {
+            #region BMP Header
+
+            Bitmap bitmap;
```
fanoI: I would prefer to have ctor instead of a static method to create a bitmap, we could have:

```cs
Bitmap(String); // Create a Bitmap from a file
Bitmap(byte[]); // Create a Bitmap from a byte array 
Bitmap(Stream); // Create a Bitmap from a Stream
```

the first one could be simply empty doing:
```cs
Bitmap(String path) {
    using (FileStream fs = new FileStream(path, FileMode.Open)) {
           return new Bitmap(fs);
    }
```
Bitmap(byte[]) is similar it creates a Stream using MemoryStream(array), only the last one does really something and it pratically the code you have written between "using" in CreateBitmap().

Tests/Cosmos.Kernel.Tests.Fat/System.IO/FileTest.cs
```diff
@@ -227,23 +228,49 @@ public static void Execute(Debugger mDebugger)
...
+            {
+                if (textBytes[i] != readBytes[i])
+                {
+                    same = false;
+                    line = i;
+                    break;
+                }
+            }
+            Assert.IsTrue(same, "Reading large files works using read all bytes does not work. Error at position: " + line);
+
```
fanoI: This for() can be replaced by the helper method ByteArrayAreEquals() that you can find in the HelperMethods class.

fanoI: Can you do big file test with FileStream Read / Write methods, too?

valentinbreiz:
@quajak you fixed the problem with File.ReadAllBytes? Or did you find an other way to read the bytes of the file?

quajak:
@valentinbreiz I fixed it.

valentinbreiz:
You fixed it in FatStream.cs?

quajak:
Well the test I made runs and bitmaps are working, so I persume so.

valentinbreiz:
@quajak with your fixes on FATStream.cs I have that:
https://www.diffchecker.com/I5lfd2nn
The left one is stored in byte[] and the right one is get with File.ReadAllBytes.

quajak:
I am unable to reproduce the error.

fanoI:
The tests regarding File.ReadAllBytes() if there others problems better to open another issue, we had to fix FatStream problems here to permit the Bitmap to finish correctly but is better to have a PR that does only a thing, it become complex to review it otherwise.

PR #942, opened 2018-04-05
GUID Implementation (PR fix for #670)

Siaranite:
This PR is unfinished, but no work could be done due to it being orphaned and there were conflicts preventing it from being merged, so I've kept the patches, merged them into this fork and resubmitted it for more work to be done

### DIFF
--- a/Tests/Cosmos.Compiler.Tests.Bcl/Kernel.cs
+++ b/Tests/Cosmos.Compiler.Tests.Bcl/Kernel.cs
@@ -64,6 +64,9 @@ namespace Cosmos.Compiler.Tests.Bcl
 
                 // System.Text
                 EncodingTest.Execute();
+				
+				// Disable the failing test for now...
+                //GuidTest.Execute();
 
                 TestController.Completed();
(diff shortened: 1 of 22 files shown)


fanoI:
The problem is causes by IL2CPU problem with structures using sequential layout.

Tests/Cosmos.Compiler.Tests.Bcl/Kernel.cs
```diff
@@ -64,6 +64,9 @@ protected override void Run()
 
                 // System.Text
                 EncodingTest.Execute();
+				
+				// Disable the failing test for now...
+                //GuidTest.Execute();
```
jp2masa: The IL2CPU bug should have been fixed by CosmosOS/IL2CPU#52, try enabling the tests.

valentinbreiz: I tried the test and it's still failing :/
PS: sorry for the merge but I reseted my commit

Issue #590, opened 2017-02-28
Array.Sort: Error ocurred while invoking IL2CPU


ERamaM:
Using Array.Sort gives IL2CPU error

Excerpt of malfunctioning code:

```csharp
            int[] array = new int[3];
            array[0] = 0;
            array[1] = 4;
            array[2] = -1;
            Array.Sort(array);
            for (int i = 0; i < 3; i++)
            {
                Console.WriteLine(array[i]); 
            }
```

mihail-mojsoski:
You used Linq.
Try not to use Linq (by creating your own Sort extension).

quajak:
Did this resolve the error? Can this issue be closed?

MishaTY:
Can this be reproduced?

astromediaonly:
I have similar problem with linq throwing exeption during build 
``Exception: System.Exception: <project folder>\IL2CPU``
for example with this code: ``dictionary.Values.OrderBy(w => w.ZIndex)``
It should be easily reproduced - its probably caused by most of its functions like those I guess. Is linq intentionaly unsupported or its just not supported atm?

MishaTY:
It's just unsupported currently due to missing features in IL2CPU

MishaTY:
Still not fixed:
```
1>Native code encountered, plug required.
1>  DO NOT REPORT THIS AS A BUG.
1>  Please see http://www.gocosmos.org/docs/plugs/missing/
1>  Need plug for: System.Void  System.Reflection.MetadataImport._GetNamespace(System.IntPtr, System.Int32, System.Void**)(Plug Signature: System_Void__System_Reflection_MetadataImport__GetNamespace_System_IntPtr__System_Int32__System_Void___ ).
1>  Static: True
1>  Assembly: System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e
1>  Called from:
1>System.Reflection.MetadataImport::System.MdUtf8String GetNamespace(Int32)
1>System.RuntimeType+RuntimeTypeCache::System.String GetNameSpace()
1>System.RuntimeType::System.String get_Namespace()
1>System.Type::System.String get_Namespace()
1>
1>
1>   at Cosmos.IL2CPU.ILScanner.ScanMethod(MethodBase aMethod, Boolean aIsPlug, String sourceItem) in C:\Source\Cosmos\IL2CPU\source\Cosmos.IL2CPU\ILScanner.cs:line 625
```

[...]

PR #949, opened 2018-04-11
Array Sort Implementation (Issue #590)

midonator1305:
Add an Implementation for `Array.Sort(Array)`


midonator1305:
`System.Enum.InternalCompareTo(System.Object, System.Object)` needs to be plugged

midonator1305:
I’m not sure what’s going wrong but the array class seems a bit broken. Maybe my code isn’t good enough but I don’t have any solution. Hopefully someone else will have the skill to fix this

PR #951, opened 2018-04-14
Fixed TimeSpan and DateTime

### LINKED ISSUE #500: TimeSpan doesn't work.
Problem: 
2>Build FAILED.
Build has been canceled.

Example taken from: https://msdn.microsoft.com/en-us/library/system.timespan(v=vs.110).aspx

Code that doesn't work:
TimeSpan interval = new TimeSpan(2, 14, 18);
Console.WriteLine(interval.ToString());

TestCase(Project): http://dropmefiles.com/XXfuP

jp2masa:
Fixes #500.

## Changes
- Fixed TimeSpan and DateTime.
- Added tests.

Depends on CosmosOS/IL2CPU#54.


source/Cosmos.Core_Plugs/System/DateTimeImpl.cs
```diff
@@ -1,68 +0,0 @@
...
-                BCDtoBIN(raw[0]) //SECOND
-            );
-        }
-
-        // TODO: get timezone
-        public static DateTime get_UtcNow() => get_Now();
-
-        private static int[] GetRawDate()
-        {
-            IOPort p70 = new IOPort(0x70);
```
Arawn-Davies: I think these IOPorts are already in the RTC IOGroup, would it be better to just use those instead of making a new instance of these?

jp2masa: Those are removed lines. I moved the plug to System and it uses RTC from HAL.

Arawn-Davies: Oh right didn't notice :') thanks for pointing that out

PR #952, opened 2018-04-14
Plugged Console.Beep

Arawn-Davies:
Not sure about nuget.exe and nuget.exe.old, it must've upgraded during build.
But anyway, here is Console.Beep plugged, usable from both Cosmos.System.PCSpeaker and System.Console classes. Two enums have also been added to Cosmos.System so users can easily use Notes and Durations when using Beep. These enums also hold the default values used for Console.Beep, which is a frequency of 800 hertz and duration of 200 milliseconds.

(Siaranite is my alternative account which I'm signed into Visual Studio with, which explains why the commit is authored by another account)


Arawn-Davies:
Currently works, but for some reason the beeping doesn't stop, but it should be a simple fix
I'll have to do that tomorrow though

Siaranite:
Now has a Console.Beep() demo with Mary Had a Little Lamb!
All issues sorted

PR #953, opened 2018-04-14
[CGS] Add support for 24 bit bitmaps

quajak:
If other color types should be supported I can also try to add them.

### DIFF
--- a/source/Cosmos.System2/Graphics/Bitmap.cs
+++ b/source/Cosmos.System2/Graphics/Bitmap.cs
@@ -52,121 +52,116 @@ namespace Cosmos.System.Graphics
 
         private void CreateBitmap(Stream stream)
         {
-            using (stream)
+            #region BMP Header
+
+            Byte[] _int = new byte[4];
+            Byte[] _short = new byte[2];
+            //Assume that we are using the BMP (Windows) V3 header format
+
+            //reading magic number to identify if BMP file (BM as string - 42 4D as Hex) - bytes 0 -> 2
+            stream.Read(_short, 0, 2);
+            if (0x424d != BitConverter.ToInt16(_short, 0))
+                throw new Exception("Header is not from a BMP");
+
+            //read size of BMP file - byte 2 -> 6
+            stream.Read(_int, 0, 4);
+            uint fileSize = (uint)BitConverter.ToInt32(_int, 0);
+
+            stream.Position = 10;
+            //read header - bytes 10 -> 14 is the offset of the bitmap image data
+            stream.Read(_int, 0, 4);
+            uint pixelTableOffset = (uint)BitConverter.ToInt32(_int, 0);
+
+            //now reading size of BITMAPINFOHEADER should be 40 - bytes 14 -> 18
+            stream.Read(_int, 0, 4);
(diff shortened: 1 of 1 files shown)


source/Cosmos.System2/Graphics/Bitmap.cs
```diff
@@ -52,107 +52,121 @@ public Bitmap(uint Width, uint Height, byte[] pixelData, ColorDepth colorDepth)
...
-            stream.Position = 10;
-            //read header - bytes 10 -> 14 is the offset of the bitmap image data
-            stream.Read(_int, 0, 4);
-            uint pixelTableOffset = (uint)BitConverter.ToInt32(_int, 0);
-
-            //now reading size of BITMAPINFOHEADER should be 40 - bytes 14 -> 18
-            stream.Read(_int, 0, 4);
-            uint infoHeaderSize = (uint)BitConverter.ToInt32(_int, 0);
-            if (infoHeaderSize != 40)
+            using (stream)
```
fanoI: You are already using "using" in the Ctor that takes a file path, so it is not need to write it here.

source/Cosmos.System2/Graphics/Bitmap.cs
```diff
@@ -52,107 +52,121 @@ public Bitmap(uint Width, uint Height, byte[] pixelData, ColorDepth colorDepth)
...
+
+                Byte[] _int = new byte[4];
+                Byte[] _short = new byte[2];
+                //Assume that we are using the BMP (Windows) header format
+                //I am using http://www.fastgraph.com/help/bmp_header_format.html
+                //and https://upload.wikimedia.org/wikipedia/commons/c/c4/BMPfileFormat.png?1519566101894as as reference
+
+                //reading magic number to identify if BMP file (BM as string - 42 4D as Hex) - bytes 0 -> 2
+                stream.Read(_short, 0, 2);
+                if ("42-4D" != BitConverter.ToString(_short))
```
fanoI: It is not needed to convert the "magic number" in string to check if it is correct, it is more efficient to convert '_short' to a proper short and simply do "==". As you are at it create a BMP_MAGIC_NUMBER constant.

source/Cosmos.System2/Graphics/Bitmap.cs
```diff
@@ -52,107 +52,121 @@ public Bitmap(uint Width, uint Height, byte[] pixelData, ColorDepth colorDepth)
...
+                if (planes != 1)
+                    throw new Exception("Number of planes is not 1! Can not read file!");
+                //now reading size of bits per pixel (1, 4, 8, 24, 32) - bytes 28 - 30
+                stream.Read(_short, 0, 2);
+                ushort pixelSize = (ushort)BitConverter.ToInt16(_short, 0);
+                //TODO: Be able to handle other pixel sizes
+                if (!(pixelSize == 32 || pixelSize == 24))
+                {
+                    throw new NotImplementedException("Can only handle 32bit pictures!");
+                }
```
fanoI: As 24 bit pictures are now supported you should change it to "Can only handle 32bit or 24bit pictures!"

Issue #924, opened 2018-03-04
Cosmos DiskManager


fanoI:
The idea of DiskManager is to give the possibility to do the same things you will do with this Windows utility:

![](https://fthmb.tqn.com/FECjEWnjzMEweXyhTl_wZHAqXXw=/768x0/filters:no_upscale()/disk-management-windows-10-58a5d33a3df78c345b052f96.PNG)

(or if you prefer the Unix fdisk utility features)

so we need:

- [x] List Volumes
- [x] Format Volume
- [x] Create Partition
- [x] Set / Get Volume Letter
- [x] Set / Get Volume Label

some of these functionalities are already present in .NET in the class DriveInfo so we will start making that working.

Arawn-Davies:
Could this be done as an abstract class? Seeing as virtual filesystems are easily made with the Filesystem class, would some functions be filesystem type specific or would it just be one disk manager that can do things regardless of the used filesystem?

fanoI:
I'd say only one DiskManager a lot of these things will be simply calls to Cosmos VFSManager that know what is the underlying filesystem the drive has.

TheUnknown-007:
Any updates?

MishaTY:
Implemented in cosmos vfs api

PR #955, opened 2018-04-15
Driveinfo work

fanoI:
This PR adds support for DriveInfo that is the starting point of Cosmos DiskManager (#924).

Another important thing that is changed is how VFS check what filesystem a partition has using a registration mechanism: this open the possibility to create more Filesystem for Cosmos.


source/Cosmos.System2/FileSystem/FileSystem.cs
```diff
@@ -45,5 +35,13 @@ public static FileSystemType GetFileSystemType(Partition aDevice)
         public string mRootPath { get; }
 
         public long mSize { get; }
+
+        public abstract long mAvailableFreeSpace { get; }
```
jp2masa: Property names should be pascal cased (in this case `AvailableFreeSpace`).

fanoI:
In reality the file in which the file is contained is "FatDiretoryEntry.cs" I wanted to change it to "FatDirectoryEntry.cs" but I've noticed that then Github would seen the file totally changed and this would "obstruct" the review. I don't have noticed that instead I've changed the class name there, I'll change it.

Regarding "mAvailableFreeSpace" I think all this class has similar problems for example there are other properties that seem "members" (they start with "m"), maybe better to change them all? For example instead  mSize => Size, mLabel => Label and so on?

jp2masa:
>I wanted to change it to "FatDirectoryEntry.cs" but I've noticed that then Github would seen the file totally changed and this would "obstruct" the review.

In those cases, you can rename the file in a separate commit (no need for a PR).

>maybe better to change them all?

I think so, we should use the standard naming conventions for public APIs.

charlesbetros:
I would verify with Chad regarding proper naming of properties, fields, locals, params. There are some specific rules. Not sure about public properties, but the others have prefixes we follow.

czhower:
Im writing up a doc now. Will post soon regarding naming.

Issue #926, opened 2018-03-06
[Discussion] Make a working PCNETII driver.


valentinbreiz:
I submit this to try to centralize the information in order to find a solution to this problem. 

The driver is located [here](https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.HAL2/Drivers/PCI/Network/AMDPCNetII.cs). 

Since the new PCI implementation me and my team succeed to find and start the AMD PCNETII driver and get some information with that driver like MAC address. But the problem is we can't handle anything, the handler is defined [here](https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.HAL2/Drivers/PCI/Network/AMDPCNetII.cs#L100). But we did many tests and we've never been able to invoke the HandleNetworkInterrupt method.

I think the problem is in the PCI implementation (that maybe blocks the driver initialization) or in the Interrupt system or we're just too stupid to send something to the network card.. 

PS: We did our tests [here](https://github.com/aura-systems/Cosmos/tree/impl-basicnetworking).

fanoI:
Do you have registered the IRQ handler to call your method?
For example look at https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.HAL2/PS2Keyboard.cs#20, you have a similar method in your class but it is commented.

Maybe you should hinerit from NetworkDevice instead of Device:https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.HAL2/NetworkDevice.cs

valentinbreiz:
Yes I registered it like the keyboard. And the NetworkDevice.cs file can be used for multiple drivers but we don't need it without the ring system.

Project-Magenta:
In their test: https://github.com/aura-systems/Cosmos/blob/impl-basicnetworking/source/Cosmos.HAL2/Drivers/PCI/Network/AMDPCNetII.cs#L99

fanoI:
How you indeed send something to the network card? I imagine it has not an IP assigned yet so... how you do this?

valentinbreiz:
I don't know how.

valentinbreiz:
![image](https://user-images.githubusercontent.com/18724279/37257275-7cbaa214-2567-11e8-909a-e5d7c3e16218.png)

@fanoI, PCI scan detected the network card, detected the right interrupt line, read the correct MAC address. But the driver still seems not usable.

fanoI:
But your network card has already an IP? What are Src and Dest?

[...]

PR #974, opened 2018-05-06
PCNETII Driver now working again.

valentinbreiz:
This fixes https://github.com/CosmosOS/Cosmos/issues/926.

### DIFF
--- a/source/Cosmos.Core/PIC.cs
+++ b/source/Cosmos.Core/PIC.cs
@@ -95,8 +95,8 @@ namespace Cosmos.Core
             // set masks:
             Master.Data.Byte = masterMask;
             IOPort.Wait();
-            //Slave.Data.Byte = slaveMask;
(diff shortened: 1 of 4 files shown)


source/Cosmos.Core/PIC.cs
```diff
@@ -95,8 +95,8 @@ private void Remap(byte masterStart, byte masterMask, byte slaveStart, byte slav
             // set masks:
             Master.Data.Byte = masterMask;
             IOPort.Wait();
-            Slave.Data.Byte = slaveMask;
-            IOPort.Wait();
+            //Slave.Data.Byte = slaveMask;
```
jp2masa: Not sure if this is the correct fix. I think that the problem is the mask (passed on line 47). I'm not sure why the mask is set to `0xEF`, but probably there's some readon. Try replacing `0xEF` with `0xEB`.

Issue #965, opened 2018-04-29
Add ATAPI support


fanoI:
So the CD in which Cosmos is running can be seen by Cosmos itself this opens the possibility to install Cosmos on the HDD!

Here you can find some documentation on ATAPI: https://wiki.osdev.org/ATAPI

Arawn-Davies:
I've got started with this, enough to read a single sector, with ejecting the drive [WIP], I'll open a PR :)

MishaTY:
@Arawn-Davies Is it done?

PR #975, opened 2018-05-08
[WIP] ATAPI Drive Support

Arawn-Davies:
This pull request aims to bring support for ATAPI drives, including reading from a specified sector, detecting the capacity of the inserted medium and ejecting of drives (CD Drives), fulfilling issue #965 .

This is nowhere near ready, as they'll be more changes (with the Read_Sector() void and IDE class (MBR to be read with ATA HDDs only if detected?), but it's a start. I've also left comments which review some parts of certain classes , e.g. Extended Boot Records and MBR Partitions as to how it can be updated, for instance

```
foreach (MBR.PartInfo partition in xMBR.Partitions)
```
instead of
```
for (int i = 0; i < xMBR.Partitions.Count; i++)
```

### DIFF
--- a/source/Cosmos.HAL2/BlockDevice/IDE.cs
+++ b/source/Cosmos.HAL2/BlockDevice/IDE.cs
@@ -37,9 +37,9 @@ namespace Cosmos.HAL.BlockDevice
             else if (xATA.DriveType == AtaPio.SpecLevel.ATAPI)
             {
                 /*  Not sure how this would effect the Filesystem - should be fine unless drive master-slave options are changed  */
-                BlockDevice.Devices.Add(xATA);
-                ATAPI.Init();
-                ATAPI.Test();
+                //BlockDevice.Devices.Add(xATA);
+                //ATAPI.Init();
+                //ATAPI.Test();
                 Ata.AtaDebugger.Send("ATA device with speclevel ATAPI found, Support is currently WIP!");
                 return;
             }


source/Cosmos.HAL2/BlockDevice/ATAPI.cs
```diff
@@ -0,0 +1,187 @@
...
+                    Console.Write(DataBuffer[p].ToString() + ",");
+                }
+                else
+                {
+                    Console.WriteLine(DataBuffer[p].ToString());
+                }
+            }
+        }
+
+        public static void ReadBlock(uint lba)
```
fanoI: Maybe it is because is all debug / experimental code but the argument 'lba' is not used by the function?

source/Cosmos.HAL2/BlockDevice/ATAPI.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                DataBuffer[i] = IO.Data.Word;
+                //DataBuffer[i] = Read16(0x1F0);
+            }
+            IRQReceived = false;
+        }
+
+        public override void ReadBlock(ulong aBlockNo, ulong aBlockCount, byte[] aData)
+        {
+            IRQReceived = false;
```
fanoI: Same thing here: aBlockNo, ulong aBlockCount and neither aData are used?

source/Cosmos.HAL2/BlockDevice/ATAPI.cs
Arawn-Davies: It's just experimental code for now, found it in an old driver when I was comparing which one worked and which didn't. It won't be included in the final part.

I was also a bit confused how to specify which block to read from the ATAPI drive (and if a ATAPI drive is a block device?)

https://github.com/CosmosOS/Cosmos/blob/fa4c223452a751f889ff7c53d55c7d8318b9acdb/source/Cosmos.HAL2/BlockDevice/BlockDevice.cs#L40

PR #979, opened 2018-05-13
Update install-VS2017.bat

NanhuGITHUB:
Made The .NET Version 4.7.2.

### DIFF
--- a/install-VS2017.bat
+++ b/install-VS2017.bat
@@ -1,6 +1,6 @@
 @echo off
 
-set DOTNET_VERSION=4.7.1
+set DOTNET_VERSION=4.7.2
 
 echo Check for installed .NET Framework %DOTNET_VERSION%
 


NanhuGITHUB:
Ummm, I have to Do It In other Places. Closing Now.

AridTag:
This is just a note to anyone who stumbles on this like I did and they have:

- Windows 10
- .Net 4.7
- .Net 4.7.2

but they do not have .net 4.7.1 

I found that things seem to work fine if you manually create the registry key
`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\.NETFramework\v4.0.30319\SKUs\.NETFramework,Version=v4.7.1`

jp2masa:
We removed the check in install-VS2017.bat, so this should no longer be an issue. Thanks for the contribution anyway.

PR #980, opened 2018-05-13
Changed some classes of CGS to be struct

fanoI:
Changed some classes of CGS to be struct as they should have been from the beginnning.

- Mode and Point are now structures
- The copy of System.Drawing.Color is not needed anymore the real System.DrawingColor is used instead
- Updated CGS Test Kernel
- Made SVGAII a little more faster (but this not the complete solution)

Needs this ILCPU change before being safe to merge: https://github.com/CosmosOS/IL2CPU/pull/57


source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -429,8 +429,7 @@ protected bool CheckIfModeIsValid(Mode mode)
         {
             Global.mDebugger.SendInternal($"CheckIfModeIsValid");
 
-            if (mode == null)
-                return false;
+            /* Mode has been changed to be a structure so it cannot be null anymore */
```
jp2masa: Nit: I think that this comment doesn't make it easier to understand the code, and may create some confusion too.

source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -452,12 +451,7 @@ protected bool CheckIfModeIsValid(Mode mode)
 
         protected void ThrowIfModeIsNotValid(Mode mode)
         {
-            if (mode == null)
-            {
-                Global.mDebugger.SendInternal($"mode is null raising exception!");
-                throw new ArgumentNullException(nameof(mode));
-            }
-
+            /* Mode has been changed to be a structure so it cannot be null anymore */
```
jp2masa: Nit: I think that this comment doesn't make it easier to understand the code, and may create some confusion too.

source/Cosmos.System2/Graphics/Mode.cs
```diff
@@ -8,30 +8,36 @@ namespace Cosmos.System.Graphics
      */
     //public struct Mode : IEquatable<Mode>, IComparable<Mode>
```
jp2masa: This is not directly related to this PR, but I think that it doesn't make sense to implement `IComparable<Mode>` here. Also, it doesn't seem to be used anywhere.

source/Cosmos.System2/Graphics/Point.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        public Point(int x, int y)
+        {
+            this.x = x;
+            this.y = y;
+        }
+
+        private int x;
+
+        public int X
```
jp2masa: Nit: Maybe use auto-implemented properties here?

```cs
public int X { get; set; }
public int Y { get; set; }
```

PR #981, opened 2018-05-14
Fix DrawImage

jeuxjeux20:
I fixed the errored `DrawImage` method that before accessed the wrong image data colors (`imageY * image.Width` was wrong, it would skip a lot of data and scramble everything)


jeuxjeux20:
Nevermind I forgot how bitmap works lol

PR #990, opened 2018-05-27
Alternative method of GetAmountOfRAM().

KaanGaming:
I added this to Users/KaanGaming folder because the CPU class in the source of Cosmos.Core looks very compiled. So, I didn't want to start a pull request about of CPU class. Instead, created a new folder & created a new file inside it.


zarlo:
files in /users do not get compiled in to cosmos

PR #1013, opened 2018-07-25
Added Int16 and Int32 StackOverflow test

Geramy:
Added Int16 and Int32 StackOverflow test for add and sub.ovf Ticket: IL2CPU/#9 and IL2CPU/#10
https://github.com/CosmosOS/IL2CPU/issues/9


Tests/Cosmos.Compiler.Tests.Bcl/System/Int32Test.cs
```diff
@@ -141,6 +142,32 @@ public static void Execute()
...
+            val3o = 10000;
+            try
+            {
+                val3o += 2147483647;
+            }
+            catch (OverflowException e)
+            {
+                efuse = true;
+            }
+            Assert.IsTrue(efuse == false, "Add_Ovf for Int32 doesn't work");
```
jp2masa: Should be `Assert.IsTrue(efuse, ...)`.

Tests/Cosmos.Compiler.Tests.Bcl/System/Int32Test.cs
```diff
@@ -141,6 +142,32 @@ public static void Execute()
...
+            val3o = -10000;
+            try
+            {
+                val3o -= 2147483647;
+            }
+            catch (OverflowException e)
+            {
+                efuse = true;
+            }
+            Assert.IsTrue(efuse == false, "Sub_Ovf for Int32 doesn't work");
```
jp2masa: Should be `Assert.IsTrue(efuse, ...)`.

Tests/Cosmos.Compiler.Tests.Bcl/System/Int64Test.cs
```diff
@@ -147,6 +148,32 @@ public static void Execute()
...
+            val3o = 1000000;
+            try
+            {
+                val3o += long.MaxValue;
+            }
+            catch (OverflowException e)
+            {
+                efuse = true;
+            }
+            Assert.IsTrue(efuse == false, "Add_Ovf for Int16 doesn't work");
```
jp2masa: Should be `Assert.IsTrue(efuse, ...)`.

Tests/Cosmos.Compiler.Tests.Bcl/System/Int64Test.cs
```diff
@@ -147,6 +148,32 @@ public static void Execute()
...
+            val3o = -10000;
+            try
+            {
+                val3o -= long.MaxValue;
+            }
+            catch (OverflowException e)
+            {
+                efuse = true;
+            }
+            Assert.IsTrue(efuse == false, "Sub_Ovf for Int16 doesn't work");
```
jp2masa: Should be `Assert.IsTrue(efuse, ...)`.

Geramy: Thanks. Now the test fails because it doesn't throw an exception correct?

Geramy: It should be false, because if it doesn't overflow like I want it to then it's not working, i'm a little confused why you think i should do efuse not not == false.

fanoI:
The sub_ovf instruction should be emitted only if the code is "checked" so the test is failing as the normal sub is emitted, you should surround the "-" operation with the checked instruction:

```c#
checked {
      var x = a - b;
}
```

jp2masa:
It makes sure that the arithmetic operations throw on overflow (i.e. emits ovf opcodes). It can be written like this too:
```cs
var x = checked(a - b);
```

PR #1023, opened 2018-08-08
IL2CPU.Reflection

jp2masa:
## Changes
- Fixed plugs, build targets and debug engine.

Depends on CosmosOS/IL2CPU#66.

**WARNING:** Don't merge until the last commit is removed.


source/Cosmos.Core_Asm/Delegate/DelegateInvokeAsm.cs
```diff
@@ -33,7 +35,7 @@ public override void AssembleNew(Assembler aAssembler, object aMethodInfo)
             XS.ClearInterruptFlag();
 
             XS.Comment("Get Invoke list count");
-            var xGetInvocationListMethod = typeof(MulticastDelegate).GetMethod("GetInvocationList");
+            var xGetInvocationListMethod = TypeOf(typeof(MulticastDelegate)).Methods.Single(m => m.Name == "GetInvocationList");
```
mterwoord: Please dont use LINQ here, but a GetMethod(string) thing, like the original. That one oculd them be implemented by dictionary lookup, instead of enumerating the list.

source/Cosmos.Core_Plugs/System/TypeImpl.cs
```diff
@@ -24,16 +24,16 @@ public static Type GetTypeFromHandle(ulong aHandle)
             return xType;
         }
 
-        public static bool op_Equality(CosmosRuntimeType aLeft, CosmosRuntimeType aRight)
+        public static bool op_Equality(Type aLeft, Type aRight)
```
mterwoord: Why is this type changing?

source/Cosmos.VS.DebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
```diff
@@ -131,7 +131,14 @@ int IDebugPendingBreakpoint2.Bind()
 
                             // We have the method. Now find out what Sequence Point it belongs to.
                             var xSPs = xDebugInfo.GetSequencePoints(asm.Pathname, xMethod.MethodToken);
-                            var xSP = xSPs.Single(q => q.LineColStart <= xPos && q.LineColEnd >= xPos);
+                            var xSP = xSPs.Single(
```
mterwoord: Dd the position stuff suddenly change in the debug info database?

source/Kernel-X86/10-CPU/Cosmos.CPU_Asm/Delegate/DelegateCtorAsm.cs
```diff
@@ -16,26 +18,26 @@ public class DelegateCtorAsm : AssemblerMethod {
...
             XS.Comment("-- ldarg 2");
             Ldarg.DoExecute(xAssembler, xMethodInfo, 2);
             XS.Comment("-- stfld _methodPtr");
-            Stfld.DoExecute(xAssembler, xMethodInfo, "System.IntPtr System.Delegate._methodPtr", xMethodInfo.MethodBase.DeclaringType, true, false);
+            Stfld.DoExecute(xAssembler, xMethodInfo, "System.IntPtr System.Delegate._methodPtr", xMethodInfo.MethodInfo.DeclaringType, true, false);
             XS.Comment("Saving ArgSize to field");
             uint xSize = 0;
-            foreach (var xArg in xMethodInfo.MethodBase.DeclaringType.GetMethod("Invoke").GetParameters()) {
-                xSize += ILOp.Align(ILOp.SizeOfType(xArg.ParameterType), 4);
+            foreach (var xArg in xMethodInfo.MethodInfo.DeclaringType.Methods.Single(m => m.Name == "Invoke").ParameterTypes) {
```
mterwoord: Like before, please replace with GetMethod(string) call

source/Cosmos.Core_Asm/Delegate/DelegateInvokeAsm.cs
jp2masa: `GetMethod` right now is a simple `foreach` which returns when it finds a match. It's not good for performance, but I think that we shouldn't think too much about this for now. Also, there shouldn't be many cases where we want to get a specific member by name, so I'm not sure if a `Dictionary<string, MethodInfo>` would make the whole compilation faster...
I also have some plans to improve performance, but for now I tried to keep it simple (although it's a bit slower, which is somehow understandable as the core libraries used to be loaded by the runtime and we were compiling them, and now the compilation is completely isolated).

source/Cosmos.Core_Plugs/System/TypeImpl.cs
jp2masa: Because that's the real signature, no idea why the other one was working, it wasn't supposed to I think.

source/Cosmos.VS.DebugEngine/AD7.Impl/AD7PendingBreakpoint.cs
jp2masa: Nothing changed in the database, I simply removed `SequencePoint` (https://github.com/CosmosOS/IL2CPU/pull/66/commits/44c5a1dd396497239e6dcaf94fbc3c95b82bfc4c#diff-ad3cca2313932e14cee35379036aa0adL325) as there's a new `SequencePoint` class in the new library, but it doesn't have `LineColStart` nor `LineColEnd`. If you want, I can add extension methods to replace this property or create a wrapper for the new `SequencePoint` class which adds these properties.

source/Cosmos.Core_Asm/Delegate/DelegateInvokeAsm.cs
mterwoord: Sure, but having the GetMethod do the foreach means we have it at 1 spot which we can easily optimize later..

PR #1063, opened 2018-10-16
Replaced Array.Length assembler with inline IL

jp2masa:
## Changes
- Replaced Array.Length assembler with inline IL.


MishaTY:
Is this pr needed?

quajak:
Inline IL would make writing certain plugs easier than the current approach, so we need that part of the PR. 
In the case of Array.Length the resulting method would be slower than what we currently have, so I am not sure if its better in this case.

Issue #1027, opened 2018-08-23
Test / Plug more BCL classes


fanoI:
Verify (adding tests to TestRunner) that those classes are working in Cosmos and if not plug the needed methods:

- [x] StringBuilder
- [x] StringReader
- [x] StringWriter
- [ ] XmlReader
- [ ] XmlWriter

quajak:
Both XmlReader and XMlWriter require extensive plugging as they depend on threading and require improvements in IL2CPU so they can not currently be plugged.

valentinbreiz:
Cosmos Gen3 has replaced the IL2CPU-based Gen2 codebase, and `gen3` is now the default branch of this repository. This issue was filed against Gen2, which is no longer developed, so it is being closed as part of the Gen3 cutover (#3088).

If you still hit this on Gen3, please open a fresh issue. The compiler, the build pipeline and the debugging story are all different now, so a new report will get a far more useful answer than reviving this one. Start here: https://cosmosos.github.io/

Gen2 remains on the `gen2` branch, and this issue stays readable.

PR #1080, opened 2018-11-04
Added tests for StringBuilder, StringReader and StringWriter (for #1027)

fanoI:
Added tests for StringBuilder, StringReader and StringWriter (for #1027)

- FAT1 and FAT2 tests are joined in an unique FAT test
- IOTest refactored similarly as BCLTest


Tests/Cosmos.TestRunner.Full/Cosmos.TestRunner.Full.csproj
```diff
@@ -15,7 +15,21 @@
     </ItemGroup>
 
     <ItemGroup>
-        <ProjectReference Include="..\Kernels\**\*.*proj" />
+        <ProjectReference Include="..\Kernels\BoxingTests\BoxingTests.csproj" />
```
jp2masa: Why change this?

fanoI: I've not changed this on purpose, it is possible Visual Studio changed it when I've removed FAT2 Test? If it is so I'm unsure if makes sense that we change it again... it is a generated file.

jp2masa: Project files are not really generated, and they should be minimally readable. This doesn't affect readability that much, so it's fine I think.

Issue #868, opened 2018-01-20
Audit system/hal/core for console output


fanoI:
Low level drivers / modules as VFSManager should never write using Console.Write / Console.WriteLine but at least an example of this can be seen here: https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.System2/FileSystem/VFS/VFSManager.cs#L278

if debug logs are needed, we have the debugger for that.

KrzysztofSzewczyk:
Also, catch clausues should be removed because kernel should handle exceptions. I can pick on this.

Kiirx:
Can be closed

quajak:
Outside of start up, there should be none left

PR #1099, opened 2018-12-26
Upgrade to .NET Core 2.1

jp2masa:
## Changes
- Upgrade to .NET Core 2.1.


Tests/Kernels/Cosmos.Compiler.Tests.Bcl/System/Text/StringBuilderTest.cs
```diff
@@ -38,7 +37,7 @@ public static void Execute()
             sb.Clear();
 
             // Capacity does not change after Clear
-            Assert.IsTrue(sb.Capacity == 32, "StringBuilder.Capacity after Clear is wrong)");
+            Assert.IsTrue(sb.Capacity == 26, "StringBuilder.Capacity after Clear is wrong");
```
fanoI: This is wrong by specifications StringBuilder.Clear() should not change the Capacity property:
https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.clear?view=netframework-4.7.2

but it is not a Cosmos issue really but it seems a Net Core 2.1 bug

Tests/Kernels/Cosmos.Kernel.Tests.Fat/System.IO/DriveInfoTest.cs
```diff
@@ -61,7 +61,8 @@ public static void Execute(Debugger mDebugger)
 
             var xDi = MyDrive.RootDirectory;
 
-            Assert.IsTrue(xDi.Name == MyDrive.Name, "RootDirectory failed");
+            // first doesn't have slash, second does, probably related to path changes in .NET Core 2.1
+            //Assert.IsTrue(xDi.Name == MyDrive.Name, $"RootDirectory failed");
```
fanoI: This is working on a Console Net Core 2.1 application so what could be changed that now this is broken in Cosmos?

source/Cosmos.System2_Plugs/System/IO/CosmosFileSystem.cs
```diff
@@ -27,10 +27,10 @@ public static class CosmosFileSystem
          */
         private static string mCurrentDirectory = string.Empty;
 
-        public static void CreateDirectory(object aThis, string fullPath)
+        public static void CreateDirectory(string fullPath)
```
fanoI: Removing `object aThis` does not break the plug here?

source/Cosmos.System2_Plugs/System/Int16Impl.cs
```diff
@@ -40,20 +43,23 @@ public static Int16 Parse(string s)
...
-                Int16 ind = (Int16)digits.IndexOf(s[i]);
+                var ind = (short)digits.IndexOf(s[i]);
                 if (ind == -1)
                 {
-                    global::System.Console.Write("Digit '");
-                    global::System.Console.Write(s[i]);
-                    global::System.Console.WriteLine("' not found!");
+                    Console.Write("Digit '");
+                    Console.Write(s[i]);
+                    Console.WriteLine("' not found!");
```
fanoI: Better to not write on Console, we should use the debugger for this or better yet as we are throwing after to write in the exception message what went wrong.

source/Cosmos.System2_Plugs/System/IO/CosmosFileSystem.cs
jp2masa: The file systems are no longer objects, it's now a static class.

Tests/Kernels/Cosmos.Kernel.Tests.Fat/System.IO/DriveInfoTest.cs
jp2masa: Probably related to some local change, forgot to take a look at this, will do it now.

Tests/Kernels/Cosmos.Compiler.Tests.Bcl/System/Text/StringBuilderTest.cs
jp2masa: Probably their docs shouldn't be so specific... `Capacity` is always implementation specific, and we shouldn't really rely on it for testing IMO, the only thing we can test is that `Capacity >= Length`.

source/Cosmos.System2_Plugs/System/Int16Impl.cs
jp2masa: I just simplified the names, we should take a look at that later (#868).

Tests/Kernels/Cosmos.Kernel.Tests.Fat/System.IO/DriveInfoTest.cs
jp2masa: Actually, not related to my local changes. I had to plug `PathInternal.IsRoot(ReadOnlySpan<char>)`, and now it's working, I'll push the changes.

fanoI (dismissed):

zarlo:
should this be renamed to Upgrade to .net 50 (and YAY)

AcaiBerii:
:o

quajak (dismissed):

PR #1120, opened 2019-02-07
Added VB test kernel.

wessupermare:
I've added a Visual Basic test kernel and some basic string equality checks (as those were previously failing).

### DIFF
--- a/Test.sln
+++ b/Test.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2010
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28531.58
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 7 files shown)


mterwoord:
Why do we need to manually change files in the obj folder?

wessupermare:
Visual Basic doesn't require namespaces to be qualified. Because Cosmos.System exists in the same namespace tree, the System.... in the <Assembly:> tags is incorrectly assumed to be Cosmos.System. Thus, the Global. qualification is needed before each line. I can't really think of a way around this without putting the VB test kernel in a separate namespace which would probably not go over well with the test kernel.

fanoI:
What does the VB Cosmos template (the guess demo) to solve this problem?
The VB TestRunner works for you?

wessupermare:
I'm not quite sure what you're asking. The TestRunner successfully executes the VB test kernel, if that's what you mean. I haven't made any changes to the VB template itself, just added a test kernel.

quajak:
This should be merged in the .net 5.0 work before that is finished

PR #1148, opened 2019-04-29
Fixes for issues that make Cosmos not compile

VisualDevelopment:
Bumped Inno Setup Version, since the latest one is Inno Setup 6, which makes the builder not recognize the install.


quajak:
I think it makes sense more sense to check for both inno setup 5 and 6, so we can support both versions.

VisualDevelopment:
Hmmm.... Alright

VisualDevelopment:
This fixes issue #1150

source/Cosmos.Build.Builder/Services/InnoSetupService.cs
```diff
@@ -5,25 +5,25 @@ namespace Cosmos.Build.Builder.Services
...
-                    if (key?.GetValue("InstallLocation") is string innoSetupPath)
-                    {
-                        if (Directory.Exists(innoSetupPath))
-                        {
-                            return innoSetupPath;
-                        }
-                    }
-                }
+				for(int i = 5; i < 7; i++){
+					using (var key = localMachineKey32.OpenSubKey(System.String.Format(@"SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Inno Setup {0}_is1", i), false))
```
jp2masa: Nit: I think it would be cleaner to update the const string and use it here.

VisualDevelopment:
What do you mean exactly?

VisualDevelopment:
But okay, i guess i will just update the string

VisualDevelopment:
I don't know why all these files are missing, though

quajak:
Could you try to undo all the changes to the files except InnoSetupService.cs, as they are not needed?

VisualDevelopment:
Yeah, sure, new commit coming in a bit.

VisualDevelopment:
I'm having a problem uhh

VisualDevelopment:
@quajak It says the files are ignored by my .gitignore, which i did not change

VisualDevelopment:
I found a fix

quajak:
Nice. Could you please also revert the changes to BaseIOGroups.cs, Bootstrap.cs and VMWareSVGAII.cs?

VisualDevelopment:
There aren't any changes if you look closely, I don't know why github thinks I did changes.

quajak:
I don't know exactly what changed, most likely spaces were converted to tabs. The easiest way to fix these changes is to revert the two commits and only change that one line again.

VisualDevelopment:
It's telling me the files are the same after overwriting them with freshly pulled ones.

VisualDevelopment:
I'll do your solution then.

VisualDevelopment:
Wait what???
Now more files are deleted???
give me a break man xd

quajak:
Maybe its easier, to close this pr and open a new branch with the single line change and make a new pr.

VisualDevelopment:
I guess i'll do that

PR #1175, opened 2019-05-21
Add a way to create partitions

Nik300:
I'm trying to create a partition in my disk by my Kernel made in Cosmos... the problem is: **THERE ISN'T ANY WAY TO CREATE IT**!!!!


quajak:
Is this a pull request or only an issue? It is currently not possible to partion a disk with cosmos

Issue #584, opened 2017-02-24
Comparer Issue


APERSONA32:
I was trying to come up with some kind of user system but when I tried running the code I got this:

![capture7](https://cloud.githubusercontent.com/assets/24322118/23285294/2638c90e-fa28-11e6-9e46-61ecd0f4eefc.PNG)

The code I'm running is the code on the left hand side.

jp2masa:
IndexOf on List\<T\> isn't implemented in Cosmos, try using a "for" and check item by item.

fanoI:
@jp2masa  I'm asking myself if it would be possible to do a naive plug of IndexOf() that does the for directly so the user code is in C# and when we have a better / more efficient plug they will benefit of this without changing their code.
What do you think?

jp2masa:
I'm waiting for netcore to review plugs, if we still have this problem on netcore better plug IndexOf like you are saying.

fanoI:
Yes I concord this activity must be done after Net Core.

APERSONA32:
I thought I should also add that Contains() also causes the same error.

quajak:
To use IndexOf, Contains and Dictionaries comparers have to be added. Comparers for Int32 and string exist in https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.Core_Plugs/System/Collections/Generic/ComparerHelpersImpl.cs. Adding new comparers is relativly easy and would be done in that file. For help ask in the gitter chat.

disks86:
I'll take a look at this after work tomorrow if nobody else is working on it.

disks86:
The original issue isn't reproducible with the latest code. Regardless I'm working on additional comparers in a fork.

disks86:
@quajak @jp2masa  I've run into an error I can't make sense of. In my local test application Dictionary<string, int> is valid but Dictionary<int, int> gives an IL2CPU error of being unable to compare string and int. The question is where is the string coming from it's key and value are int in that case.

charlesbetros:
What IL opcode is it complaining about?

disks86:
@charlesbetros the opcode was Bne_Un. Which I think is only for int and unordered float.

charlesbetros:
Resolved

PR #1181, opened 2019-06-01
 Comparer Issue #584

disks86:
This PR in combination with a separate PR for IL2CPU resolves issue #584 .

### DIFF
--- a/source/Cosmos.Core_Plugs/System/Collections/Generic/ComparerHelpersImpl.cs
+++ b/source/Cosmos.Core_Plugs/System/Collections/Generic/ComparerHelpersImpl.cs
@@ -26,32 +26,151 @@ namespace Cosmos.Core_Plugs.System.Collections.Generic
         {
             Debugger.DoBochsBreak();
 
+            //TODO: Do application level testing to determine the most frequent comparisons and do those type checks first.
+
+            /*
+             * For now I've made some assumptions about how common each type is.
+             * Strings are very common for controls, messages, console input/output, etc.
+             * Bytes will be used for binary operations and protocols.
+             * Int I believe will be used for return codes and 32 registers (which I expect to be common with x86 and ARM).
+             * Long should be semi-unique identifiers.
+             * I don't except Char to be as common as string but I expect to be used in roughly the same way.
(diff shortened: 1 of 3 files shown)


disks86:
Check won't pass until IL2CPU work-around is merged.

disks86:
The IL2CPU work-around was merged. This just needs to be reviewed and merged to resolve issue #584 .

source/Cosmos.Core_Plugs/System/Collections/Generic/ComparerHelpersImpl.cs
```diff
@@ -1,117 +1,969 @@
...
+    {
+        public override int Compare(char x, char y)
+        {
+            return x.CompareTo(y);
+        }
+    }
+
+    public class NullableCharComparer : Comparer<Char?>
+    {
+        public override int Compare(Char? x, Char? y)
```
quajak: I think the comparer is incorrectly handling nulls, as when only one of the two types is null, the other one is larger, rather than the two being equal.

source/Cosmos.Core_Plugs/System/Collections/Generic/ComparerHelpersImpl.cs
```diff
@@ -1,117 +1,969 @@
...
+                return comparable.CompareTo(y);
+            }
+
+            IComparable comparable2 = y as IComparable;
+            if (comparable2 != null)
+            {
+                return -comparable2.CompareTo(x);
+            }
+
+            throw new NotImplementedException();
```
quajak: Can you please implement the comparisons when the objects are null?

source/Cosmos.Core_Plugs/System/Collections/Generic/ComparerHelpersImpl.cs
disks86: Added in master.

source/Cosmos.Core_Plugs/System/Collections/Generic/ComparerHelpersImpl.cs
disks86: Added in master.

PR #1187, opened 2019-06-12
Update MouseManager.cs

zarlo:
why not most people will not likely do this in there code so why not here


quajak:
Could you please capitalise the method?

PR #1220, opened 2019-09-06
Update to the MainWindowViewModel.cs

rickasheye:
changes to check if the visual studio installation is on the same drive as the installer if not cancel if so continue as without it can cause errors


jp2masa:
The only `Move` is here: https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.Build.Builder/Dependencies/ReposDependency.cs#L76.

It shouldn't be hard to fix that code, something like this should work (replacing lines 75-76):

```cs
zipArchive.ExtractToDirectory(repo.LocalPath);
```

PR #1237, opened 2019-10-20
Feature/cpu info plugs

quajak:


valentinbreiz:
what needs to be done left?

quajak:
AFAIK everything is done. @charlesbetros do you know if anything still needs to be done?

Issue #1333, opened 2020-05-26
CPU Exception when using StrToByteArray


EliaSulimanov:
#### Area of Cosmos - What area of Cosmos are we dealing with?
OS

#### Expected Behaviour - What do you think that should happen?
StrToByteArray should return byte array made out of given string

#### Actual Behaviour - What unexpectedly happens?
CPU exception being thrown on run-time.

#### Reproduction - How did you get this error to appear?
I have created new file named log on root dir.
Then, checked its size and got 0 as expected.
Right after that I tried to init new byte array, using StrToByteArray, this is the code:
`byte[] t = ByteToString.StrToByteArray("Hello World!");`

as this line called, I get CPU exception.
tried to debug this line, as soon as I step into, the error appear.

#### Version - Were you using the User Kit or Dev Kit? And what User Kit version or Dev Kit commit (Cosmos, IL2CPU, X#)?
User Kit 20190628

quajak:
This is a duplicate of #1334

EliaSulimanov:
I think the issue #1334 is unrelated to this issue, as the discussion here is about really specific method causing CPU exception, while there the discussion is about wide range behavior related to the way VS2019 is coping with cpu exception.
The issue in #1334 can be reproduced in other ways, as I mentioned there.
I used the same code to show this 2 issues, but other code apply as well on the issue in #1334

quajak:
The problem here is that`ByteToString.StrToByteArray` is used to convert strings such as `060050` to `3C-32`. For your case look at using the `Encoding` classes.
The CPU Exception in this case is caused by an uncaught exception.

PR #1336, opened 2020-05-27
Added condition on StrToByteArray in order to prevent CPU exception

EliaSulimanov:
Added new condition check on the str that passed to the StrToByteArray function.
This condition check if the passed string is in a numeric format.
Currently there is no check for this, and as byte.Parse is called on the non-numeric string it throws System.FormatException, which is currently seems to be unhandled, causing CPU exception.
This condition should prevent this behaviour.

This problem discussed on #1333 issue.

### DIFF
--- a/source/Cosmos.Common/ByteToString.cs
+++ b/source/Cosmos.Common/ByteToString.cs
@@ -8,7 +8,7 @@ namespace Cosmos.Common
     {
         public static byte[] StrToByteArray(string str)
         {
-            if (str.Length == 0 || !StringHelper.IsNumeric(str))
+            if (str.Length == 0 || !StringHelper.IsNumeric(str) || !StringHelper.IsLengthDivisible(str, 3))
                 throw new Exception("Invalid string value in StrToByteArray");
 
             byte val;
--- a/source/Cosmos.Common/ByteToString.cs
+++ b/source/Cosmos.Common/ByteToString.cs
@@ -8,7 +8,7 @@ namespace Cosmos.Common
     {
         public static byte[] StrToByteArray(string str)
         {
-            if (str.Length == 0)
+            if (str.Length == 0 || !StringHelper.IsNumeric(str))
                 throw new Exception("Invalid string value in StrToByteArray");
 
             byte val;
--- a/source/Cosmos.Common/ByteToString.cs
+++ b/source/Cosmos.Common/ByteToString.cs
@@ -8,7 +8,7 @@ namespace Cosmos.Common
     {
         public static byte[] StrToByteArray(string str)
         {
-            if (str.Length == 0 || !StringHelper.IsNumeric(str) || !StringHelper.IsLengthDivisible(str, 3))
+            if (str.Length == 0 || !StringHelper.IsNumeric(str) || str.Length % 3 != 0)
                 throw new Exception("Invalid string value in StrToByteArray");
 
             byte val;
--- a/source/Cosmos.Common/StringHelper.cs
+++ b/source/Cosmos.Common/StringHelper.cs
@@ -410,6 +410,16 @@ namespace Cosmos.Common {
             }
             return (int)StringComparisonResultEnum.Equal;
         }
+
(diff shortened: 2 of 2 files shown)


source/Cosmos.Common/StringHelper.cs
```diff
@@ -410,6 +410,16 @@ internal enum StringComparisonResultEnum {
...
         }
+
+        /// <summary>
+        /// Check if string is numeric.
+        /// </summary>
+        /// <param name="aString">A string to check if numeric.</param>
+        /// <returns>Returns TRUE if string is numeric.</returns>
+        public static bool IsNumeric(string aString) {
+            int i = 0;
+            return int.TryParse(aString, out i);
```
quajak, line 421: While this test works for short strings, it will not work for arbitrarily long strings.

source/Cosmos.Common/ByteToString.cs
```diff
@@ -8,7 +8,7 @@ static public class ByteToString
     {
         public static byte[] StrToByteArray(string str)
         {
-            if (str.Length == 0)
+            if (str.Length == 0 || !StringHelper.IsNumeric(str))
```
quajak, line 11: Can you also check that the length of the string is divisible by 3

source/Cosmos.Common/StringHelper.cs
EliaSulimanov: Fixed it, changed the method

source/Cosmos.Common/ByteToString.cs
EliaSulimanov: Sure, done

source/Cosmos.Common/StringHelper.cs
```diff
@@ -427,6 +427,18 @@ internal enum StringComparisonResultEnum {
...
+        /// Check if string lenght is divisible by a given number.
+        /// </summary>
+        /// <param name="aString">A string to check if it's length is divisible.</param>
+        /// <param name="aNum">A denominator</param>
+        /// <returns>Returns TRUE if string length divisible by the given number.</returns>
+        public static bool IsLengthDivisible(string aString, int aNum) {
+            if (aNum != 0 && aString.Length % aNum == 0) {
+                return true;
+            }
+            return false;
```
quajak, line 441: I dont think this has to be its own function

source/Cosmos.Common/StringHelper.cs
```diff
@@ -410,6 +410,36 @@ internal enum StringComparisonResultEnum {
...
+
+        /// <summary>
+        /// Check if string is numeric.
+        /// </summary>
+        /// <param name="aString">A string to check if numeric.</param>
+        /// <returns>Returns TRUE if string is numeric.</returns>
+        public static bool IsNumeric(string aString) {
+            for (int i = 0; i < aString.Length; i++) {
+                if (!char.IsDigit(aString[i])) {
+                    if (i == 0 && aString[i].Equals('-')) {
```
quajak, line 422: Bytes should always be positive, so I dont think we have to check for minus.

source/Cosmos.Common/StringHelper.cs
EliaSulimanov: I thought it might be useful for the user in the future.
You think its better to remove it completely and do do this operation in the original function, or leave it as helper function but inline the method in the original?

source/Cosmos.Common/StringHelper.cs
EliaSulimanov: Done

source/Cosmos.Common/StringHelper.cs
EliaSulimanov: Changed it.

quajak: Can you please remove the method? As it can be easily done inline, I dont think we will need it as a helper

EliaSulimanov: Sure

Issue #1334, opened 2020-05-26
CPU exception on VM makes VS2019 crash


EliaSulimanov:
#### Area of Cosmos - What area of Cosmos are we dealing with?
Compiler

#### Expected Behaviour - What do you think that should happen?
CPU exceptions on VM should not make VS2019 crush, it makes the debuging process impossible.

#### Actual Behaviour - What unexpectedly happens?
Any CPU exception that occurre on the VM, makes VS2019 crush and restart, which disconnecting the debuger from the VM

#### Reproduction - How did you get this error to appear?
Used this line, which made the CPU exception to be thrown:
`byte[] t = ByteToString.StrToByteArray("Hello World!");`
and few seconds after the CPU exception is thrown, VS2019 restarts.

#### Version - Were you using the User Kit or Dev Kit? And what User Kit version or Dev Kit commit (Cosmos, IL2CPU, X#)?
User Kit 20190628

quajak:
What is the defintion of `ByteToString.StrToByteArray`? Can you please post the necessary code to reproduce this issue.

EliaSulimanov:
ByteToString.StrToByteArray is an method provided in the Cosmos.Common namespace.
Here is the implementation of it:
https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.Common/ByteToString.cs
starting at line #9.

Here is the kernal code, hope it helps to pour some light on this issue:
`
using System;  
using System.IO;
using Cosmos.System.FileSystem;
using Cosmos.Common;
using Sys = Cosmos.System;

namespace Hello_Cosmos
{
    public class Kernel : Sys.Kernel
    {
        private static CosmosVFS cosmosVFS;
        private static readonly string rootDir = @"0:\";
`

        private bool VFSInit()
        {
            try
            {
                Console.WriteLine("VFS creation");
                cosmosVFS = new CosmosVFS();
                Console.WriteLine("VFS init");
                cosmosVFS.Initialize();
                Console.WriteLine("VFS initialized successfully");
                Sys.FileSystem.VFS.VFSManager.RegisterVFS(cosmosVFS);
                Console.WriteLine("VFS registered successfully");
            }
            catch (Exception e)
            {
                Console.WriteLine("Error occurred in VFS, more info:\n" + e.ToString());
                return false;
            }
            return true;
        }

        protected override void BeforeRun()
        {
            Console.WriteLine("Cosmos booted successfully.");

            if(!VFSInit())
            {
[...]

PR #1338, opened 2020-05-29
Made MessageBox always use UI thread

quajak:
Fixes https://github.com/CosmosOS/Cosmos/issues/1334


EliaSulimanov:
Tried this out and its working very well.
It solves #1334 .
Yet because of #1343 one must restart vs2019 after CPU exception.

PR #1352, opened 2020-06-02
[Done] System2 API docs

EliaSulimanov:
Started working on Cosmos.System namespace API docs

### DIFF
--- a/source/Cosmos.System2/Graphics/FullScreenCanvas.cs
+++ b/source/Cosmos.System2/Graphics/FullScreenCanvas.cs
@@ -1,12 +1,23 @@
 //#define COSMOSDEBUG
 using Cosmos.HAL;
+using sys = System;
 
 namespace Cosmos.System.Graphics
 {
+    /// <summary>
+    /// FullScreenCanvas class. Used to set and get full screen canvas.
+    /// </summary>
     public static class FullScreenCanvas
     {
+        /// <summary>
+        /// SVGA 2 device.
+        /// </summary>
         private static PCIDevice SVGAIIDevice = PCI.GetDevice(VendorID.VMWare, DeviceID.SVGAIIAdapter);
 
+        /// <summary>
+        /// Check if SVGA 2 Exists.
+        /// </summary>
+        /// <returns>bool value.</returns>
         public static bool SVGAIIExist()
         {
             if (SVGAIIDevice == null)
@@ -17,8 +28,16 @@ namespace Cosmos.System.Graphics
             return SVGAIIDevice.DeviceExists;
         }
 
+        /// <summary>
+        /// Video driver.
+        /// </summary>
         private static Canvas MyVideoDriver = null;
 
+        /// <summary>
+        /// Get video driver.
+        /// </summary>
+        /// <returns>Canvas value.</returns>
+        /// <exception cref="sys.ArgumentOutOfRangeException">Thrown if default graphics mode is not suppoted.</exception>
         private static Canvas GetVideoDriver()
         {
             if (SVGAIIExist())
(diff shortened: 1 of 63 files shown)


quajak:
Is this still WIP?

EliaSulimanov:
Yes. Will inform when done.

EliaSulimanov:
Currently done.
Everything is sealed but PCSpeaker and Global, they can only be continued when HAl and Core is documented.

source/Cosmos.System2/Power.cs
```diff
@@ -1,11 +1,23 @@
-﻿namespace Cosmos.System
+﻿using sysIO = System.IO;
```
EliaSulimanov, line 1: Not sure about if this addition can break anything.
would love to hear your thoughts.

source/Cosmos.System2/Graphics/FullScreenCanvas.cs
```diff
@@ -1,12 +1,23 @@
 //#define COSMOSDEBUG
 using Cosmos.HAL;
+using sys = System;
```
EliaSulimanov, line 3: Not sure about if this addition can break anything.
would love to hear your thoughts.

source/Cosmos.System2/Network/ARP/ARPPacket.cs
```diff
@@ -1,17 +1,44 @@
-﻿using Cosmos.HAL;
+﻿using sys = System;
```
EliaSulimanov, line 1: Not sure about if this addition can break anything.
would love to hear your thoughts.

source/Cosmos.System2/Network/IPv4/IPPacket.cs
```diff
@@ -1,14 +1,26 @@
-﻿using Cosmos.HAL.Network;
+﻿using sys = System;
```
EliaSulimanov, line 1: Not sure about if this addition can break anything.
would love to hear your thoughts.

source/Cosmos.System2/Network/IPv4/OutgoingBuffer.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using sys = System;
```
EliaSulimanov, line 1: Not sure about if this addition can break anything.
would love to hear your thoughts.

source/Cosmos.System2/Power.cs
quajak: I dont really get the reason for this addition.

source/Cosmos.System2/Graphics/FullScreenCanvas.cs
quajak: Again I dont really understand why you are adding this?

source/Cosmos.System2/Power.cs
EliaSulimanov: Cosmos.System shadows System, therefore IOException is unknown when documenting it.
I use this reference in line 20:
<exception cref="sysIO.IOException">Thrown on IO error.</exception>.

I can remove this line and metion System.IO.IOException, but its unknown type, as I mentioned.

source/Cosmos.System2/Graphics/FullScreenCanvas.cs
Arawn-Davies: Latest master has this line removed (no System methods are used, everything is in-house so to speak)
Makes no difference whether it's added or removed so it's been left out as it's unlikely that a Console or other System namespace method would be used with the Graphics subsystem (for now anyway...)

Can easily be added at a later stage if needed though.

EliaSulimanov: @Arawn-Davies The reason for this line was for the docs.
I've added it because Cosmos.System shadowed System, and therefore exceptions types were unknown.
Please check if the compiler idenify the exceptions type.
I believe that, for example, in line 62 the exception "sys.ArgumentOutOfRangeException" is of unknown type now.

PR #1354, opened 2020-06-08
Fix up docs project

charlesbetros:


EliaSulimanov:
Would it be able to interact with the docs I'm working on, and parse them to a nice HTML file?

charlesbetros:
That's exactly what it does :)

EliaSulimanov:
Wonderful, I'm really looking forward for it.
If it helps, the docs is generated as XML on build.

PR #1361, opened 2020-06-17
Create Simplified Chinese version of readme. / 添加简体中文版本的README

MasterYuan418:
Simplified Chinese version of readme added


quajak:
I dont  see the reason why we need to translate the readme into different languages.

PR #1373, opened 2020-07-02
VFS article

EliaSulimanov:
It seems that a guide with specific examples can shed a lot of light on the difficulties of Cosmos users.
I added a guide on VFS with some code examples, short explanations and pictures in the hope that it will serve as a useful resource for the users.


valentinbreiz:
Thanks for the work but I think documentation and articles has nothing to do in the project sources. It should be moved to the wiki or the website. @czhower?

EliaSulimanov:
@valentinbreiz You are most welcome.
The documentation is very important, without it in the source, no one can see the purpose of any of the functions.
In big projects like this, not knowing what each function does makes it useless.
More about it: [can be found in this article at dzone.com](https://dzone.com/articles/what-is-api-documentation-and-why-does-it-matter).

The aricle can be moved wherever you see fit, my recommendation is wherever it would be accessible.

quajak:
@EliaSulimanov  Thanks for the work on documentation and I agree that its important. I also agree with @valentinbreiz, I think its better if you can add to and edit the existing documenation in the wiki https://github.com/CosmosOS/Cosmos/wiki. It should be open to you to edit.

valentinbreiz:
Thanks again, I close it now

valentinbreiz:
It's merged into master now, we have made an automatic system that builds docs from the articles folder. You can see your article at [https://cosmosos.github.io](https://cosmosos.github.io)!

EliaSulimanov:
Great to know :)
I've updated it since, to be as in the wiki page, will create PR for it

PR #1381, opened 2020-07-03
VBE Double buffering  + MemoryBlock enhancement

valentinbreiz:


source/Cosmos.HAL2/Drivers/Video/VBEDriver.cs
```diff
@@ -66,6 +68,10 @@ private enum EnableValues
             NoClearMemory = 0x80,
         };
 
+        int xRes;
```
quajak, line 71: Why do we need these variables?

valentinbreiz: they were used in the clear /fill methods to get the size of the buffer but since the MemoryBlock as a Size member they are useless, I remove them

PR #1382, opened 2020-07-04
Re-added the Graphics documentation XML comments

Arawn-Davies:
Recent CGS update removed some XML comments, re-added those

### DIFF
--- a/source/Cosmos.System2/Graphics/VGACanvas.cs
+++ b/source/Cosmos.System2/Graphics/VGACanvas.cs
@@ -7,11 +7,25 @@ using static Cosmos.HAL.VGADriver;
 
 namespace Cosmos.System.Graphics
 {
+    /// <summary>
+    /// VGACanvas class. Used to control screen.
+    /// </summary>
     public class VGACanvas : Canvas
     {
+        /// <summary>
+        /// Private boolean whether VGA graphics mode is enabled or not
+        /// </summary>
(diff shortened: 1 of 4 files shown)


Arawn-Davies:
These are the files which AFAIK only had the missing comments, can't see any other files that missed them from older commits

source/Cosmos.System2/Graphics/VGACanvas.cs
```diff
@@ -30,138 +47,312 @@ public VGACanvas() : base()
...
         }
 
+        /// <summary>
+        /// Disables VGA graphics mode, parent method returns to 80x25 text mode
+        /// </summary>
         public override void Disable()
         {
             if (Enabled)
             {
-                VGAScreen.SetTextMode(TextSize.Size80x25);
```
valentinbreiz, line 45: why remove this line?

Arawn-Davies: Because FullScreen canvas returns back to 80x25 text mode anyway, no point having the method called twice :)

Arawn-Davies: FullscreenCanvas.Disable()

Arawn-Davies: cheers :)

PR #1388, opened 2020-07-06
[WIP] Create BufferedCanvas.cs

valentinbreiz:
- [ ] use ManagedMemoryBlock instead of Color array


quajak:
Is this still required now that we have double buffering for the drivers directly?

PR #1390, opened 2020-07-08
VFS complete code example

EliaSulimanov:
Complete code example of Cosmos VFS.


Arawn-Davies:
Are there any differences between this and the articles in Docs? If it's just for showing a complete example in a .cs file, I would update the Docs/Kernel/VFS article to point to it, as not many people look in the User folders

EliaSulimanov:
No, there is no difference. It is the same code, but complete working example of it.

quajak:
I would like to clean up the repository and therefore don't want to add more code to the users directory. Also if there is no difference between this and the article in docs, then I dont really see the point.

PR #1395, opened 2020-07-09
Basic DrawString and DrawChar with PC Screen Fonts (PSF)

valentinbreiz:
image

### DIFF
--- a/source/Cosmos.System2/Graphics/Canvas.cs
+++ b/source/Cosmos.System2/Graphics/Canvas.cs
@@ -498,7 +498,7 @@ namespace Cosmos.System.Graphics
 
         public void DrawChar(char c, Font aFont, Pen pen, int x, int y)
         {
-            int p = aFont.Height * c;
+            int p = aFont.Height * (byte)c;
 
             for (int cy = 0; cy < aFont.Height; cy++)
             {
--- a/source/Cosmos.System2/Graphics/Canvas.cs
+++ b/source/Cosmos.System2/Graphics/Canvas.cs
@@ -868,11 +868,26 @@ namespace Cosmos.System.Graphics
             DrawImage(image, point.X, point.Y);
         }
 
+        /// <summary>
+        /// Draw string.
+        /// </summary>
+        /// <param name="str">string to draw.</param>
+        /// <param name="aFont">Font used.</param>
+        /// <param name="pen">Color.</param>
+        /// <param name="point">Point of the top left corner of the string.</param>
         public void DrawString(string str, Font aFont, Pen pen, Point point)
         {
             DrawString(str, aFont, pen, point.X, point.Y);
         }
 
+        /// <summary>
+        /// Draw string.
+        /// </summary>
+        /// <param name="str">string to draw.</param>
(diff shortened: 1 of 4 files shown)


source/Cosmos.System2/Graphics/Fonts/Font.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Cosmos.System.Graphics.Fonts
+{
+    /// <summary>
+    /// Base class for fonts.
+    /// </summary>
+    public abstract class Font
```
quajak, line 10: I thinks its better if a font has a method to get the data for a single letter, because fonts might not have all letters or store all letters in a single array.

Also design and speed wise, I think it would be better if we dont overwrite the parameters Data, Height and Width and rather have a ctor where theses values are set.

source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -867,15 +868,70 @@ public void DrawImage(Image image, Point point)
...
+        /// <param name="y">Y coordinate.</param>
+        public void DrawChar(char c, Font aFont, Pen pen, int x, int y)
+        {
+            int p = aFont.Height * (byte)c;
+
+            for (int cy = 0; cy < aFont.Height; cy++)
+            {
+                for (byte cx = 0; cx < aFont.Width; cx++)
+                {
+                    if (aFont.ConvertByteToBitAddres(aFont.Data[p + cy], cx + 1))
```
quajak, line 928: Does this only work for fonts with width 8? If yes check before drawing. 

Also maybe cache `àFont.Data[p + cy]` as a small opimization.

valentinbreiz:
How can I test this?

source/Cosmos.System2/Graphics/Canvas.cs
valentinbreiz: No it should works with all font sizes

quajak:
Add some more code to https://github.com/CosmosOS/Cosmos/blob/master/Tests/Kernels/GraphicTest/Kernel.cs.

source/Cosmos.System2/Graphics/Canvas.cs
quajak: I meant move it outside of the inner for loop.

valentinbreiz:
Yes lol but I don't really know how to test that

source/Cosmos.System2/Graphics/Fonts/Font.cs
valentinbreiz: For now PSF only use an array for the whole character set. If we support more fonts it will be changed after?

quajak:
Maybe just write some text to the screen, mostly to ensure that it does not crash.

source/Cosmos.System2/Graphics/Fonts/Font.cs
quajak: Yeah that works

PR #1419, opened 2020-07-30
Getting started article

EliaSulimanov:


quajak:
Thanks for the article, IMO official cosmos tutorials should be fully hosted by cosmos. For videos, there is a Cosmos youtube channel, where we could post the videos.

Issue #1441, opened 2020-08-18
Cosmos.Core.CPU.GetCPUVendorName() returns weird string


LiquidRekto:
#### Area of Cosmos - What area of Cosmos are we dealing with?
Hardware Informaation

#### Expected Behaviour - What do you think that should happen?
Printing the hardware name. In this case: CPU

#### Actual Behaviour - What unexpectedly happens?
Weird String: "???u???l???l"
![image](https://user-images.githubusercontent.com/46742676/90488814-a76be000-e166-11ea-9477-0fcf7ca7cf29.png)

#### Reproduction - How did you get this error to appear?
Printing result of `Cosmos.Core.CPU.GetCPUVendorName();` to console
`Console.WriteLine(Cosmos.Core.CPU.GetCPUVendorName());`

#### Version - Were you using the User Kit or Dev Kit? And what User Kit version or Dev Kit commit (Cosmos, IL2CPU, X#)?
User Kit, version 20200708

quajak:
Could you get the individual chars in the string and cast them to bytes and post what you get?

LiquidRekto:
> Could you get the individual chars in the string and cast them to bytes and post what you get?

I have done what you have said and here's what I got:
`71 101 110 117 105 110 101 73 110 116 101 108`

fdagrewwwfdgsdfs:
If you convert individual decimals to ASCII you get "GenuineIntel"

LiquidRekto:
> If you convert individual decimals to ASCII you get "GenuineIntel"

Indeed, with base 10. I think I should make a separate function for it.

quajak:
Ok, so it seems somehow converting it to a string is the problem, what happens when you convert the individual chars into ints in Cosmos? My current hypothesis is that somehow the cast to char is not working correctly.

fdagrewwwfdgsdfs:
I looked in Cosmos's source code and found this:
`internal static void ReadCPUID(uint type, ref int eax, ref int ebx, ref int ecx, ref int edx) => throw new NotImplementedException();`

and

```
            if (CanReadCPUID() != 0)
            {
                int eax = 0;
                int ebx = 0;
                int ecx = 0;
                int edx = 0;
                ReadCPUID(0, ref eax, ref ebx, ref ecx, ref edx); // 0 is vendor name

                string s = "";
                s += (char)(ebx);
                s += (char)(ebx >> 8);
                s += (char)(ebx >> 16);
[...]

PR #1446, opened 2020-08-20
CPU Vendor name fix

EliaSulimanov:
Fixing #1441 issue.


quajak (approved):
Thanks for fixing this bug!

PR #1489, opened 2020-09-25
Add Brand String method to CPU.cs

samnad:


quajak (approved):
Looks good! Thanks for the pr.

PR #1522, opened 2020-10-16
Color.FromName System.Drawing plug

KM198912:
Color.FromName Plug

### DIFF
--- a/source/Cosmos.System2_Plugs/System/Drawing/ColorImpl.cs
+++ b/source/Cosmos.System2_Plugs/System/Drawing/ColorImpl.cs
@@ -1,0 +1,184 @@
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Text;
+using IL2CPU.API.Attribs;
+
+namespace Cosmos.System_Plugs.System.Drawing
+{
+    [Plug(Target = typeof(global::System.Drawing.Color))]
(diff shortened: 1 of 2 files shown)


source/Cosmos.System2_Plugs/System/Drawing/ColorImpl.cs
```diff
@@ -0,0 +1,184 @@
...
+                    return Color.GreenYellow;
+                    break;
+                case "Honeydew":
+                    return Color.Honeydew;
+                    break;
+                case "HotPink":
+                    return Color.HotPink;
+                    break;
+                case "":
+                    throw new Exception("Color Name must be passed to FromName");
```
ascpixel, line 176: You should change the type of the thrown exception to ArgumentException, so it would be easier to catch that specific exception.

quajak (changes requested):
The PR looks good, two minor nitpicks. Can you please remove the `break;` statements which are not needed and add a small test to the graphic kernel?

KM198912:
> 
> 
> The PR looks good, two minor nitpicks. Can you please remove the `break;` statements which are not needed and add a small test to the graphic kernel?

removed the break; but i dont know how to edit a second file in existing PR to add the test @quajak

quajak (approved):

Issue #1227, opened 2019-10-08
PS/2 Reset Failure + Possible Fix


KM198912:
#### Area of Cosmos - What area of Cosmos are we dealing with?
PS/2 Controller

#### Expected Behaviour - What do you think that should happen?
Keyboard and Mouse to behave properly

#### Actual Behaviour - What unexpectedly happens?
Using it on a laptop the multitouch errors out with 0x50
Keyboard Reset happens to late (0xFF)

#### Reproduction - How did you get this error to appear?
Burn iso to usb using Rufus and boot a laptop with it.

#### Version - Were you using the User Kit or Dev Kit? And what User Kit version or Dev Kit commit (Cosmos, IL2CPU, X#)?
Master DevKit + applied PS2 Patches

---------------------------

After talking on Gitter some about it i went to track the issue down, as my laptop does not have a serial port i had to log everything to console, doing so revealed that byte 250 (0xFA / Success) on the PS/2 Keyboard port failed, this led me to the assumption that the reset was not working correctly. (to late?) manually calling SendDeviceCommand with the reset hex in ps2controller.cs above 

                if (!SendDeviceCommand(DeviceCommand.IdentifyDevice, xSecondPort))
                {
                    return null;
                }

seems to fix the issue, aswell as adding xTestByte == 0x50 in  if (xTestByte == 0x00)
{ 
return true;
 }
making it look like 

 if (xTestByte == 0x00 || xTestByte == 0x50)
{
return true;
}

i am pretty sure this is the wrong way of doing it so i decided not to PR it but to tell my discoverys so maybe someone more capable can fix this.

bartekmuzyk:
What patches did you apply?

PR #1524, opened 2020-10-18
Fix Laptop Touchpads

KM198912:
Tested this on 3 Different Laptops from different brands, all touchpads worked without any issues


quajak (approved):

KM198912:
This does not close #1227, on some hardware we still have problems with the keyboard being recognized due to a reset issue, this only fixes the touchpad for laptops @quajak

quajak:
Ah ok my bad.

PR #1610, opened 2020-12-27
Implement Multithreading

valentinbreiz:
This pull request is not to merge in master yet, this is more an opening for discussion on the implementation of multithreading in Cosmos.

All the original code of the first commit comes from @Og-Rok's work from Aura's fork of Cosmos. (the original pull request can be found here if needed: https://github.com/aura-systems/Cosmos/pull/40)

Here is a short demonstration video: https://www.youtube.com/watch?v=rea5sAkYyec

To do:
- [x] fix PC Speaker and process scheduler conflict due to PIT
- [x] alloc from new heap system
- [x] add test kernel


source/Cosmos.Core/Processing/ProcessorScheduler.cs
```diff
@@ -0,0 +1,102 @@
...
+            context.state = ProcessContext.Thread_State.ALIVE;
+            context.arg = 0;
+            context.priority = 0;
+            context.age = 0;
+            context.parent = 0;
+            ProcessContext.m_ContextList = context;
+            ProcessContext.m_CurrentContext = context;
+
+            IOPort counter0 = new IOPort(0x40);
+            IOPort cmd = new IOPort(0x43);
```
valentinbreiz, line 27: Find a way to use another port or something else than PIT, with this using PC Speaker won't be possible and will hang the processor scheduler

TFTWPhoenix:
This would be very beneficial.

Nik300:
Guys have a huge problem with while loops...
this is the code I am threading... and it just stops after some seconds
```
public void TaskRun()
{
            while (true)
                Sys.Emulation.Threading.TaskManager.Next();
}```
```
public static void Next()
{
    foreach (Thread thread in LoadedThreads)
    {
       DateTime startTime = DateTime.Now;
       DateTime lastTime = startTime;
       while (lastTime.Second < startTime.Second + 1)
       {
            thread.exec.NextInstruction();
            lastTime = DateTime.Now;
        }
     }
}
```

**Review by Nik300 — CHANGES_REQUESTED** [review]
Really impressive changes here... but have some serious troubles with while loops!
I'll try to track down the error then publish my results

**Comment by Nik300** [comment]
UPDATE:
Ok so I just found out that the problem isn't the while loop but the keyboard instead...
Here's the thing:
When I load a task as long as I don't press any key it works... but as soon as I click a key the entire system shuts down 

🤔

**Comment by charlesbetros** [comment]
Is it any interrupt or just keyboard?

**Comment by Og-Rok** [comment]
From what i remember with this, a lot of the core kernel systems (interrupts, heap, device management) are not thread safe, meaning that you will get a lot of situations like this where threads are tripping each other up. A fix for things like the heap would be to implement something like spinlocks

**Comment by Nik300** [comment]
> Is it any interrupt or just keyboard?
> […](#)
> On Jan 4, 2021, at 12:21 PM, F4lc0131 ***@***.***> wrote: ﻿ UPDATE: Ok so I just found out that the problem isn't the while loop but the keyboard instead... Here's the thing: When I load a task as long as I don't press any key it works... but as soon as I click a key the entire system shuts down 🤔 — You are receiving this because you are subscribed to this thread. Reply to this email directly, view it on GitHub, or unsubscribe.

as of now just the keyboard... I'll try to bust and fix those errors before releasing this to real cosmos

**Comment by Nik300** [comment]
Actually guys I had an idea on how we could handle this but I'll have to work on a "lock" system, so that a thread can acquire the lock over something (in this case the Console.ReadLine) tell me what you guys think (especially @valentinbreiz)

**Comment by Nik300** [comment]
Screenshot 2021-01-04 225222
For now this should do the job... but i'd like to do something more flexible

**Comment by Nik300** [comment]
UPDATE:
Lol I just realized that there's something weirder going on...
ok so with the Console.ReadLine() fix I still have issues with loading more than one thread that use a while loop...
have to track that down too and fix that :(

**Comment by Nik300** [comment]
UPDATE:

Fixed! I think that Cosmos.Clear() needs the same treatment... because as soon as I use the command "cls" the system shuts down

**Comment by TFTWPhoenix** [comment]
With the trend we're seeing with methods like `Console.ReadLine()` `Console.WriteLine()` and `Console.Clear()` I'm guessing that there is other methods with this issue.

**Comment by valentinbreiz** [comment]
You're right @TFTWPhoenix, some parts of Cosmos are not thread safe and maybe needs some locks too
But WriteLine or Clear is working fine

**Comment by ThomasBeHappy** [comment]
Will this ever be finished? Kinda really waiting for this.

**Comment by Kiirx** [comment]
Any news?

**Comment by AsertCreator** [comment]
are SMP and hyper-threading real on Cosmos? this multithreading uses only one core/logic processor

**Comment by valentinbreiz** [comment]
Not yet but it's on our plan once multithreading using APIC will be done

**Comment by zarlo** [comment]
then we just need multi core

File: source/Cosmos.Core/Processing/ProcessorScheduler.cs (hunk 1 above)
Comment (AsertCreator) [reply source/Cosmos.Core/Processing/ProcessorScheduler.cs:27]: maybe pit?

Comment (AsertCreator) [reply source/Cosmos.Core/Processing/ProcessorScheduler.cs:27]: maybe use PIT implemented in Cosmos, not IOPort PIT?

Comment (zarlo) [reply source/Cosmos.Core/Processing/ProcessorScheduler.cs:27]: @AsertCreator they are one in the same

Comment (AsertCreator) [reply source/Cosmos.Core/Processing/ProcessorScheduler.cs:27]: oh okay. im trying to implement my own multitasking, using PIT override

**Comment by Kiirx** [comment]
Heya ! Yall got news ? I see that the todo is complete

**Comment by AsertCreator** [comment]
i think no. todo is complete, but all test kernels are failing, indicating that the processor scheduler doesnt work. (why does x86 hardware switch exists?)

**Comment by zarlo** [comment]
we are waiting for APIC

File: source/Cosmos.Core/Processing/ProcessContext.cs [hunk 2]
```diff
@@ -0,0 +1,154 @@
...
+            public System.Threading.ParameterizedThreadStart paramentry;
+            public Thread_State state;
+            public object param;
+            public int arg;
+            public uint priority;
+            public uint age;
+            public uint parent;
+        }
+
+        public const uint STACK_SIZE = 4096;
```
AsertCreator, line 41: idk, but 4096 bytes seems too small. maybe 16KiB?

Kiirx:
> are SMP and hyper-threading real on Cosmos? this multithreading uses only one core/logic processor

as I did a PR to move to limine its going to be easier for I saw that there was something with SMP

AsertCreator:
limine smp api is available only if you're using limine boot protocol. since its almost impossible to port cosmos to limine boot protocol, we need to use apic

Kiirx:
> limine smp api is available only if you're using limine boot protocol. since its almost impossible to port cosmos to limine boot protocol, we need to use apic

Its not impossible at all, IL2CPU and Cosmos.Core just needs rework

source/Cosmos.Core/Processing/ProcessContext.cs
AsertCreator: kernel's stack is 320kib and thread's is 4kib. we need to extend stack

AsertCreator:
i understood why tests are failing. only threading test is failing, it contains comment that `BeforeRun`, `Run` and `AfterRun` are not thread, they are just code. that means scheduler doesnt return back to then and test controller reports a timeout, because kernel didn't report success or failure

AsertCreator:
i got this to work (maybe). currently there are two problems: stack corruption detection and `Thread.Sleep` method. first one thinks that stack was corrupted, although it just got replaced. second one is not implemented and makes thread sleep indefinitely, in other words, it makes thread die. when these are not used, multithreading somehow and somewhat works. do i need to submit PR?

zarlo:
i would make a PR as the owner of the RP does not have time to work on this

AsertCreator:
ok

Neil-Ciciglycerin:
I **ABSOLUTELY** approve. There are so many things this would be useful for, and I can't find any bugs myself.
**PLEASE**, _**PLEASE**_ note: I'm a junior developer with _**VERY**_ little experience at the low level.

Neil-Ciciglycerin (approved):
__[REDACTED]__

source/Cosmos.Core/Processing/Mutex.cs
```diff
@@ -0,0 +1,27 @@
...
+    public unsafe class Mutex
+    {
+        public int gate;
+
+        [PlugMethod(PlugRequired = true)]
+        public static void MutexLock(int* mtx) { }
+
+        public void Lock()
+        {
+            while (gate != 0) { }
```
zarlo, line 14: we should yield here so we dont waste as much cpu time

Neil-Ciciglycerin:
i found a bug with how multithreading interacts with the AC97 driver [here](#2901)

source/Cosmos.Core/Processing/ProcessorScheduler.cs
Neil-Ciciglycerin: maybe you can implement the nanosecond time in RTC, and use RTC to approximate? or maybe (pure speculation, as a junior dev who knows nothing about low level) reimplement PIT wait to tally up the number of times IRQ 0 has been triggered since the os was powered on, and use this to wait _x_ amount of time. in the case of the second option, it would be possible to implement a system where multiple things can use PIT wait at the same time. see [osdev wiki](https://wiki.osdev.org/PIT#PIT_Channel_0_Example_Code).

valentinbreiz: RTC is too slow for scheduling and idk if we could make interrupts with that

valentinbreiz: In fact RTC may not be a bad idea until we have HPET or local APIC timers, it indeed can generate IRQs. I'll try, I hope this won't mess up date time

Neil-Ciciglycerin: Does that function as a stopgap until APIC gets finished?

Neil-Ciciglycerin:
i found a bug with graphics. i have some system init code in a function called __sysstartup__. there's a null reference exception when init-ing canvas, on main thread or other thread.

valentinbreiz:
> i found a bug with graphics. i have some system init code in a function called `__sysstartup__`. there's a null reference exception when init-ing canvas, on main thread or other thread.

Hello @Neil-Ciciglycerin I think this issue is also in the main branch unfortunally. Could you try and tell me please?

zarlo (approved):

source/Cosmos.Core/MMIO.cs
```diff
@@ -0,0 +1,149 @@
+﻿using IL2CPU.API.Attribs;
+
+namespace Cosmos.Core
+{
+    /// <summary>
+    /// MMIOBase abstract class.
+    /// </summary>
+    public unsafe abstract class MMIOBase
```
quajak, line 8: How does this differ from a MemoryBlock?

source/Cosmos.Core/ACPI.cs
```diff
@@ -245,240 +546,350 @@ public static void Reboot()
...
+
+                        Global.debugger.Send("Found APIC Interrupt Override (Bus: " + ((ulong)ovr->Bus).ToString() + ", Source:" + ((ulong)ovr->Source).ToString() + ", Interrupt:0x" + ((ulong)ovr->Interrupt).ToString("X") + ", Flags:" + ((ulong)ovr->Flags).ToString() + ")");
+                    }
+
+                    p += length;
                 }
             }
+        }
+
+        /*
```
quajak, line 760: Do we need this commented out code?

source/Cosmos.Core/Memory/Heap.cs
```diff
@@ -20,6 +21,8 @@ public enum ObjectGCStatus : ushort
     public static unsafe class Heap
     {
         private static uint* StackStart;
+        private static Mutex mMemeoryGate = new Mutex();
```
quajak, line 24: Typo

source/Cosmos.Core/Memory/Heap.cs
```diff
@@ -44,6 +47,11 @@ public static unsafe void Init()
 		/// <returns>New pointer with specified size while maintaining old data.</returns>
         public static byte* Realloc(byte* aPtr, uint newSize)
 		{
+            if (mMemeoryGate != null)
```
quajak, line 50: How can the mMemoryGate ever be null? Seems like we are doing useless checks

source/Cosmos.Core/ObjUtilities.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public static unsafe class ObjUtilities
+    {
+        public static uint GetPointer(Delegate aVal)
+        {
+            return (uint)aVal.GetHashCode();
+        }
+
+        [PlugMethod(PlugRequired = true)]
+        public static uint GetPointer(Object aVal) { return 0; }
```
quajak, line 14: We already have https://github.com/CosmosOS/Cosmos/blob/f7fd5696a37c146bc225573c014bf8caf7917fbf/source/Cosmos.Core/GCImplementation.cs#L108 which does more or less the same?
If you need a uint we also have https://github.com/CosmosOS/Cosmos/blob/f7fd5696a37c146bc225573c014bf8caf7917fbf/source/Cosmos.Core/GCImplementation.cs#L115

source/Cosmos.Core/Processing/Mutex.cs
```diff
@@ -0,0 +1,27 @@
...
+            gate = 1;
+            /*fixed (int* p = &gate)
```
quajak, line 16: Can the commented code be removed?

source/Cosmos.Core_Asm/CPU/CPUUpdateIDTAsm.cs
```diff
@@ -79,36 +79,71 @@ public override void AssembleNew(Assembler aAssembler, object aMethodInfo)
...
-                XS.SSE.FXRestore(ESP, isIndirect: true);
-
-                XS.Set(ESP, EAX); // this restores the stack for the FX stuff, except the pointer to the FX data
-                XS.Add(ESP, 4); // "pop" the pointer
-
-                XS.PopAllRegisters();
-
+                else
+                {
+                    new LiteralAssemblerCode("pushad");
```
quajak, line 117: Why not use the XS equivalent commands?

source/Cosmos.Core_Asm/ObjUtilitiesImpl.cs
```diff
@@ -0,0 +1,40 @@
...
+using System;
+using XSharp;
+using XSharp.Assembler;
+
+namespace Cosmos.Core_Asm
+{
+    [Plug(Target = typeof(ObjUtilities))]
+    public static unsafe class ObjUtilitiesImpl
+    {
+        [PlugMethod(Assembler = typeof(ObjUtilitiesGetPointer))]
```
quajak, line 13: See previous comment if we really need this method new here.

source/Cosmos.Core_Plugs/System/Threading/ThreadImpl.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        public static void MemoryBarrier()
+        {
+
+        }
+
+        public static void Ctor(ThreadStart aThis, ThreadStart aEntry)
+        {
+            Console.WriteLine("Thread started");
```
quajak, line 22: I think this should be removed.

source/Cosmos.Core_Plugs/System/Threading/ThreadImpl.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        //    public static void SleepInternal(int ms)
```
quajak, line 25: Can this code be removed?

source/Cosmos.HAL2/Global.cs
```diff
@@ -46,14 +47,35 @@ static public void Init(TextScreenBase textScreen, bool InitScrollWheel, bool In
...
             Console.Clear();
-            Console.WriteLine("Finding PCI Devices");
-            debugger.Send("PCI Devices");
-            PCI.Setup();
 
             Console.WriteLine("Starting ACPI");
             debugger.Send("ACPI Init");
             ACPI.Start();
 
+            Console.WriteLine("Finding PCI Devices");
```
quajak, line 55: I don't think we should write as much to the console, since the user cant control this behavior.

quajak (changes requested):

Neil-Ciciglycerin:
The last time I checked it wasn't in the main branch, although I was using userkit, I'll check now

source/Cosmos.Core/Memory/Heap.cs
zarlo: we could replace it with ```mMemeoryGate?.Lock();``` it would do that same

Neil-Ciciglycerin:
Yeah, it's in the main branch. I'll go start an issue.

source/Cosmos.Core/Memory/Heap.cs
quajak: I would prefer if we dont do needless checks to reduce the performance cost, as these calls happen quite often

Issue #1603, opened 2020-12-22
Add ability to choose which instance of Visual Studio at build time


valentinbreiz:
Having multiple instances of Visual Studio installed can cause build issues saying Visual Studio Workloads are not installed and / or failed to install. Detecting if multiple versions of Visual Studio are installed and add the ability to choose which one at build time would be good.

I had the issue with these for example:
![image](https://user-images.githubusercontent.com/18724279/102919091-8276d900-4488-11eb-82e9-9189a500b980.png)

charlesbetros:
We should remove the auto-install of workloads and just provide a message telling the user which to install

MishaProductions:
I think this was fixed

PR #1619, opened 2021-01-03
Disable auto-install of workloads

MishaTY:
Disable auto-install of workloads

Should fix issue #1603


quajak (changes requested):
Thanks for the PR. This is a needed change. It would be awesome, if we could tell the user which workload is missing, rather than just saying that workloads are missing.

MishaTY:
Changed it so it will tell the user on what exactly is missing.

quajak (approved):

PR #1622, opened 2021-01-04
Fix Console methods being split across threads

Nik300:
Screenshot 2021-01-05 000253

This should completely fix the bug we had with Console.ReadLine and threads ;)


source/Cosmos.System2/LockSystem.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Cosmos.System
+{
+    public class LockSystem
```
valentinbreiz, line 7: We already use a mutex class as a lock system, you should use it rather than creating another.
You can find the class here: https://github.com/CosmosOS/Cosmos/blob/impl-multitask/source/Cosmos.Core/Processing/Mutex.cs
And an example of use here: https://github.com/CosmosOS/Cosmos/blob/impl-multitask/source/Cosmos.Core/Memory/Old/Old_Heap.cs#L65

valentinbreiz (changes requested):

source/Cosmos.System2/LockSystem.cs
Nik300: thank you! ;)
working on that, will be ready pretty soon!

Nik300: Done! See [this commit](https://github.com/Nik300/Cosmos/commit/5184a701b839113c8173555d36fe65ca67cd2b90) to check if it's everything ok! ;)

source/Cosmos.System2_Plugs/System/ConsoleImpl.cs
```diff
@@ -429,17 +431,20 @@ public static ConsoleKeyInfo ReadKey(bool intercept)
             bool xControl = (key.Modifiers & ConsoleModifiers.Control) == ConsoleModifiers.Control;
 
             return new ConsoleKeyInfo(key.KeyChar, key.Key.ToConsoleKey(), xShift, xAlt, xControl);
-        }
-
+        }
+
+        private static List<char> chars = new List<char>(32);
```
valentinbreiz, line 436: Why move chars outside the ReadLine function?

source/Cosmos.System2_Plugs/System/ConsoleImpl.cs
```diff
@@ -601,17 +608,21 @@ public static void Write(bool aBool)
         /* Correct behaviour printing null should not throw NRE or do nothing but should print an empty string */
         public static void Write(object value) => Write((value ?? String.Empty));
 
+        private static Cosmos.System.Console xConsole = null;
```
valentinbreiz, line 611: Same here

source/Cosmos.System2_Plugs/System/ConsoleImpl.cs
```diff
@@ -13,7 +13,9 @@ public static class ConsoleImpl
         private static ConsoleColor mForeground = ConsoleColor.White;
         private static ConsoleColor mBackground = ConsoleColor.Black;
         private static Encoding ConsoleInputEncoding = Encoding.ASCII;
-        private static Encoding ConsoleOutputEncoding = Encoding.ASCII;
+        private static Encoding ConsoleOutputEncoding = Encoding.ASCII;
+        private static Core.Processing.Mutex mConsoleGateRead = new Core.Processing.Mutex();
+        private static Core.Processing.Mutex mConsoleGateWrite = new Core.Processing.Mutex();
```
valentinbreiz, line 18: Please specify Cosmos.Core.Processing with an using rather than this

valentinbreiz (changes requested):

source/Cosmos.System2_Plugs/System/ConsoleImpl.cs
Nik300: because if we put them into the function, every thread would have to access its own instance of characters and this would just waste memory and cause errors when threads that don't require "keyboard read" are forced to add to that list(that for them doesn't exist) a character.

source/Cosmos.System2_Plugs/System/ConsoleImpl.cs
Nik300: Right. will fix in a moment

valentinbreiz:
Closing this for same reasons as #1921

PR #1634, opened 2021-01-13
Update RestoreSources.props

KM198912:
myget currently has a issue where it cant find roslyn-tools, so disable it for the time being so the devkit builds


quajak (approved):
I cant figure out what we need from the roslyn-tools nuget source. As the build passes, i think we dont need to replace the source with a new one.

Issue #1606, opened 2020-12-23
Missing Plug for Interlocked.Exchange


ArtGo-Devloppper:
#### Area of Cosmos - What area of Cosmos are we dealing with?
IL2CPU
#### Expected Behaviour - What do you think that should happen?
I use Refference.Object (Asign a bitmap var with a ressource file)

#### Reproduction - How did you get this error to appear?
`ProgsImages[6] = new Bitmap(Ressources.Backgrounds.timg);`

#### Version - Were you using the User Kit or Dev Kit? And what User Kit version or Dev Kit commit (Cosmos, IL2CPU, X#)?
I use User Kit 20200708

Thanks for you awanser

quajak:
Whats the error message you are getting?

ArtGo-Devloppper:
Just `IL2CPU Error : System.Exeption :`

quajak:
Whats after the colon, can you please post the entire message

charlesbetros:
A missing plug I am guessing? Cosmos doesn't support embedded resources.

ArtGo-Devloppper:
> Whats after the colon, can you please post the entire message

No, I can’t because the message is just `IL2CPU Error : System.Exeption : `

quajak:
Please post the entire log from VS output. I doubt that no further information about the exception is insluded in the log.

ArtGo-Devloppper:
The vs output si thé cosmos debugger ?

quajak:
VS has an output window which should either in the debug/build log have more information about the build. The Cosmos debugger does not play a role here.

ArtGo-Devloppper:
This ?
`L'opération de génération a démarré...
1>------ Début de la génération : Projet : ArtOS, Configuration : Debug Any CPU ------
1>C:\Program Files\dotnet\sdk\5.0.101\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(28,5): warning NETSDK1138: La version cible de .NET Framework ('netcoreapp2.0') n'est pas prise en charge et ne recevra pas les mises à jour de sécurité. Consultez https://aka.ms/dotnet-core-support pour plus d'informations sur la stratégie de support.
1>ArtOS -> C:\Users\USERNAME\Documents\GitHub\ArtOS (Computers)\ArtOS (Computers)\bin\Debug\netcoreapp2.0\cosmos\ArtOS.dll
1>Message: Executing IL2CPU on assembly
[...]

PR #1641, opened 2021-01-16
Add Interlocked plugs

MishaTY:
- Adds most plugs for the Interlocked class.


quajak (approved):
Thanks for the PR! This should close https://github.com/CosmosOS/Cosmos/issues/1606

PR #1644, opened 2021-01-19
 DoubleBufferedVMWareSVGAII With DoubleBufferedSVGAIICanvas and DrawASCIIString

nifanfa:
Sorry for late


valentinbreiz (approved):
Tested and working like a charm!

nifanfa:
You can make a wonderful operating system by this in VMware workstation
A1789E2EEFC2E939B8425DBBEC40E0BB
My best work.

nifanfa:
i don't know how to use pull request before So i put it off for a long time..

nifanfa:
use the methods which's name by doublebufferedxxxxxx

source/Cosmos.System2/Graphics/Fonts/ASC16.cs
```diff
@@ -0,0 +1,39 @@
...
+        static MemoryStream ASC16FontMS = new MemoryStream(Convert.FromBase64String(ASC16Base64));
+
+        public static void _DrawACSIIString(this DoubleBufferedVMWareSVGAII vMWareSVGAII, string s, uint color, uint x, uint y)
```
valentinbreiz, line 13: You shouldn't use a new function to draw a string, we have one in Canvas.cs
What do you think @quajak?

quajak (changes requested):
Thanks for the PR, its great to have a double buffered VMWare SVGAII driver.
It would be awesome if you could do the following changes:
- Share the Register, ID, FIFO etc. classes between the two VMWare SVGAII drivers
- Make DoubleBufferedVMWareSVGAII only include the SetPixel method and define this as a child class of the other VMWareSVGAII driver. Then update SVGAIICanvas to work with both implementation (for example by passing a bool if the driver should be double buffered and changing the implementation of the Display method). This would decrease the amount of repeated code immensily.
If you have questions regarding these changes, please get in touch on gitter.

source/Cosmos.System2/Graphics/Fonts/ASC16.cs
quajak: Yeah, unless this has some advantage vs the one we have in canvas, I dont see the need.

nifanfa:
i will

nifanfa:
[](https://github.com/nifanfa/Cosmos-DrawString)

source/Cosmos.System2/Graphics/Fonts/ASC16.cs
nifanfa: changed

nifanfa:
now DoubleBufferedVMWareSVGAII supported canvas

nifanfa:
After drawing then use Canvas.Display() to draw it on the screen

nifanfa:
GIF 2021-1-21 17-08-08  
No Lag

nifanfa:
QQ截图20210121171008

source/Cosmos.System2/Graphics/Fonts/ASC16.cs
```diff
@@ -0,0 +1,43 @@
...
+        static MemoryStream ASC16FontMS = new MemoryStream(Convert.FromBase64String(ASC16Base64));
+
+        //Note: the size of font is 8*16
+        
+        public static void DrawACSIIString(this Canvas canvas, Pen pen, string s, int x, int y)
```
quajak, line 16: The two differences between this method and the DrawString in https://github.com/CosmosOS/Cosmos/blob/1e4cbb65e83a800666aba4ca8e244e7585e76594/source/Cosmos.System2/Graphics/Canvas.cs#L964, I can see is that this one supports multiline strings and that the other supports more than one font. So you could you please add multiline support to the other implementation? Or is there some advantage to this one, which I am not seeing.

quajak (changes requested):

source/Cosmos.System2/Graphics/Fonts/ASC16.cs
nifanfa: @quajak we can draw converted font from windows

nifanfa: i think pcsfont are too old and nobody will make new font..

quajak: Ah, what font format is it? How do you convert it? Just want to allow the user to also be able to create and load their own fonts.

nifanfa: DZK

nifanfa: actually this file has no a exactly format. but there's a lot tool to generate this files.

nifanfa: QQ截图20210121193216
  like this

nifanfa: this format of fonts always used for monito or MCU

quajak: In that case, I would prefer if you add support for this font format by adding a class inheriting from Font, similar to PCScreenFont, so that users can also load custom fonts from this format. This would allow both types of fonts to be used by the method in canvas. If you also add multiline support to the method in Canvas, then we have everything offered from the new method, without having two seperate methods to confuse users.

nifanfa: QQ截图20210121204220
QQ截图20210121204354

source/Cosmos.System2/Graphics/Fonts/ASC16.cs
```diff
@@ -0,0 +1,43 @@
+﻿using Cosmos.System.Graphics;
+using System;
+using System.Drawing;
+using System.IO;
+using System.Text;
+
+namespace Cosmos.System.Graphics.Fonts
+{
+    public static class ASC16
```
valentinbreiz, line 9: This class should be removed for now. We need to enhance the font system to support multiple font formats

source/Cosmos.System2/Graphics/DoubleBufferedSVGAIICanvas.cs
```diff
@@ -0,0 +1,336 @@
...
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+
+namespace Cosmos.System.Graphics
+{
+    /// <summary>
+    /// SVGAIIScreen class. Used to draw ractengales to the screen. See also: <seealso cref="Canvas"/>.
+    /// </summary>
+    public class DoubleBufferedSVGAIICanvas : Canvas
```
valentinbreiz, line 13: The idea of a double buffered canva is great. But shouldn't we implement one DoubleBufferedCanvas class that can use the double buffered VBE driver and also the VMWare SVGAII one? @quajak?

valentinbreiz (changes requested):

source/Cosmos.System2/Graphics/DoubleBufferedSVGAIICanvas.cs
quajak: I think the better design would be that every canvas has a buffered and unbuffered version implemented in the same class and toggled by some bool(at least accesible in the constructor). We can then also have a FullScreenCanvas.GetDoubleBufferedCanvas which returns the double buffered version. I think this is a better design since double buffering implementation depends on the driver but there should be some shared code between the buffered and non buffered implementations.

valentinbreiz (approved):
For now this can be merged but some work on CGS needs to be done to implement non double buffered VBE driver.

source/Cosmos.HAL2/Drivers/PCI/Video/DoubleBufferedVMWareSVGAII.cs
```diff
@@ -0,0 +1,703 @@
...
+    /// VMWareSVGAII class.
+    /// </summary>
+    public class DoubleBufferedVMWareSVGAII
+    {
+        #region Parameter
+
+        /// <summary>
+        /// Register values.
+        /// </summary>
+        public enum Register : ushort
```
quajak, line 16: The following types/definitions have no reason to be duplicated by the other vmare svgaii driver.

quajak (approved):

Tests/Kernels/GraphicTest/Kernel.cs
```diff
@@ -166,13 +166,13 @@ protected override void Run()
                 TestBitmaps();
 
                 /* First test with the DefaultMode */
-                Canvas canvas = FullScreenCanvas.GetFullScreenCanvas();
```
quajak, line 169: Can we add some test with double buffered canvases?

source/Cosmos.HAL2/Drivers/PCI/Video/VMWareSVGAII/SVGAII.cs
```diff
@@ -154,7 +154,7 @@ public enum Register : ushort
         /// <summary>
         /// ID values.
         /// </summary>
-        private enum ID : uint
+        public enum ID : uint
```
quajak, line 157: I think they all should be `protected ` and not `public`.

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -92,21 +106,28 @@ public override void DrawPoint(Pen aPen, int aX, int aY)
...
+                {
+                    xColor = AlphaBlend(xColor, GetPointColor(aX, aY), xColor.A);
+                }
+
+                mSVGAIIDebugger.SendInternal($"Drawing point to x:{aX}, y:{aY} with {xColor.Name} Color");
+                _xSVGADriver.SetPixel((uint)aX, (uint)aY, (uint)xColor.ToArgb());
+                mSVGAIIDebugger.SendInternal($"Done drawing point");
+                /* No need to refresh all the screen to make the point appear on Screen! */
+                //xSVGAIIDriver.Update((uint)x, (uint)y, (uint)mode.Columns, (uint)mode.Rows);
+                _xSVGADriver.Update((uint)aX, (uint)aY, 1, 1);
```
quajak, line 129: I think it would be better if this is the only line which is not executed when DoubleBuffered, is there a reason to not have AlphaBlending when double buffering?

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -151,7 +172,20 @@ public override void DrawPoint(Pen aPen, float aX, float aY)
...
         public override void DrawFilledRectangle(Pen aPen, int aX_start, int aY_start, int aWidth, int aHeight)
         {
-            _xSVGADriver.Fill((uint)aX_start, (uint)aY_start, (uint)aWidth, (uint)aHeight, (uint)aPen.Color.ToArgb());
+            if (DoubleBuffered)
+            {
+                for (uint h = 0; h < aHeight; h++)
+                {
+                    for (uint w = 0; w < aWidth; w++)
+                    {
+                        _xSVGADriver.SetPixel((uint)(w + aX_start), (uint)(aY_start + h), (uint)aPen.Color.ToArgb());
```
quajak, line 181: Fill should also work with the double buffered implementation, so we dont have to handle it any differently and can just call `Fill` as well.

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -313,7 +347,14 @@ private void SetGraphicsMode(Mode aMode)
         /// <exception cref="NotImplementedException">Thrown if VMWare SVGA 2 has no rectange copy capability</exception>
         public override void Clear(Color aColor)
         {
-            _xSVGADriver.Fill(0, 0, (uint)Mode.Columns, (uint)Mode.Rows, (uint)aColor.ToArgb());
+            if (DoubleBuffered)
```
quajak, line 350: Again in both cases we should be able to just call Clear

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -360,7 +415,14 @@ public void CreateCursor()
         /// <exception cref="NotImplementedException">Thrown if VMWare SVGA 2 has no rectangle copy capability</exception>
         public void CopyPixel(int aX, int aY, int aNewX, int aNewY, int aWidth = 1, int aHeight = 1)
         {
-            _xSVGADriver.Copy((uint)aX, (uint)aY, (uint)aNewX, (uint)aNewY, (uint)aWidth, (uint)aHeight);
+            if (!DoubleBuffered)
```
quajak, line 418: I think its easier if we just implemenent copy in the double buffered driver to remove the difference here.

quajak (changes requested):

Tests/Kernels/GraphicTest/Kernel.cs
valentinbreiz: I will make a new pull request with simple buffered VBE canva and add some tests to it.

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
valentinbreiz: it can be used for double buffering too but it will cause performance issues if we do more if / else / else if ...

quajak: I just want to move the if in this method to only around the update command and have the drawing have the same logic regardless of double buffering.

elkaamee326:
> You can make a wonderful operating system by this in VMware workstation
> A1789E2EEFC2E939B8425DBBEC40E0BB
> My best work.

How did you create that animation where the bitmap icons move up and then down? It seems very interesting...

nifanfa:
https://github.com/CosmosOS/Cosmos/pull/1845

Arawn-Davies:
How's the progress coming along? This looks fantastic :D

nifanfa:
> How's the progress coming along? This looks fantastic :D

Hey.  It's actually not good.  Need to implement GC.  Using the GUI will cause the system to only run for a few minutes.
#

nifanfa:
oh shit i checked the wrong buttom

quajak:
It looks like your fork was deleted, if you recreate nifanfa:master with the changes we can maybe reopen the issue or you will have to open a new pr with the changes.

PR #1652, opened 2021-01-22
RTL8168 Implementation

### LINKED ISSUE #1650: RTL8168/9 support
To do:
- [x] port old driver to Cosmos
- [ ] fix driver and finish driver
- [ ] do tests

valentinbreiz:


source/Cosmos.Core/Ports.cs
```diff
@@ -0,0 +1,86 @@
...
+namespace Cosmos.Core
+{
+    public static class Ports
+    {
+        /// <summary>
+        /// Reads a byte
+        /// </summary>
+        /// <param name="port"></param>
+        /// <returns></returns>
+        public static byte InB(ushort port)
```
quajak, line 14: I prefer we go with the same approach we have for other drivers as well, which is that we define all the the IOPorts we need and then access the one we actually need. For example see https://github.com/CosmosOS/Cosmos/blob/a3d0a074f1f9db340694769116892aea925b8fde/source/Cosmos.Core/IOGroup/ATA.cs#L13 
That approach should be quicker since we arnt constantly creating new objects and less error prone than calculating the port address every time.

source/Cosmos.HAL2/Drivers/PCI/Network/RTL8168.cs
```diff
@@ -0,0 +1,358 @@
...
+
+            Ports.OutD((ushort)(BaseAddress + 0x20), mTxDescriptor.Offset);
+            Console.WriteLine("addresstx desc: " + mTxDescriptor.Offset);
+
+            Ports.OutD((ushort)(BaseAddress + 0xE4), mRxDescriptor.Offset);
+            Console.WriteLine("addressrx desc: " + mRxDescriptor.Offset);
+
+            if (((GetMacVersion() & 0x7cf00000) == 0x54100000) || ((GetMacVersion() & 0x7cf00000) == 0x54000000))
+            {
+                Console.WriteLine("8168H Detected!");
```
quajak, line 141: Remove this Console.WriteLine

source/Cosmos.HAL2/Drivers/PCI/Network/RTL8168.cs
```diff
@@ -0,0 +1,358 @@
...
+
+            //Console.WriteLine("Netcard version: 0x" + System.Utils.Conversion.DecToHex((int)GetMacVersion() & 0x7cf00000));
+            //Console.WriteLine("Netcard version: 0x" + System.Utils.Conversion.DecToHex((int)GetMacVersion() & 0x7c800000));
+
+        }
+
+        public bool Reset()
+        {
+            Ports.OutB((ushort)(BaseAddress + 0x37), 0x10); /* Send the Reset bit to the Command register */
+            while ((Ports.InB((ushort)(BaseAddress + 0x37)) & 0x10) != 0) { } /* Wait for the chip to finish resetting */
```
quajak, line 158: Should we have timeout in this method, in which case we return false?

source/Cosmos.HAL2/Drivers/PCI/Network/RTL8168.cs
```diff
@@ -0,0 +1,358 @@
...
+            return true;
+        }
+
+        protected void HandleNetworkInterrupt(ref IRQContext aContext)
+        {
```
quajak, line 163: Are all the Console.WriteLines in this method required? They seem like debug output and should be removable

source/Cosmos.HAL2/Network/NetworkInit.cs
```diff
@@ -41,6 +41,26 @@ public static void Init()
...
                     #endregion
 
+                    #region RTL8168
+
+                    if (device.VendorID == 0x10EC && device.DeviceID == 0x8168)
+                    {
+
+                        Console.WriteLine("NIC IRQ: " + device.InterruptLine);
+
+                        var RTL8168Device = new RTL8168(device);
```
quajak, line 51: How is the device accessible from outside the class? Again I dont like the Console.WriteLines, maybe make the Init(bool aShowOutput) have a toggle for console output?

quajak (changes requested):

PR #1653, opened 2021-01-26
[WIP] Sound blaster 16 Driver and AudioManager class

MishaTY:
Sound blaster 16 Driver and AudioManager class


source/Cosmos.HAL2/Drivers/PCI/Audio/AudioManager.cs
```diff
@@ -0,0 +1,78 @@
...
+    public class AudioManager
+    {
+        public static List<SoundCard> devices = new List<SoundCard>();
+
+        public static void Init()
+        {
+            devices.Clear(); //just in case
+            try
+            {
+                Console.WriteLine("Searching for sound cards");
```
quajak, line 20: Cosmos itself shouldnt be writing to the console, since that is something the individual os should be able to control.

source/Cosmos.HAL2/Drivers/PCI/Audio/AudioManager.cs
```diff
@@ -0,0 +1,78 @@
...
+                    Console.WriteLine("No supported sound cards sound.");
+                }
+                else
+                {
+                    Console.WriteLine("Sound card(s) initialized successfully!");
+                }
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine("Error while initializing sound: " + ex.Message);
```
quajak, line 41: Does the sound blaster init regularly throw exceptions? If not please dont use a try catch block, since cosmos currently cant handle multilayered try catch blocks.

source/Cosmos.HAL2/Drivers/PCI/Audio/AudioManager.cs
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+        /// <summary>
+        /// Play unsigned 8-bit PCM on first Sound card
+        /// </summary>
+        /// <param name="audio"></param>
+        public static void PlayAudio(byte[] audio)
```
quajak, line 48: Could you add a second defintion for `PlayAudio` which allows the user to select which device should be used to play the sound?

source/Cosmos.HAL2/Drivers/PCI/Audio/AudioManager.cs
```diff
@@ -0,0 +1,78 @@
...
+            if (devices.Count == 0)
+            {
+                throw new InvalidOperationException("No sound cards are installed!");
+            }
+            devices[0].PlaySound(audio);
+        }
+        /// <summary>
+        /// Stops all sound on all sound cards
+        /// </summary>
+        public static void StopAllSound()
```
quajak, line 59: Could you please add a `StopSound(int device)` method as well?

source/Cosmos.HAL2/SoundBlaster16.cs
```diff
@@ -0,0 +1,334 @@
...
+        const byte DSP_SETTIME = 0x40;
+        const byte DSP_SETPLAYBACKRATE = 0x41;
+        #endregion
+
+        #region Mixer Ports
+        const ushort MixerPort = 0x224;
+        const ushort MixerDataPort = 0x225;
+        #endregion
+        private byte[] Buffer = new byte[64000]; //64K Buffer
+        private byte* bufferinmem;
```
quajak, line 50: Can you please replace the byte* with a `MemoryBlock`? This should make the HAL project again not need unsafe code. As a rule of thumb, no driver in HAL should require using unsafe code.

source/Cosmos.HAL2/SoundBlaster16.cs
```diff
@@ -0,0 +1,334 @@
...
+            {
+                //Clean up
+                PlayingSound = false;
+                IsAudioBiggerThenBuffer = false;
+                AudioStop = 0;
+                audiodata = null;
+            }
+        }
+        #region I/O
+        private void MixerPort_write(byte port, byte data)
```
quajak, line 286: Could you please rename all the IO methods so that they have a CamelCase name?

source/Cosmos.HAL2/Drivers/PCI/Audio/Generic/SoundCard.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Cosmos.HAL.Drivers.PCI.Audio.Generic
+{
+    public abstract class SoundCard
+    {
```
quajak, line 8: Maybe add methods to `SetVolume` and `GetVolume`.

quajak (changes requested):
Thanks for the PR! All of the requested changes are just about making the code follow the standards of the codebase.

elkaamee326:
I’ve tested the code out, and it worked!(It played real audio) But for some reason, when I tried to put a huge byte array, about the size of a 2.5 mb wav file, Nasm stopped the build, and it gave me an error about the macro recursion being “endless”, because the byte array that I am using is too big. Is Cosmos capable of playing large files, or is there a limit on the data size, that can be played? 
Cosmos Userkit version: The latest current userkit version.

source/Cosmos.HAL2/SoundBlaster16.cs
```diff
@@ -0,0 +1,324 @@
...
+                //No sound blaster installed
+                return;
+            }
+
+            //Get DSP version
+            WriteDSP(DSP_GETVERSION);
+            sb16_version_major = ReadDSP();
+            sb16_version_minor = ReadDSP();
+
+            Console.WriteLine("SoundBlaster Version: " + DSPVersion.ToString());
```
quajak, line 83: Please also remove this Console.WriteLine

quajak (changes requested):

quajak:
> I’ve tested the code out, and it worked!(It played real audio) But for some reason, when I tried to put a huge byte array, about the size of a 2.5 mb wav file (If you were to take the byte array and turn it into a wav file again), Nasm stopped the build, and it gave me an error about the macro recursion being “endless”, because the byte array that I am using is too big. Is Cosmos capable of playing large files (byte arrays), or is there a limit on the data size, that can be played?
> Cosmos Userkit version: The latest current userkit version.

You can load larger files by including them directly in the vmdk file.

Docs/articles/Kernel/Sound.md
```diff
@@ -0,0 +1,32 @@
+# Network
```
valentinbreiz, line 1: Network?

source/Cosmos.HAL2/Global.cs
```diff
@@ -67,6 +68,8 @@ static public void Init(TextScreenBase textScreen)
             mDebugger.Send("Network Devices Init");
             NetworkInit.Init();
 
+            AudioManager.Init();
```
valentinbreiz, line 71: Could you send some log through the debugger?
mDebugger.Send("Audio Manager Init"); or something

source/Cosmos.HAL2/SoundBlaster16.cs
```diff
@@ -0,0 +1,324 @@
...
+        private void MixerPortWrite(byte port, byte data)
+        {
+            Outb(MixerPort, port);
+            Outb(MixerDataPort, data);
+        }
+        private byte MixerPortRead()
+        {
+            return Inb(MixerDataPort);
+        }
+        private static byte Inb(ushort port)
```
valentinbreiz, line 294: Oh, I also added an inb and outb function for another pull request: https://github.com/CosmosOS/Cosmos/blob/rtl8168/source/Cosmos.Core/Ports.cs
@quajak do you think that every device needs an IOGroup class like the PCNETII driver Cosmos.Core/IOGroup/Network/AMDPCNetIIIOGroup.cs), or we can just add a class for IO Ports with functions like inb, outb and so on?

quajak (dismissed):

valentinbreiz (changes requested):
Also please add your article in Docs/articles/toc.md

source/Cosmos.HAL2/SoundBlaster16.cs
quajak: I think it should be fine if we just use the `Port` directly in the driver class. Or is there some advantage to IOGRoups I am not thinking of?

valentinbreiz (dismissed):

elkaamee326:
> > I’ve tested the code out, and it worked!(It played real audio) But for some reason, when I tried to put a huge byte array, about the size of a 2.5 mb wav file (If you were to take the byte array and turn it into a wav file again), Nasm stopped the build, and it gave me an error about the macro recursion being “endless”, because the byte array that I am using is too big. Is Cosmos capable of playing large files (byte arrays), or is there a limit on the data size, that can be played?
> > Cosmos Userkit version: The latest current userkit version.
> 
> You can load larger files by including them directly in the vmdk file.

Would I use something like this? 
```byte[] wav = File.ReadAllBytes(@"0:\Filename.wav");```
And then call this?
```AudioManager.PlayAudio(wav);```

Everytime I would try this It would shutdown.
Or is there another way I can do this without the VM shutting down?

MishaTY:
Yes, you would do something like that, but remember that it only supports uncompressed unsigned 8 bit 16KHz PCM audio.

elkaamee326:
I tried that, but it shutdown for some reason, when I tried to load the file, the file is Unsigned 8bit PCM at 8KHz.

MishaTY:
Its supposed to be 8KHz.

elkaamee326:
Okay.

MishaTY:
Its supposed to be 16Khz sampling rate.

elkaamee326:
The VM keeps on shutting down, when I try to use File.ReadAllBytes.

elkaamee326:
I even changed the rate.

MishaTY:
Ok, I will take a look at it.

elkaamee326:
It’s not the audio, it is the filesystem.

MishaTY:
Maybe try a smaller file and see weather that works?

elkaamee326:
I tried that, and by it didn’t work? What can I use other than File.ReadAllBytes() to play the contents of the wav file?

quajak:
What is the size of the files you are trying to read? File.ReadAllBytes should be working

elkaamee326:
It is about 2.5 mb.

elkaamee326:
Even the smaller file didn't work.

quajak:
Does File.ReadAllBytes never work? Do other file operations work?

elkaamee326:
Do I use the wav file? Or do I put it into a text file (the byte array contents)?

elkaamee326:
This is the code:
`
        protected override void BeforeRun()
        {
            CosmosVFS fs = new CosmosVFS();
            VFSManager.RegisterVFS(fs);
            AudioManager.Init();

            //byte[] wav = fs.GetFile(@"0:\Windows-XP-Installation-Music.wav"); // This is the big file I was trying to load

            try
            {
                byte[] wav = File.ReadAllBytes(@"0:\Drum.wav");
                AudioManager.PlayAudio(wav);
            }
            catch (Exception ex)
            {
                Console.WriteLine(ex.Message.ToString());
            }

        }
`

Is this the correct way to load a audio file with this implementaton?

quajak:
If no file system operation is working, then I think the vmdk got corrupted when you added the file.

elkaamee326:
It works amazingly, when I convert the file into a byte array and I play it.

valentinbreiz:
If the size of the file is upper than 4096 bytes ReadAllBytes will not work. We have issues reading large files and this need to be fixed.

elkaamee326:
Okay, then thank you.

quajak:
What work is required before this can be merged? @MishaTY @valentinbreiz

MishaTY:
I need to do some more testing, and make sure that the IRQ fires.

quajak:
Ok, just remove the WIP when you are done. Would be cool if you could add some test to the TestRunner which confirms that the driver doesnt crash as well.

boristyukin:
@MishaTY this is a great enhancement, hope you can finish it soon

elkaamee326:
Just a question, is the IRQ supposed to be invoked when the user calls AudioManager.Init(); or is it called on AudioManager.PlayAudio(byte[] Sound); ? Because the Program_MixerPort function in the Enable sector has an Invoke method (for the IRQ), thx :D

MishaTY:
The sound card will invoke the IRQ when the sound finishes playing.

elkaamee326:
So far, which IRQ worked the best?

MishaTY:
Right now, IRQ 10 works

elkaamee326:
When I used IRQ 10, it worked but on the VM I put a `Console.WriteLine("Got IRQ");`. The problem is that it doesn't stop saying the `Console.WriteLine();` Message that I put. Was that supposed to happen?

MishaTY:
I am closing this PR because I could not get this to work properly. Plus, This will only work in virtual machines, since no one has a sound blaster 16 today. Maybe I will try to make a Intel HDA Driver or AC97 Driver.

elkaamee326:
Okay, good luck on your next implementation :D

MishaTY:
Thanks!

elkaamee326:
Would it be possible if you can use the same implementation for this interface, except you use the registers for a different sound card driver? `Ie: same code (or modified), different registers.`?

MishaTY:
Yes, You can reuse this code, but it will only work with the Sound blaster 16. Right now, I am not going to work on audio, since the other sound cards are way harder then the sound blaster 16.

PR #1717, opened 2021-04-09
Update VMWareSVGAII.cs

KM198912:
Fix SVGA to avoid a Index out of Bounds crash that happened due to a double memory write.
With this we disable the Driver before writing new values to the FIFO and initiating the memory again, before we initiated the memory regardless if it was already set or not and sent garbage to the FIFO causing it to crash. Also introduce the ability to switch between Graphics And Text mode.

### DIFF
--- a/source/Cosmos.HAL2/Drivers/PCI/Video/VMWareSVGAII.cs
+++ b/source/Cosmos.HAL2/Drivers/PCI/Video/VMWareSVGAII.cs
@@ -517,6 +517,8 @@ namespace Cosmos.HAL.Drivers.PCI.Video
         /// <param name="depth">Depth.</param>
         public void SetMode(uint width, uint height, uint depth = 32)
         {
+            //Disable the Driver before writing new values and initiating it again to avoid a memory exception
+            Disable();
             // Depth is color depth in bytes.
             this.depth = (depth / 8);
             this.width = width;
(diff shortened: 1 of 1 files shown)


source/Cosmos.HAL2/Drivers/PCI/Video/VMWareSVGAII.cs
```diff
@@ -753,6 +755,11 @@ public void DefineCursor()
                 WriteToFifo(0xFFFFFF);
             WaitForFifo();
         }
+        //Allow to enable the Driver again after it has been disable (switch between text and graphics mode currently this is SVGA only)
+        public void Enable()
```
quajak, line 759: Please call this method in SetMode insted of line 529

quajak (changes requested):

source/Cosmos.HAL2/Drivers/PCI/Video/VMWareSVGAII.cs
```diff
@@ -753,6 +755,11 @@ public void DefineCursor()
                 WriteToFifo(0xFFFFFF);
             WaitForFifo();
         }
+        //Allow to enable the Driver again after it has been disable (switch between text and graphics mode currently this is SVGA only)
```
quajak, line 758: Please also give a summary for the method for documentation

quajak (approved):

PR #1729, opened 2021-04-20
Canvas bug fixes/ImageCanvas support

napalmtorch:
Added color overload for all canvas functions
Fixed un-implemented functions
Option to manually select driver
Option to disable debugging
Horizontal and vertical font spacing
ImageCanvas for manipulating bitmaps


source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -1,398 +1,169 @@
...
+        /// Get and set active video mode
+        /// </summary>
+        private Mode mode;
+        public override Mode Mode
+        {
+            get { return mode; }
+            set
+            {
+                mode = value;
+                if (SendDebug) { debugger.SendInternal($"Set mode property to {value}"); }
```
quajak, line 73: I dont really see the need for SendDebug. Most if not all debug statements should be able to be removed once the code has been finished. The rest can be toggled via COSMOSDEBUG. So I think its better if you remove them.

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -1,398 +1,169 @@
...
+
+        /// <summary>
+        /// Draw point with color at specified position
+        /// </summary>
+        /// <param name="color"></param>
+        /// <param name="x"></param>
+        /// <param name="y"></param>
+        public override void DrawPoint(Color color, int x, int y)
+        {
+            if (x < 0 || x >= Mode.Width || y < 0 || y >= Mode.Height) { if (SendDebug) { debugger.SendInternal("Tried to draw pixel out of bounds at " + x.ToString() + ", " + y.ToString()); } return; }
```
quajak, line 141: Is there a reason why DrawPoint does not just throw an error when trying to draw at an invalid location?

quajak: The same goes for the remaining functions in this file.

source/Cosmos.System2/Graphics/VBECanvas.cs
```diff
@@ -1,392 +1,162 @@
...
+namespace Cosmos.System.Graphics
+{
+    public class VBECanvas : Canvas
+    {
+         // debugger
+        internal Debugger debugger = new Debugger("System", "VBECanvas");
+
+        // driver
+        private VBEDriver driver;
+        private Color pointCol;
```
quajak, line 17: Please remove this field

source/Cosmos.System2/Graphics/VBECanvas.cs
```diff
@@ -1,392 +1,162 @@
...
+        /// Get and set active video mode
+        /// </summary>
+        private Mode mode;
+        public override Mode Mode
+        {
+            get { return mode; }
+            set
+            {
+                mode = value;
+                if (SendDebug) { debugger.SendInternal($"Set mode property to {value}"); }
```
quajak, line 55: Same idea as before, please remove this debug line.

source/Cosmos.System2/Graphics/VBECanvas.cs
```diff
@@ -1,392 +1,162 @@
...
+        /// Create new VBE/VESA canvas with specified mode
+        /// </summary>
+        /// <param name="mode"></param>
+        public VBECanvas(Mode mode) { this.Mode = mode; }
+
+        // set graphics mode
+        private void SetMode(Mode mode)
+        {
+            this.mode = mode;
+            if (!CheckIfModeIsValid(mode)) { throw new Exception($"Unsupported mode {mode}"); }
```
quajak, line 75: Make this a NotSupportedException

source/Cosmos.System2/Graphics/VGACanvas.cs
```diff
@@ -1,417 +1,124 @@
...
+        /// Create new VGA canvas with specified mode
+        /// </summary>
+        /// <param name="mode"></param>
+        public VGACanvas(Mode mode) { this.Mode = mode; }
+
+        // set graphics mode
+        private void SetMode(Mode mode)
+        {
+            this.mode = mode;
+            if (!CheckIfModeIsValid(mode)) { throw new Exception($"Unsupported mode {mode}"); }
```
quajak, line 62: Fix Exception type

source/Cosmos.System2/Graphics/VGACanvas.cs
```diff
@@ -1,417 +1,124 @@
...
+            if (driver == null) { driver = new VGADriver(); }
+            if (mode.Width == 720) { driver.SetGraphicsMode(VGADriver.ScreenSize.Size720x480, VGADriver.ColorDepth.BitDepth4); }
```
quajak, line 64: Please use a switch

quajak (changes requested):
Thank you for the PR and great work! Most of the changes I requested are just code style and minor nitpicks. Overall please don't reformat the entire files in these cases and please figure out a way to do undo that change.  Please also don't cram multiple statements on a single line. While I understand that having debug output during normal execution might seem nice, I don't really see the point. While working on the driver directly we can add them but unless the debug output indicates that something went wrong, it shouldn't stay in the final code.

Tests/Kernels/GraphicTest/Kernel.cs
```diff
@@ -3,7 +3,6 @@
 using Cosmos.TestRunner;
```
quajak, line 3: Can you please also add a few tests for ImageCanvas especially? Maybe call the standard draw tests on a few ImageCanvases.

source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -1,1202 +1,1498 @@
...
+        /// </summary>
+        public abstract Mode Mode { get; set; }
+
+        /// <summary>
+        /// Toggle the use of debugging - potentially faster when disabled
+        /// </summary>
+        public bool SendDebug = true;
+
+        // temporary variables
+        private int arrayX, arrayY;
```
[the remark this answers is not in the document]
napalmtorch: Reasoning for this is because of Cosmos Memory Management causing the OS to crash faster when more locals are being used. I didn't plan on this being a permanent solution, just a temporary one until Memory Management is working properly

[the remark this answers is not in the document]
quajak: `int` is a value type and is therefor not stored in ram but in the stack. I am also positive that we get simple garbage collection working within a month or two, so going for the permanent solution would be better.

PR #1734, opened 2021-04-23
MemoryManager

KM198912:
Use the new MemoryManager using RAT, implement several new methods that make life easier. Introduce a quick Free method to be used in hotspots that create many objects until a GC is in place.

Needs https://github.com/CosmosOS/IL2CPU/pull/103


source/Cosmos.Core/CPU.cs
```diff
@@ -11,15 +11,33 @@ namespace Cosmos.Core
...
+        /// Get amount of RAM in MB's.
+        /// </summary>
+        public static uint GetAmountOfRAM()
+        {
+            if (MemoryMapExists())
+            {
+                var memoryMap = GetMemoryMap();
+                for (int i = 0; i < memoryMap.Length; i++)
+                {
+                    if (memoryMap[i].Type == 1 && memoryMap[i].Length != 0)
```
zarlo, line 29: this `memoryMap[i].Length != 0` is not need (and just needs more clock cycles)
as if it is 0 then it would just be `totalRam = totalRam + 0` as X + 0 = X

source/Cosmos.Core/GCImplementation.cs
```diff
@@ -1,66 +1,126 @@
...
+        /// </summary>
+        /// <returns>Returns the used PageSize by the MemoryManager in Bytes.</returns>
+        public static uint GetUsedRAM()
+        {
+            return (Memory.RAT.TotalPageCount - Memory.RAT.GetPageCount(Memory.RAT.PageType.Empty)) * Memory.RAT.PageSize;
+        }
+        /// <summary>
+        /// Initialise the Memory Manager, this should not be called anymore since it is done very early during the boot process.
+        /// </summary>
+        public static unsafe void Init()
```
zarlo, line 94: may want to make this internal not public since it should not be called from out side of Cosmos's boot precess

zarlo (commented):
we may want to stay with Native over uint as as Native on a 64 bit system should be a 64 bit number and on a 32bit it will be 32 bit

source/Cosmos.Core/GCImplementation.cs
JuliusCaesar12: Making this private would prevent it from being called from the boot process as well considering that the bootstrap is not in this class

source/Cosmos.Core/CPU.cs
AcaiBerii: Seems unneccesary ofc but there are reasons why we may need it.

source/Cosmos.Core/GCImplementation.cs
zarlo: if it was internal (can only be call by code in the same dll) it can still be called

MishaTY:
Sorry, that was me trying to push something to my fork, but it ended up here.

quajak:
Most likely depends on https://github.com/CosmosOS/IL2CPU/pull/135

source/Cosmos.Core/CPU.cs
```diff
@@ -404,25 +404,41 @@ public static unsafe MemoryMapBlock[] GetMemoryMap()
...
         {
-            MemoryMapBlock bestMap = null;
-            ulong bestSize = 0;
-            var maps = GetMemoryMap();
+            if (!Multiboot2.MemoryMapExists())
+            {
+                return null;
+            }
+
+            var rawMap = new RawMemoryMapBlock[64];
```
quajak, line 414: Do we still need the `rawMap` structure?

quajak (changes requested):

source/Cosmos.Core/Cosmos.Core.csproj
```diff
@@ -18,6 +18,7 @@
     </ItemGroup>
 
     <ItemGroup>
+        <ProjectReference Include="$(IL2CPURepoRoot)source\Cosmos.IL2CPU\Cosmos.IL2CPU.csproj" />
```
MishaTY, line 21: I don't think we should reference IL2CPU inside of cosmos.

MishaTY:
Fixes: https://github.com/CosmosOS/Cosmos/issues/1780

quajak (approved):

PR #1736, opened 2021-04-24
Fix IRQs

MishaTY:
This PR fixes a problem where IRQs don't get called because they are masked. Right now, Cosmos just masks all interrupts, expect for the keyboard and mouse, which could cause problems with drivers.

### DIFF
--- a/source/Cosmos.Core/PIC.cs
+++ b/source/Cosmos.Core/PIC.cs
@@ -42,7 +42,7 @@ namespace Cosmos.Core
         /// </summary>
         public void EoiMaster()
         {
-            Master.Cmd.Byte = (byte) Cmd.EOI;
+            Master.Cmd.Byte = (byte)Cmd.EOI;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


charlesbetros:
These were masked due to an issue (on VMware I think) with stack overflow. Specifically the problem was with PIT. I am not able to duplicate the problem on Bochs, but this needs to be checked on VMware before we unmask.

MishaTY:
@charlesbetros I checked the PIT timer, and it does not do a stack overflow in VMware.

charlesbetros:
We need to do a bit of investigation before unmasking these. Thanks for being patient. :)

Arawn-Davies:
I can confirm I'm currently getting stack overflow with PIT on VMware. Anytime System.Threading.Thread.Sleep(ms); or a PIT-specific method is getting called :/ 


On a side note, as PIT is in same ring as PCI, we could do a check in the class in case it's just VMware's crap implementation of the Programmable Interrupt Timer that is causing the problem, but more investigation definitely needs to be done before deciding to skip PIT functions on VMware guests completely 👍🏽 

could be a completely abstract issue for all we know atm :)

MishaTY:
> I can confirm I'm currently getting stack overflow with PIT on VMware. Anytime System.Threading.Thread.Sleep(ms); or a PIT-specific method is getting called :/
> 
> On a side note, as PIT is in same ring as PCI, we could do a check in the class in case it's just VMware's crap implementation of the Programmable Interrupt Timer that is causing the problem, but more investigation definitely needs to be done before deciding to skip PIT functions on VMware guests completely 👍🏽
> 
> could be a completely abstract issue for all we know atm :)

That was caused by a bad plug, it is fixed now.

PR #1750, opened 2021-05-02
Fix ATA Device name decoding

MishaTY:
Fix ATA device model/firmware version/serial number decoding. With the fix these values are correct and not garbage.
Screenshot:
Screenshot

### DIFF
--- a/source/Cosmos.HAL2/BlockDevice/AtaPio.cs
+++ b/source/Cosmos.HAL2/BlockDevice/AtaPio.cs
@@ -253,9 +253,9 @@ namespace Cosmos.HAL.BlockDevice
 		protected string GetString(ushort[] aBuffer, int aIndexStart, int aStringLength)
 		{
             //Convert ushort[] to byte[]
-            byte[] array = new byte[aStringLength / 2];
+            byte[] array = new byte[aStringLength];
             int counter = 0;
-            for (int i = aIndexStart; i < aIndexStart + (aStringLength/2); i++)
+            for (int i = aIndexStart; i < aIndexStart + (aStringLength / 2); i++)
             {
                 var item = aBuffer[i];
                 var bytes = BitConverter.GetBytes(item);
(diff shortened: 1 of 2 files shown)


source/Cosmos.HAL2/BlockDevice/AtaPio.cs
```diff
@@ -250,20 +250,22 @@ public Status SendCmd(Cmd aCmd, bool aThrowOnError)
 			return xStatus;
 		}
 
-		protected string GetString(UInt16[] aBuffer, int aIndexStart, int aStringLength)
+		protected string GetString(ushort[] aBuffer, int aIndexStart, int aStringLength)
 		{
-			// Would be nice to convert to byte[] and use
-			// new string(ASCIIEncoding.ASCII.GetChars(xBytes));
-			// But it requires some code Cosmos doesnt support yet
-			var xChars = new char[aStringLength];
```
quajak, line 258: Can you please keep the old array based approach and add the BitConverter fix to it?
I think it makes more sense to use an array, since we know the length of the string.

source/Cosmos.HAL2/BlockDevice/AtaPio.cs
```diff
@@ -250,20 +250,22 @@ public Status SendCmd(Cmd aCmd, bool aThrowOnError)
...
-			}
-			return new string(xChars);
-		}
+            //Convert ushort[] to byte[]
+            var array = new List<byte>();
+            for (int i = aIndexStart; i < aIndexStart + (aStringLength/2); i++)
+            {
+                var item = aBuffer[i];
+                var bytes = BitConverter.GetBytes(item);
+                for (int i2 = 1; i2 >= 0; i2--)
```
quajak, line 261: I think the intent is clearer if you just do `array.Add(bytes[1]); array.Add(bytes[0]);` than use a for loop.

quajak (changes requested):
Thanks for the PR! I think we should keep the code a bit closer to the original, unless there is a good reason to switch to a list. 
Also do you think we can add a test to the testrunner to ensure that the code works and doesn't break in the future?

source/Cosmos.HAL2/BlockDevice/AtaPio.cs
MishaTY: I don't even know why I did not even think about that.

quajak (approved):

PR #1779, opened 2021-05-20
Add ATAPI driver

MishaTY:
Adds a basic ATAPI driver.

TODO
- [x] Testing on real hardware (none of my hardware supports IDE but sata)
- [x] Test on QEMU
- [x] Test on Virtualbox
- [x] Test on Bochs

### DIFF
--- a/source/Cosmos.HAL2/BlockDevice/IDE.cs
+++ b/source/Cosmos.HAL2/BlockDevice/IDE.cs
@@ -26,17 +26,17 @@ namespace Cosmos.HAL.BlockDevice
         private static void Initialize(Ata.ControllerIdEnum aControllerID, Ata.BusPositionEnum aBusPosition)
         {
             var xIO = aControllerID == Ata.ControllerIdEnum.Primary ? Core.Global.BaseIOGroups.ATA1 : Core.Global.BaseIOGroups.ATA2;
-            var xATA = new AtaPio(xIO, aControllerID, aBusPosition);
-            if (xATA.DriveType == AtaPio.SpecLevel.Null)
+            var xATA = new ATA_PIO(xIO, aControllerID, aBusPosition);
+            if (xATA.DriveType == ATA_PIO.SpecLevel.Null)
             {
                 return;
             }
-            else if (xATA.DriveType == AtaPio.SpecLevel.ATA)
+            else if (xATA.DriveType == ATA_PIO.SpecLevel.ATA)
             {
                 BlockDevice.Devices.Add(xATA);
                 Ata.AtaDebugger.Send("ATA device with speclevel ATA found.");
             }
-            else if (xATA.DriveType == AtaPio.SpecLevel.ATAPI)
+            else if (xATA.DriveType == ATA_PIO.SpecLevel.ATAPI)
             {
                 var atapi = new ATAPI(xATA);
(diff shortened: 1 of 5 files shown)


source/Cosmos.HAL2/BlockDevice/IDE.cs
```diff
@@ -28,15 +28,19 @@ private static void Initialize(Ata.ControllerIdEnum aControllerID, Ata.BusPositi
...
             {
                 BlockDevice.Devices.Add(xATA);
                 Ata.AtaDebugger.Send("ATA device with speclevel ATA found.");
             }
             else if (xATA.DriveType == AtaPio.SpecLevel.ATAPI)
             {
-                Ata.AtaDebugger.Send("ATA device with speclevel ATAPI found, which is not supported yet!");
+                var atapi = new ATAPI(xATA);
+                BlockDevice.Devices.Add(atapi);
+                Ata.AtaDebugger.Send("ATA device with speclevel ATAPI found");
```
MishaTY, line 43: Should we add the ATAPI to the BlockDevices array?

charlesbetros (changes requested):
Please add the comments back. Especially when a type is public

MishaTY:
I can't test on real hardware because the dvd drive nor the hard drive is not showing up in the Blockdevice.Devices

Arawn-Davies:
@charlesbetros have you tested on real hardware or can I do it? been meaning to set up a real debugging host and this seems perfect for it!

charlesbetros:
I haven't tested real hardware. Not too concerned with that though. Feel free to test it.

I would like Atapio and Atapi to both inherit the ATA base class though before merging. They share a lot of code and constants and IO ports. The shared bits should be in the base class.

Arawn-Davies:
@charlesbetros ATAPI and AtaPio are both currently inheriting BlockDevice which is what at the end of the day the System2.FatFileSystem methods call to read specific sectors. C# doesn't have Multiple Inheritance like Perl so I'm not entirely sure whether to inherit Core.ATA, which is just an IOGroup with channel and controller stuff (Primary/Secondary,Master/Slave) with similar polling Wait() methods, or just keep as is

Both do have an ATA IOGroup instance of an object in each class, which I suppose is how the inheritance is done atm. I think for future USB XHCI block devices, best leave ATA code out of that can of worms 😂 :)

Arawn-Davies:
Personally I consider this ready to merge, but of course only when everyone's ok :)
This will need some TestRunner tests being done, but that could be done in a separate PR

Arawn-Davies:
Personally I consider this ready to merge, but of course only when everyone's ok :)
This will need some TestRunner tests being done, but that could be done in a separate PR

charlesbetros:
Sounds good. Merge away if it looks good.

Arawn-Davies (approved):
XML comments have been added back in the latest commit. All good to go!

Arawn-Davies:
> Sounds good. Merge away if it looks good.
think you need to finish your review, can't merge :)

charlesbetros (approved):

MishaTY:
That was github, did not mean to merge.

quajak:
Was there something left to do?

MishaTY:
Never mind, I thought that github dismissed the reviews, but @charlesbetros did.

PR #1792, opened 2021-05-27
Tcp Implementation

valentinbreiz:
https://github.com/CosmosOS/Cosmos/issues/1781

- [x] Add TcpListener class
- [x] Move tcp state machine to Tcp.cs
- [x] Documentation

### DIFF
--- a/Docs/articles/Kernel/Network.md
+++ b/Docs/articles/Kernel/Network.md
@@ -51,6 +51,26 @@ using(var xClient = new UdpClient(4242))
 
 ## TCP
 Like UDP, TCP has to create a client and call Connect() to specify the remote machine address before sending or receiving data.
+
+Server :
+```csharp
+using(var xServer = new TcpListener(4242))
+{
+    /** Start server **/
+    xServer.Start();
+    
+    /** Accept incoming TCP connection **/
+    var client = xServer.AcceptTcpClient(); //blocking
+    
+    /** Stop server **/
+    xServer.Stop();
+
+    /** Send data **/
+    client.Send(Encoding.ASCII.GetBytes(message));
+}
+```
+
+Client :
 ```csharp
 using(var xClient = new TcpClient(4242))
 {
--- a/source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
+++ b/source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
@@ -46,12 +46,16 @@ namespace Cosmos.System.Network.IPv4.TCP
         /// <returns>TcpClient</returns>
         internal static TcpClient GetClient(ushort destPort)
         {
-            if (clients.ContainsKey((uint)destPort))
-            {
-                return clients[(uint)destPort];
-            }
+            TcpClient client;
 
-            return null;
+            if (clients.TryGetValue(destPort, out client))
+            {
+                return client;
+            }
+            else
+            {
+                return null;
+            }
         }
 
         /// <summary>
--- a/source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
+++ b/source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
@@ -240,8 +240,8 @@ namespace Cosmos.System.Network.IPv4.TCP
             source.address = packet.SourceIP;
             source.port = packet.SourcePort;
 
-            var tmp = StateMachine.data;
-            StateMachine.data = null;
+            var tmp = StateMachine.Data;
+            StateMachine.Data = null;
             return tmp;
         }
 
@@ -264,8 +264,8 @@ namespace Cosmos.System.Network.IPv4.TCP
             source.address = packet.SourceIP;
             source.port = packet.SourcePort;
 
-            var tmp = StateMachine.data;
-            StateMachine.data = null;
+            var tmp = StateMachine.Data;
+            StateMachine.Data = null;
             return tmp;
         }
 
--- a/source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
+++ b/source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
@@ -79,9 +79,9 @@ namespace Cosmos.System.Network.IPv4.TCP
         /// <exception cref="ArgumentException">Thrown if localPort already exists.</exception>
         public TcpClient(int localPort)
         {
-            StateMachine.rxBuffer = new Queue<TCPPacket>(8);
-
             StateMachine = new Tcp();
+
+            StateMachine.rxBuffer = new Queue<TCPPacket>(8);
 
             StateMachine.Status = Status.CLOSED;
             StateMachine.LastSequenceNumber = 0;
--- a/source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
+++ b/source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
@@ -61,8 +61,6 @@ namespace Cosmos.System.Network.IPv4.TCP
         /// <exception cref="ArgumentException">Thrown if localPort already exists.</exception>
         internal TcpClient(Tcp stateMachine)
         {
-            StateMachine.rxBuffer = new Queue<TCPPacket>(8);
-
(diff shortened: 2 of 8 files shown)


source/Cosmos.System2/Network/IPv4/TCP/TCPPacket.cs
```diff
@@ -409,14 +415,28 @@ internal byte[] TCP_Data
             }
         }
 
+        /// <summary>
+        /// Convert a byte to a hexadecimal char.
+        /// </summary>
+        /// <param name="b"></param>
+        /// <returns>Hex string.</returns>
+        static string ToHex(byte b)
```
quajak, line 423: There should already be a method to do this `.ToString("X")` on a byte should return the hexadecimal representation.

source/Cosmos.System2/Network/IPv4/TCP/Tcp.cs
```diff
@@ -0,0 +1,528 @@
...
+        FIN_WAIT2,
+        CLOSE_WAIT,
+        CLOSING,
+        LAST_ACK,
+        TIME_WAIT,
+        CLOSED
+    }
+
+    /// <summary>
+    /// Tcp class. Used to manage the TCP state machine
```
quajak, line 28: Could you expand on the comment a bit more about how it is used by the rest of the code, so its easier for someone reading the code to understand the class in the context of the network stack.

source/Cosmos.System2/Network/IPv4/TCP/Tcp.cs
```diff
@@ -0,0 +1,528 @@
...
+            OutgoingBuffer.AddPacket(packet);
+            NetworkStack.Update();
+        }
+
+        /// <summary>
+        /// Contatenate two byte arrays. https://stackoverflow.com/a/45531730
+        /// </summary>
+        /// <param name="first">First byte array.</param>
+        /// <param name="second">Byte array to concatenate.</param>
+        private byte[] Concat(byte[] first, byte[] second)
```
quajak, line 474: Can you move this to an internal ArrayHelper class in System2?

source/Cosmos.System2/Network/IPv4/TCP/Tcp.cs
```diff
@@ -0,0 +1,528 @@
...
+            }
+            return output;
+        }
+
+        /// <summary>
+        /// Split byte array into chunks of a specified size.
+        /// </summary>
+        /// <param name="buffer">Byte array to split.</param>
+        /// <param name="chunksize">Chunk size.</param>
+        internal byte[][] ArraySplit(byte[] buffer, int chunksize = 1000)
```
quajak, line 504: Same with this one

source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
```diff
@@ -15,78 +15,19 @@
...
-        /// <summary>
-        /// Last recveived Connection Sequence number.
+        /// Clients dictionary.
         /// </summary>
-        private uint LastSequenceNumber;
-
-        private bool WaitingAck = false;
-
-        private byte[] data;
+        private static Dictionary<uint, TcpClient> clients;
```
quajak, line 30: Do we need the list of clients here? Wouldn't it be better if every Listener keeps a list of clients which connected via it?
Looking at the code, I don't really see the need for the Dictionary here.

quajak (changes requested):

source/Cosmos.System2/Network/IPv4/TCP/TCPPacket.cs
```diff
@@ -52,14 +52,20 @@ internal static void TCPHandler(byte[] packetData)
         {
             var packet = new TCPPacket(packetData);
 
-            Global.mDebugger.Send("[Received] TCP packet from " + packet.SourceIP.ToString() + ":" + packet.SourcePort.ToString());
-
             if (packet.CheckCRC())
             {
-                var receiver = TcpClient.GetClient(packet.DestinationPort);
-                if (receiver != null)
+                var client = TcpClient.GetClient(packet.DestinationPort);
```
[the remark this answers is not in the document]
valentinbreiz: When there is no TcpClient with the same port as the packet destination port

source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
valentinbreiz: It's a static list containing all tcp clients linked by their ports, it's necessary to handle multiple connections

source/Cosmos.System2/Network/IPv4/TCP/TCPPacket.cs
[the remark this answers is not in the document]
quajak: Is that something which should happen under normal usage? Should the os be aware that it happened or can we just be quiet in that case?

source/Cosmos.System2/Network/IPv4/TCP/TCPPacket.cs
quajak: Ah, which one? This is something which should be plugged. I thought I had worked on similar plugs a short while ago.

source/Cosmos.System2/Network/IPv4/TCP/TCPClient.cs
quajak: From where is it used?

valentinbreiz: When a new packet is received here: https://github.com/CosmosOS/Cosmos/blob/feature/tcp/source/Cosmos.System2/Network/IPv4/TCP/TCPPacket.cs#L57

Docs/articles/Kernel/Network.md
```diff
@@ -51,6 +51,26 @@ using(var xClient = new UdpClient(4242))
...
+using(var xServer = new TcpListener(4242))
+{
+    /** Start server **/
+    xServer.Start();
+    
+    /** Accept incoming TCP connection **/
+    var client = xServer.AcceptTcpClient(); //blocking
+    
+    /** Stop server **/
+    xServer.Stop();
```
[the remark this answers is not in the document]
valentinbreiz: Stop is called via Dispose indeed but it's to show all possible method calls inside the class

source/Cosmos.System2/Network/IPv4/TCP/TCPPacket.cs
valentinbreiz: Need plug for: System.String  System.Number.FormatInt32(System.Int32, System.String, System.Globalization.NumberFormatInfo).
04:05:29.380991         Error:   Called from :
04:05:29.381530         Error: System.Byte::System.String ToString(System.String)

source/Cosmos.System2/Network/IPv4/TCP/TCPPacket.cs
```diff
@@ -52,14 +52,20 @@ internal static void TCPHandler(byte[] packetData)
...
+                if (client != null)
                 {
-                    receiver.ReceiveData(packet);
+                    client.StateMachine.ReceiveData(packet);
+                    return;
+                }
+
+                var listener = TcpListener.GetListener(packet.DestinationPort);
+                if (listener != null)
```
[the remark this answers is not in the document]
valentinbreiz: If it's null, it means that no tcp port is opened and the received packet will be passed since no connection is opened

Docs/articles/Kernel/Network.md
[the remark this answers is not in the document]
quajak: I think it would be better if you mention that you should either use a `using` block or call Stop explicitly. The current example is confusing and would encourage incorrect behavior.

source/Cosmos.System2/Network/IPv4/TCP/TCPPacket.cs
quajak: Tracked in https://github.com/CosmosOS/Cosmos/issues/1797

quajak (approved):

PR #1799, opened 2021-06-02
FTP Server Implementation + TCB Implementation

valentinbreiz:
TODO:
- [x] Multiple TCP connection support
- [x] fix path
- [x] fix folder with space
- [x] fix PASV mode

### DIFF
--- a/source/Cosmos.System2/Network/IPv4/TCP/FTP/FtpServer.cs
+++ b/source/Cosmos.System2/Network/IPv4/TCP/FTP/FtpServer.cs
@@ -285,10 +285,10 @@ namespace Cosmos.System.Network.IPv4.TCP.FTP
             {
                 try
                 {
-                    if (Directory.Exists(CurrentDirectory + command.Content))
+                    if (Directory.Exists(CurrentDirectory + "\\" + command.Content))
                     {
+                        CurrentDirectory = CurrentDirectory + "\\" + command.Content;
                         Directory.SetCurrentDirectory(CurrentDirectory);
-                        CurrentDirectory = CurrentDirectory + command.Content + @"\";
                     }
                     else if (File.Exists(CurrentDirectory + command.Content))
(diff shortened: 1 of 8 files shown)


bad-codr:
how i use

valentinbreiz:
When it will be ready @bad-codr ;)

quajak (approved):

source/Cosmos.System2/Network/IPv4/TCP/FTP/FtpServer.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// </summary>
+        public void Listen()
+        {
+            while (Listening)
+            {
+                tcpListener = new TcpListener(21);
+                tcpListener.Start();
+                var client = tcpListener.AcceptTcpClient();
+
+                global::System.Console.WriteLine("Client : New connection from " + client.StateMachine.LocalAddress.ToString());
```
quajak, line 99: User should be able to disable this

source/Cosmos.System2/Network/IPv4/TCP/FTP/FtpServer.cs
```diff
@@ -0,0 +1,180 @@
...
+        private void ReceiveRequest(FtpClient ftpClient)
+        {
+            var ep = new EndPoint(Address.Zero, 0);
+
+            try
+            {
+                var data = Encoding.ASCII.GetString(ftpClient.Control.Receive(ref ep));
+                data = data.Remove(data.Length - 2, 2);
+
+                global::System.Console.WriteLine("Client : " + data);
```
quajak, line 139: Same here. In future it would be best if we could use events to notify the OS that something happened but I dont know if they are currently supported

quajak (changes requested):

quajak (approved):

PR #1807, opened 2021-06-06
[Documentation] Improve VFS article

AnErrupTion:
I've kind of rewrote the VFS article to be more clear, have the example code pieces more consistent, and also added a new section for formatting the drive. Some typos have also been fixed but there might also still be some.

### DIFF
--- a/Docs/articles/Kernel/VFS.md
+++ b/Docs/articles/Kernel/VFS.md
@@ -31,6 +31,8 @@ After our VFS has been initialized, we can use more interesting functions, let's
 
 **Note**: You don't have to format your drive if you're debugging your Cosmos project with VMWare. The build will automatically add an already formatted FAT32 VMDK file for your convenience.
 
+**Note 2**: You can only format a drive that already has been formatted with FAT32.
+
 You can format your drive with the Format() function, just like this:
 
 ```C#
--- a/Docs/articles/Kernel/VFS.md
+++ b/Docs/articles/Kernel/VFS.md
@@ -3,56 +3,73 @@
 In this article we will discuss about using Cosmos VFS (virtual file system).
 Cosmos VFS and the VFS manager classes, let you manage your file system.
 
+**Attention**: **Always** format your drive with Cosmos and **only** Cosmos if you plan to use it with Cosmos. Using any other tool such as Parted, FDisk (or any other tool) might lead to weird things when using that drive with Cosmos' VFS. Those tools are much more advanced and might format and read/write to the disk differently than Cosmos.
+
 First, we should create and initialize an instance of the VFS, this will initialize the partition and files-system lists, as well as register the new VFS.
 This is essential for using the VFS.
 
-We start with creating global CosmosVFS, this line should appear outside of any function, and before the BeforeRun() function.  
+We start with creating a global CosmosVFS, this line should appear outside of any function, and before the BeforeRun() function.
 
 ```C#
-CosmosVFS fs = new Sys.FileSystem.CosmosVFS();
+var fs = new Sys.FileSystem.CosmosVFS();
 ```
 
-Next, we register our VFS at the VFS manager, this will initiate the VFS and make it usable, add this to your kernel's BeforeRun():
+Next, we register our VFS at the VFS manager, this will initiate the VFS and make it usable, add this to your kernel's BeforeRun() function:
 
 ```C#
 Sys.FileSystem.VFS.VFSManager.RegisterVFS(fs);
 ```
 
-After the initialization process is done, a message like this would apper on your screen:
+After the initialization process is done, a message like this would appear on your screen:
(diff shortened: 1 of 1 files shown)


Docs/articles/Kernel/VFS.md
```diff
@@ -3,56 +3,73 @@
...
-This message is printed by the register method and it provides info about the file system.
+This message is printed by the RegisterVFS() method and it provides info about the partition.
+
+After our VFS has been initialized, we can use more interesting functions, let's go over some of them:
+
+## Format drive
+
+**Note**: You don't have to format your drive if you're debugging your Cosmos project with VMWare. The build will automatically add an already formatted FAT32 VMDK file for your convenience.
+
+You can format your drive with the Format() function, just like this:
```
MishaTY, line 34: Please add that Cosmos can only format the drive that already has FAT32.

AnErrupTion: Really? Well I didn't know that. I'll add that.

Arawn-Davies:
Just wondering, why are you using the VFS methods instead of the plugged System.IO methods, kinda defeats the purpose of them 😂really I think the documentation should include the minimum needed VFS partitioning and initialization methods, then try and use System.IO where possible

AnErrupTion:
The original article used them, and also why not?

quajak:
I think its better to use the actual system.io methods.

Arawn-Davies (changes requested):
> The original article used them, and also why not?

Because, as stated around many places, the articles were written a long time ago before System.IO was plugged. Plugging the .NET framework (along with assemblies like System, System.IO) has long been one of the goals of Cosmos, so we should be using them.

Please try and update the PR to use the System.IO methods as opposed to VFS methods.
This way, if developers want to implement their own custom FS like my Siaranite FS
 [Siaranite-Solutions](https://github.com/Siaranite-Solutions/SDFS) , we can just plug System.IO instead of having to go a level lower using VFS.

Docs/articles/Kernel/VFS.md
```diff
@@ -71,36 +90,38 @@ foreach (var directoryEntry in directory_list)
...
 
-This one is more tricky,
-We need to get a directoryEntryList, find files in the list and print the content to the screen.
+This one is a bit more tricky. We'll need to use our list of DirectoryEntry, loop through all files and print their raw content to the standard output.
 
-of course, we start with geting the directory listing;
+Of course, we'll start with geting the list of DirectoryEntry:
 
 ```C#
 var directory_list = fs.GetDirectoryListing(@"0:\");
```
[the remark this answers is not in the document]
AnErrupTion: Has FileInfo been plugged yet or not?

AnErrupTion:
Done, I modified the article.

Arawn-Davies:
> Done, I modified the article.

cheers mate, we're just updating gocosmos.org's code highlighting issues but I'll get the PR reviewed after :)

Docs/articles/Kernel/VFS.md
```diff
@@ -72,122 +91,86 @@ foreach (var directoryEntry in directory_list)
...
 
-## Write to a file
+## Write to file
 
-Now we will write to an existing file.  
-Writing to a file is almost the same as reading from a file.  
-Always remember, we should use try catch block.  
+Now we will write to an existing file.
+Writing to a file is almost the same as reading from a file.
+Always remember that we should put our code in a try catch block.
```
Arawn-Davies, line 147: Could we eliminate some of these try {} catch (Exception ex) {} blocks? 
Any C# user should know that if a null parameter is passed to a method (in this case a non-existent file or directory), it's going to throw an exception.

If you just add a reminder about that at the beginning of the article that will suffice, In practice and for code cleanliness, I don't think it's a good idea to add a try-catch block every time a System.IO method is called. From personal experience, it's better to write project-specific FS functions that can catch these exceptions, as an operating system kernel will be highly frequently accessing IO. Maybe add a note for that too :+1:

Arawn-Davies (changes requested):
Aside from this wee bit, it's a great article, thanks! When changed I'll go ahead and merge

Docs/articles/Kernel/VFS.md
quajak: The problem is that currently the fs is not very stable and very easy to use incorrectly, which means if you dont use try catch blocks, errors will just cause the kernel to crash, making it a lot harder to debug. So I think its fine if we for now encourage users to use try catch blocks often.

Arawn-Davies: that's fair enough then :)  I'll go ahead and merge

Arawn-Davies (approved):

Arawn-Davies:
@AnErrupTion can you remerge master into your fork? Won't let me merge until it's up-to-date again

AnErrupTion:
Done!

Arawn-Davies:
@valentinbreiz I've reinstated auto-merge so this PR will merge as soon as @AnErrupTion updates his fork.

PR #1829, opened 2021-06-11
Remove Cosmos.HAL.Power.cs "bridge"

AnErrupTion:
The Power class in Cosmos.HAL is completely useless as it's only a sort of a "bridge" to the ACPI and CPU class. What I've did is deleting that class and in the Power class of Cosmos.System, edited the functions to directly use the ACPI and CPU class.
I've also kept the comment about ACPI.Reboot(); not implemented yet.


charlesbetros:
There are multiple rings in cosmos. System cannot access anything lower than HAL. That's why the "pass through" exists.

AnErrupTion:
Ah I see, the rings aren't implemented yet (or were but got scraped) and I didn't remember they existed. I guess I'll close this PR.

PR #1830, opened 2021-06-11
Plug Environment.Exit()

AnErrupTion:
This could be useful for coders familiar with C# and who use Environment.Exit() by accident instead of Power.Shutdown().

Don't mind the other commits, I've reverted them :v (I'm stupid I know)


charlesbetros (approved):
Thanks for the PR! Great idea.

PR #1835, opened 2021-06-16
Update issue templates

quajak:
Include information about the prerelease enabled in the nuget package manager


Arawn-Davies:
I think sometimes projects use the nuget.org package store instead of the Local Cosmos Package Feed which can cause issues, so I've added a checkbox to remind users to ensure that the local store is selected

Arawn-Davies (approved):

PR #1837, opened 2021-06-18
Fix PS2Mouse for Real Hardware

JuliusCaesar12:
The mouse stopped working when a scroll wheel wasn't present.

### DIFF
--- a/source/Cosmos.HAL2/PS2Mouse.cs
+++ b/source/Cosmos.HAL2/PS2Mouse.cs
@@ -75,8 +75,8 @@ namespace Cosmos.HAL
                 mDebugger.SendInternal("(PS/2 Mouse) Mouse ID: " + mMouseID);
             }
 
-            //SendCommand(Command.SetDefaults);
-            //mPS2Controller.WaitForAck();
+            SendCommand(Command.SetDefaults);
+            mPS2Controller.WaitForAck();
 
             INTs.SetIrqHandler(12, HandleMouse);
 
@@ -128,6 +128,13 @@ namespace Cosmos.HAL
                 {
                     mMouseCycle++;
                 }
+                else 
+                {
+                    mMouseByte[0] = 0;
+                    mMouseByte[1] = 0;
+                    mMouseByte[2] = 0;
+                    mMouseByte[3] = 0;
+                }
             }
             else if (mMouseCycle == 1)
             {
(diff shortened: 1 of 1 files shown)


source/Cosmos.HAL2/PS2Mouse.cs
```diff
@@ -137,19 +144,18 @@ public void HandleMouse(ref INTs.IRQContext context)
             else if (mMouseCycle == 2)
             {
                 mMouseByte[2] = IO.Data.Byte;
-
-                if (HasScrollWheel)
```
Arawn-Davies, line 141: Why remove the check for a ScrollWheel?
Not every PS/2 mouse has one, so please put the checks back in place

Arawn-Davies (commented):
Do you have a project I can use to test the changes? I just need something simple and lightweight - the code you used to test this would be ideal 👍🏽

Arawn-Davies (changes requested):
Do you have a project I can use to test the changes? I just need something simple and lightweight - the code you used to test this would be ideal 👍🏽

We do need checks for the mousewheel, as not every PS/2 mouse has one, and there are differences between ps2 and imps2.

JuliusCaesar12:
Re-adding the checks have made the mouse stay at the bottom both in the VM and on my laptop.

JuliusCaesar12:
https://pastebin.com/Y5F709s0
This is a simple test kernel I created to test this (it only uses textmode, so it's not that intensive)

quajak:
Does it mean that we are incorrectly detecting mouse wheels?

JuliusCaesar12:
> Does it mean that we are incorrectly detecting mouse wheels?

I'm not sure. I think it may have something to do with the fact that the trackpad is possibly a Synaptic pad.
I looked at the source code for ReactOS and looked at mouse.c; what I found was that the driver would only switch to finding the z movement if the device was not a generic ps2 device or a ps2pp device.

quajak:
Ok, I suppose we can do something similar. @Arawn-Davies @valentinbreiz @KM198912 thoughts?

KM198912:
I think this could be done and synaptics issues I had before I think there is a push with that

KM198912:
also i think we should make the disabling of that feature a optional user kernel setting since not everyone might have this issue and instead of completly disabling it we should see if we can find a reliable way to detect all buttons, @charlesbetros @quajak thoughts?

quajak:
I think currently the best way would be to set a flag in the mouse driver, which normally is disabled.

MishaTY:
Not needed anymore because of #1989 which allows to disable the scroll wheel.

PR #1843, opened 2021-06-21
Revert Appveyor.yml changes due to API issues

Arawn-Davies:
GitHub doesn't have a 'skipped' build status for CI providers, and only has failure or success statuses (boolean?).  AppVeyor would either need to send a succeed status to the GitHub PR in order for merging to be available, or GitHub need to add a new skipped status in its API. A webhook can be setup using a Personal Access Token to manually set the build status to 'succeeded'/'fail'

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -60,15 +60,15 @@ artifacts:
   - path: 'TestResult.xml'
 
 
-skip_commits:
-  files:
-    - Docs/
-    - .github/
-    - Artwork/
-    - '**/*.md'
-    - LICENSE.txt
-    - README.txt
-    - .gitignore
-    - source/Archive/
-    - Users/
+#skip_commits:
+#  files:
+#    - Docs/
+#    - .github/
+#    - Artwork/
+#    - '**/*.md'
+#    - LICENSE.txt
+#    - README.txt
+#    - .gitignore
+#    - source/Archive/
+#    - Users/
     
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,6 +13,7 @@ init:
 
 before_build:
   - ps: |
+      cd C:\Cosmos
       $range = "$env:APPVEYOR_REPO_BRANCH..$env:APPVEYOR_PULL_REQUEST_HEAD_COMMIT"
       if (-not (git diff --name-only --diff-filter=d $range | Where-Object { $_ -NotMatch "\.md$" })) {
         Write-Host "Only Markdown files were modified, skipping CI run."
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -60,15 +70,15 @@ artifacts:
...
-    - README.txt
-    - .gitignore
-    - source/Archive/
-    - Users/
+#skip_commits:
+#  files:
+#    - Docs/
+#    - .github/
+#    - Artwork/
+#    - '**/*.md'
```
quajak, line 78: Should this one also be commented out?

appveyor.yml
```diff
@@ -5,10 +5,20 @@ platform: Any CPU
...
+
+
 init:
 - cmd: |
     git clone https://github.com/CosmosOS/IL2CPU.git c:\IL2CPU --depth 1
 
+install:
+  - ps: |
+      $range = "$env:APPVEYOR_REPO_BRANCH..$env:APPVEYOR_PULL_REQUEST_HEAD_COMMIT"
+      if (-not (git diff --name-only --diff-filter=d $range | Where-Object { $_ -NotMatch "\.md$" })) {
```
quajak, line 17: Can we extend this to include more files from the skip list below?

quajak (changes requested):

appveyor.yml
Arawn-Davies: I've had to comment out the old excludes because it's causing GitHub to indefinitely wait for a status. I'm currently looking into the GitHub API to set the success status for these kinds of commits involving MD files and more

appveyor.yml
Arawn-Davies: We could, but in as I replied to the other changes requested, I think we need to fix GitHub status checks before we worry about adding more excludes :)

quajak (dismissed):

appveyor.yml
Arawn-Davies: Shallow clone being set to true prevented the install: git diff to work properly that's fixed. so the skip_commits: list isn't needed, only to reference which files should be checked for in a git diff

appveyor.yml
Arawn-Davies: AppVeyor exits the build based on the git diff, so the list can now be extended 👍🏽

Arawn-Davies:
@quajak I've expanded the list of exclusions to include yml, md, txt and .git* files, but I'm not that great with powershell as to regex with string[]s, so at least the minimum is implemented, more refining can be done later 🙂

quajak (approved):
Thanks

Issue #1844, opened 2021-06-21
Fatal Memory Management Problem(Freeze)


nifanfa:
# Heap.MemAlloc(uint aSize) calls every time. but no free. It cause system to freeze.(Out of memory)

nifanfa:
System stuck when calling update

quajak:
I thought we fixed that, are you on the newest devkit?

nifanfa:
> I thought we fixed that, are you on the newest devkit?

no

nifanfa:
@quajak I can't build the newest cosmos can you build it for me?
![QQ截图20210621184653](https://user-images.githubusercontent.com/54205437/122750427-118e6d80-d2c1-11eb-8839-baab8e16653b.png)

nifanfa:
![QQ截图20210621190353](https://user-images.githubusercontent.com/54205437/122752365-71861380-d2c3-11eb-9985-3a08c440fb4f.png)

nifanfa:
@quajak  the system still crashes within a minute

nifanfa:
![QQ截图20210621191334](https://user-images.githubusercontent.com/54205437/122753431-d5f5a280-d2c4-11eb-9bd8-75aced09c5f6.png)

nifanfa:
![QQ截图20210621191643](https://user-images.githubusercontent.com/54205437/122753716-3ab0fd00-d2c5-11eb-906b-2fdc16761c1d.png)

nifanfa:
The code like this will crash in a few minutes

nifanfa:
This can be discovered hardly if Remote Debugging enabled

quajak:
Yes the problem in this case is that Garbage Collection is currently still missing. I am currently working on fixing that.

AcaiBerii:
Congratulations, I can't even get Cosmos Kernels to run with the latest devkit!

nifanfa:
![GIF 2021-6-22 15-14-21](https://user-images.githubusercontent.com/54205437/122918712-fccdda80-d391-11eb-86eb-357c1abb6c58.gif)

nifanfa:
Heap.MemAlloc calls every time. but no free. It cause system to freeze.

nifanfa:
by the way system always calls heap.memalloc(16) i don't know what does 16 mean

nifanfa:
> Yes the problem in this case is that Garbage Collection is currently still missing. I am currently working on fixing that.

[...]

PR #1845, opened 2021-06-21
Draw String With Custom Charset And Custom Fonts.

nifanfa:
QQ截图20210621202723
****
QQ截图20210621195934
# It support Anti-Aliasing(not real) and it's fast. Don't worry about the performance! 
****
QQ截图20210621194656
👉[Generator Here](https://github.com/CosmosOS/Cosmos/files/6686362/Debug.zip)
👉[Source](https://github.com/nifanfa/BitFont)
****
### Actually I want add the generator here but i don't know how to modify it...
QQ截图20210621195205
****
QQ截图20210621195802


nifanfa:
QQ截图20210621202723

nifanfa:
By the way. i think this problem must be solved #1844

AcaiBerii:
How do you use it?

nifanfa:
> How do you use it?

image

elkaamee326:
Does it support the `DoubleBufferedVMwareSVGAII` driver?

nifanfa:
> Does it support the `DoubleBufferedVMwareSVGAII` driver / canvas?

it only support canvas. but if you want you can implement it!

nifanfa:
it just uses DrawPoint()

source/Cosmos.System2/Graphics/Fonts/BitFont.cs
```diff
@@ -0,0 +1,143 @@
...
+        public BitFontDescriptor(string aCharset, MemoryStream aMS, int aSize)
+        {
+            this.Charset = aCharset;
+            this.MS = aMS;
+            this.Size = aSize;
+        }
+    }
+
+    /// <summary>
+    /// For More https://github.com/nifanfa/BitFont
```
Arawn-Davies, line 22: Please don't redirect users to another repo.
 If you want to add further functionality into Cosmos, please update/create an article in the Docs/ subdirectory.

nifanfa: sure. i just forgot to remove it.

terminal-cs (approved):
This new method is very fast and has no downsides, it works very well.

quajak:
Thanks for the PR. It would be very cool to have better font support in Cosmos. Just a few questions:
- Is there any specification for the format of the BitFont? It would be very good if you could add some comment, so that other people know how to create a BitFont.
- When implementing the other font support for Cosmos, I created the abstract class https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.System2/Graphics/Fonts/Font.cs would it be possible to make BitFont implement it? Do you think the interface should be changed?
- For the drawing methods, could they also be part of the Canvas class directly? That would keep the interface consistent. Even better would be if drawing strings using BitFont or PCScreenFont can be done using the same method.

terminal-cs:
i think im noticing an issue with it actualy, the font looks "wobily" and uneven, you can see it in this image, you need to properly align the fonts, but from figure 3 it looks ok
Screenshot 2021-07-16 151904
Screenshot 2021-07-16 151920
Screenshot 2021-07-16 151848

terminal-cs:
nianfa, could you please respond to quajak plez? this is cool and should be implemented.

nifanfa:
> nianfa, could you please respond to quajak plez? this is cool and should be implemented.

Yes. But in fact, there is no difference between bitfont in Mosa and Cosmos

nifanfa:
> i think im noticing an issue with it actualy, the font looks "wobily" and uneven, you can see it in this image, you need to properly align the fonts, but from figure 2 it looks ok Screenshot 2021-07-16 151904 Screenshot 2021-07-16 151920 Screenshot 2021-07-16 151848

That

nifanfa:
> i think im noticing an issue with it actualy, the font looks "wobily" and uneven, you can see it in this image, you need to properly align the fonts, but from figure 2 it looks ok Screenshot 2021-07-16 151904 Screenshot 2021-07-16 151920 Screenshot 2021-07-16 151848

You can simply modify the generator

terminal-cs:
Cool, finaly reopened, but I think this needs to be refined a bit

RSTPlays:
How do I use it? It’s a bit confusing and will it support the command line?

Neil-Ciciglycerin:
Why did you delete the repo?

Neil-Ciciglycerin:
Luckily I already had this merged!

PR #1875, opened 2021-07-12
[vfs] Implement ISO9660 file system

MishaTY:
Implement ISO9660 file system, allowing users to read from a DVD drive


quajak (changes requested):
Looks good. I think we should try add tests so that it doesnt break in future without us noticing.

MishaTY:
Still don't know why tests are failing, looks like this is a bug in the cosmos VFS.

quajak (approved):

PR #1877, opened 2021-07-13
I've updated the UI a bit so it looks a bit less scattered and off

terminal-cs:


terminal-cs:
heres what it looks like
Screenshot 2021-07-12 230217

source/Cosmos.Build.Builder/BuildTasks/CreateSetupTask.cs
```diff
@@ -41,7 +41,7 @@ protected override string GetExePath()
 
             if (!File.Exists(innoSetupCompilerPath))
             {
-                throw new InvalidOperationException("Inno Setup installation detected, but the compiler doesn't exist!");
+                throw new InvalidOperationException("An Inno Setup installation was detected, but the compiler doesn't exist!");
```
quajak, line 44: Maybe add the path to the message, so that its easier to figure out the error when the message is shown.

quajak (changes requested):
Thanks for the PR!

source/Cosmos.Build.Builder/BuildTasks/CreateSetupTask.cs
terminal-cs: I don't know what the error output variable is for this specific area, do you know?

terminal-cs: > Maybe add the path to the message, so that its easier to figure out the error when the message is shown.

plus why would that part need a path output? it is just stating that inno compiler doesnt exist and needs to be installed

quajak: I would add `innoSetupCompilerPath` to the message, since thats whats being checked for. IMO its equally likely that the compiler is not installed or the path is wrong, and if we can destinguish between the two that seems like a win.

terminal-cs: would this work?
Screenshot 2021-07-16 150755

quajak: Yes thats what i had in mind. Although i think you are missing a $ before the string.

terminal-cs: > Yes thats what i had in mind. Although i think you are missing a $ before the string.

Alright, I've pushed it and it's doing appveyor thing

quajak (approved):

PR #1888, opened 2021-07-23
Fix building on Windows 7

AnErrupTion:
Source : https://github.com/NuGet/NuGetGallery/issues/8176#issuecomment-683923724

This enables TLS 1.2 in the registry if a certain update is installed.

### DIFF
--- a/install-VS2019.bat
+++ b/install-VS2019.bat
@@ -1,5 +1,22 @@
 @echo off
+setlocal
+for /f "tokens=4-5 delims=. " %%i in ('ver') do set VERSION=%%i.%%j
 
+if "%version%" == "6.1" goto seven
+
+:seven
+echo Detected Windows 7! Checking for update KB3140245...
+
+for /f "delims=" %%i in ("wmic qfe | find /c ""KB3140245""") do set update=%%i
+if "%update%" == "0" goto update404
+
+reg add "HKLM\SYSTEM\CurrentControlSet\Control\SecurityProviders\SCHANNEL\Protocols\TLS 1.2\Client" /v DisabledByDefault /t REG_DWORD /d 0 /f /reg:32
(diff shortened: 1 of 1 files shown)


install-VS2019.bat
```diff
@@ -1,28 +1,51 @@
...
-
-echo Building Builder.sln
-"%MSBuild%" Builder.sln /nologo /maxcpucount /nodeReuse:false /verbosity:minimal /t:Restore;Build
-
-start "Cosmos Builder" "source\Cosmos.Build.Builder\bin\Debug\Cosmos.Build.Builder.exe" "-VSPATH=%InstallDir%" %*
+@echo off
+setlocal
+for /f "tokens=4-5 delims=. " %%i in ('ver') do set VERSION=%%i.%%j
+
+if "%version%" == "6.1" goto seven
```
quajak, line 5: Should this not check for version != 6.1 and then goto start?

install-VS2019.bat
```diff
@@ -1,28 +1,51 @@
...
+
+:seven
+echo Detected Windows 7! Checking for update KB3140245...
+
+for /f "delims=" %%i in ("wmic qfe | find /c ""KB3140245""") do set update=%%i
+if "%update%" == "0" goto update404
```
quajak, line 11: Can you please use an if block for what to do and not a goto

quajak (changes requested):
Thanks for the PR.

install-VS2019.bat
AnErrupTion: it checks if the NT version is 6.1, if yes then it goes to :seven and checks for a specific update, etc

quajak: My understanding of batch might be wrong, but doesnt the code end up executing the seven block even if the NT version is not 6.1?

AnErrupTion: ah that's true lol, didn't see that one

quajak (approved):

PR #1907, opened 2021-08-17
Update ConsoleImpl.cs to use exceptions

BlitzWolfMatthew:
Added checks for x & y to be at least 0. Also updated error messages to use exceptions.

### DIFF
--- a/source/Cosmos.System2_Plugs/System/ConsoleImpl.cs
+++ b/source/Cosmos.System2_Plugs/System/ConsoleImpl.cs
@@ -36,24 +36,24 @@ namespace Cosmos.System_Plugs.System
 
         public static int get_BufferHeight()
         {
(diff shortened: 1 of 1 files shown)


source/Cosmos.System2_Plugs/System/ConsoleImpl.cs
```diff
@@ -36,24 +36,24 @@ public static void set_BackgroundColor(ConsoleColor value)
 
         public static int get_BufferHeight()
         {
-            WriteLine("Not implemented: get_BufferHeight");
+            throw new NotImplementedException("Not implemented: get_BufferHeight");
             return -1;
```
quajak, line 40: Remove the unnecessary return statements please

quajak (changes requested):
Thanks!

quajak (approved):

PR #1909, opened 2021-08-18
Remove before-after Run() debug lines

BlitzWolfMatthew:
Several users found them annoying. They can be easily added manually by whoever wants them.

### DIFF
--- a/source/Cosmos.System2/Kernel.cs
+++ b/source/Cosmos.System2/Kernel.cs
@@ -113,9 +113,7 @@ namespace Cosmos.System
                 while (!mStopped)
                 {
                     //Network.NetworkStack.Update();
-                    //Global.mDebugger.Send("Really before Run");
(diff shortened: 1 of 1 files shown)


raxracks (approved):
Nothing wrong with commenting them out but I don't understand why they aren't just being completely removed.

BlitzWolfMatthew:
> Nothing wrong with commenting them out but I don't understand why they aren't just being completely removed.

I can remove them if that would make it better ¯\_(ツ)_/¯ , but I saw other commented code that isn't used so I thought that I could just leave it there.
I will just wait until a CoreDev also reviews it.

quajak:
I think its better to remove the code entirely, the code does have lots of commented out parts and I am trying to remove them except if there they are often used when developing to make the code nicer.

quajak (approved):

PR #1938, opened 2021-10-07
Making the build tasks cross-platform

xafero:
It's related to the IL2CPU patches for running the compiler on Non-Windows like Linux and/or Mac.
First step is to look for things like ".exe", ".bat", "\" and replace them with "", ".sh", "/" 
to allow the dotnet build task to work the same way.


zarlo:
would a good next step for this (in its own RP) have it so all the paths (or just more) configurable not just CosmosBuildDir

PR #1951, opened 2021-10-16
Increased speed for drawing circles in vbecanvas

terminal-cs:
title says it all


terminal-cs:
stupid checks, trying again

PR #1989, opened 2021-11-06
Allow disabeling of drivers from user kernel and implement network debugger from Aura (WIP)

KM198912:
This will allow users to disable the loading of the mousewheel/ps2 drivers/network drivers/IDE drivers from their kernel by calling the onboot method and setting booleans in Sys.global.init also it integrates the network debugger from Aura OS which Valentin ported to work with cosmos which will make debugging on real hardware easier once the neccesary Network Drivers have been made , i still have to create the documentation for this so for now this is a WIP


KM198912:
Also switching from booleans to bitflags since it makes it nicer and easier to understand what is being disabled

MishaTY:
When is this going to get merged? This is a good feature.

quajak:
I thought we first want switch to use a bitflag to enable the drivers.

KM198912:
> I thought we first want switch to use a bitflag to enable the drivers.

that is the plan but i need to fix my computer first to be able to finish it, unless someone else could take that on, my new PSU arrives on wednesday, if nothing else is broken i can go fix it then

quajak (approved):

Issue #1984, opened 2021-11-04
80x25 Line Clearing In 90x60


Winksplorer:
#### Area of Cosmos - What area of Cosmos are we dealing with?
lines

#### Expected Behaviour - What do you think that should happen?
The Whole Line Should be cleared.

#### Actual Behaviour - What unexpectedly happens?
It Thinks The Screen Is 80x25 When It Isn't, So This Happens.
![image](https://user-images.githubusercontent.com/49623720/140244503-68c62a79-b1e5-4605-adfb-6a5e20b8d77e.png)

#### Reproduction - How did you get this error to appear?
Use Any Console Command That Uses Lines, for example, writeline and readline.
In A 90x60 Console.

#### Version - Were you using the User Kit or Dev Kit? And what User Kit version or Dev Kit commit (Cosmos, IL2CPU, X#)?
dev kit 106027 latest revision

quajak:
It looks like TextScreen.cs https://github.com/CosmosOS/Cosmos/blob/afc07521701e638241b7ecdf527e108dfc7b7171/source/Cosmos.HAL2/TextScreen.cs#L12 and https://github.com/CosmosOS/Cosmos/blob/afc07521701e638241b7ecdf527e108dfc7b7171/source/Cosmos.System2/Graphics/VGATextMode.cs#L10 dont communicate currently, which means that TextScreen never understands that the VGA changed resolution. Further TextScreen cant currently handle other sizes, although that should be relatively easily fixed.

I think the best way to fix this is to merge Core.IOGroup.TextScreen into the VGATextMode and then update the HAL2.TextScreen to also have a change of mode method.

@valentinbreiz @MishaTY @charlesbetros  Any experience with/thoughts on this?

PR #2029, opened 2021-11-30
Fix 90*60 Console functions not working

BlitzWolfMatthew:
The title explains explains everything.
Need a review so I know what to change, the code is a bit messy right now.

### DIFF
--- a/Tests/Cosmos.TestRunner.Full/TestKernelSets.cs
+++ b/Tests/Cosmos.TestRunner.Full/TestKernelSets.cs
@@ -15,6 +15,7 @@ namespace Cosmos.TestRunner.Full
         // Stable kernel types: the ones that are stable and will run in AppVeyor
         public static IEnumerable<Type> GetStableKernelTypes()
         {
+            yield return typeof(ConsoleTest.Kernel);
             yield return typeof(BoxingTests.Kernel);
             yield return typeof(Compiler.Tests.TypeSystem.Kernel);
             yield return typeof(Compiler.Tests.Bcl.Kernel);
@@ -32,9 +33,9 @@ namespace Cosmos.TestRunner.Full
 
             yield return typeof(GraphicTest.Kernel);
             yield return typeof(NetworkTest.Kernel);
-            /* Please see the notes on the kernel itself before enabling it */
+            // Please see the notes on the kernel itself before enabling it
             //yield return typeof(ConsoleTest.Kernel);
-            /* This is a bit slow and works only because ring check is disabled to decide if leave it enabled */
+            // This is a bit slow and works only because ring check is disabled to decide if leave it enabled
             yield return typeof(MemoryOperationsTest.Kernel);
             yield return typeof(ProcessorTests.Kernel);
         }
--- a/Tests/Cosmos.TestRunner.Full/TestKernelSets.cs
+++ b/Tests/Cosmos.TestRunner.Full/TestKernelSets.cs
@@ -15,7 +15,6 @@ namespace Cosmos.TestRunner.Full
         // Stable kernel types: the ones that are stable and will run in AppVeyor
         public static IEnumerable<Type> GetStableKernelTypes()
         {
-            yield return typeof(ConsoleTest.Kernel);
             yield return typeof(BoxingTests.Kernel);
             yield return typeof(Compiler.Tests.TypeSystem.Kernel);
             yield return typeof(Compiler.Tests.Bcl.Kernel);
--- a/Tests/Kernels/ConsoleTest/Kernel.cs
+++ b/Tests/Kernels/ConsoleTest/Kernel.cs
@@ -32,20 +32,9 @@ namespace ConsoleTest
         {
             try
             {
-                testConsoleEncoding();
+                TestConsoleEncoding();
 
-                Console.SetWindowSize(90, 60);
-                while (true)
-                {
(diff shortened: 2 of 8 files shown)


Tests/Cosmos.TestRunner.Full/TestKernelSets.cs
```diff
@@ -15,6 +15,7 @@ public static IEnumerable<Type> GetKernelTypesToRun()
         // Stable kernel types: the ones that are stable and will run in AppVeyor
         public static IEnumerable<Type> GetStableKernelTypes()
         {
+            yield return typeof(ConsoleTest.Kernel);
```
quajak, line 18: Console test should not be enabled for appveyor testrunner.

Tests/Kernels/ConsoleTest/Kernel.cs
```diff
@@ -32,100 +32,119 @@ protected override void Run()
         {
             try
             {
-                Console.Clear();
```
quajak, line 35: Please rework the Console tests. I think we should run a series of tests for each resolution, similar to how its done in Graphics Test.

BlitzWolfMatthew:
#1984

source/Cosmos.System2_Plugs/System/ConsoleImpl.cs
```diff
@@ -15,7 +15,7 @@ public static class ConsoleImpl
         private static Encoding ConsoleInputEncoding = Encoding.ASCII;
         private static Encoding ConsoleOutputEncoding = Encoding.ASCII;
 
-        private static readonly Cosmos.System.Console mFallbackConsole = new Cosmos.System.Console(null);
+        public static Cosmos.System.Console mFallbackConsole = new Cosmos.System.Console(null);
```
quajak, line 18: Why is this public?

source/Cosmos.System2_Plugs/System/ConsoleImpl.cs
```diff
@@ -570,7 +570,55 @@ public static void SetWindowPosition(int left, int top)
...
+            }
+            else
+            {
+                throw new Exception("Invalid text size.");
+            }
+            mFallbackConsole.Cols = mFallbackConsole.mText.Cols;
+            mFallbackConsole.Rows = mFallbackConsole.mText.Rows;
+
+            //VGA memory
+            ((Cosmos.HAL.TextScreen)mFallbackConsole.mText).IO.Memory = new Cosmos.Core.MemoryBlock(0xB8000, (uint)(width * height * 2));
```
quajak, line 617: I think its better if we add a method SetWindowSize to TextScreen and let TextScreen do the necessary changes internaly

quajak (changes requested):

Tests/Kernels/ConsoleTest/Kernel.cs
```diff
@@ -32,100 +32,136 @@ protected override void Run()
...
             catch (Exception e)
             {
                 mDebugger.Send("Exception occurred: " + e.Message);
                 mDebugger.Send(e.Message);
                 TestController.Failed();
-            }
+            }            
+        }
+
+        public void TestConsoleEncoding()
```
zarlo, line 49: the ```Console.ReadKey();``` in this method on line 84, 102 why are they there?

source/Cosmos.HAL2/TextScreen.cs
```diff
@@ -46,8 +46,16 @@ public TextScreen()
             TextScreenHelpers.Debug("End of TextScreen..ctor");
         }
 
-        public override ushort Rows { get { return 25; } }
-        public override ushort Cols { get { return 80; } }
+        public void UpdateWindowSize()
```
zarlo, line 49: check the values of Rows and Cols as anyone can set the values

zarlo (changes requested):

Tests/Kernels/ConsoleTest/Kernel.cs
BlitzWolfMatthew: This is how I found it, I didnt write that test. I only chnaged "germanic" to "german".

quajak: I dont think these tests are meant to be run automatic and the ReadKeys allow the developer to check if the output looks correct.

source/Cosmos.HAL2/TextScreen.cs
```diff
@@ -46,8 +46,16 @@ public TextScreen()
...
+        {
+            IO.Memory = new Cosmos.Core.MemoryBlock(0xB8000, (uint)(Cols * Rows * 2));
+            mRAM = new Cosmos.Core.MemoryBlock(0xB8000, (uint)(Cols * Rows * 2)).Bytes;
```
quajak, line 52: Nitpick: `mRAM  = IO.Memory.Bytes;`

quajak (changes requested):
Please add the checks @zarlo suggested and fix the nitpick. The rest looks good.

source/Cosmos.HAL2/TextScreen.cs
BlitzWolfMatthew: Rows and Cols are public but mFallBackConsole is private thus it cannot be accessed and changed by the user

quajak (approved):

PR #2042, opened 2021-12-23
Makefile

ilobilo:
Added Makefile
Haven't tested it tho, dotnet was too slow, but it should work


ilobilo:
making new pull request

PR #2050, opened 2022-01-04
Fix package conflict

valentinbreiz:
Needs https://github.com/CosmosOS/IL2CPU/pull/137

Since two csproj with the same code lead to this: 
image

This is a dirty hack to fix the builder issue many people are facing


charlesbetros:
Let's see if we can share the code between the projects

MishaTY (approved):

PR #2074, opened 2022-01-23
DoubleBuffering Implementation

valentinbreiz:
- [x] Implement double buffering for VMWare SVGAII
- [x] Restore VBE double buffering
- [x] Double buffered Canvas


source/Cosmos.HAL2/Drivers/Video/VBEDriver.cs
```diff
@@ -297,7 +301,7 @@ public void CopyVRAM(int aStart, int[] aData, int aIndex, int aCount)
         /// <param name="aCount">A count.</param>
         public void CopyVRAM(int aStart, byte[] aData, int aIndex, int aCount)
         {
-            lastbuffer.Copy(aStart, aData, aIndex, aCount);
```
zarlo, line 300: why copy it in to VRAM and not the buffer

MishaTY (dismissed):

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -94,23 +95,33 @@ public override Mode Mode
...
-            mSVGAIIDebugger.SendInternal($"Drawing point to x:{aX}, y:{aY} with {xColor.Name} Color");
-            _xSVGADriver.SetPixel((uint)aX, (uint)aY, (uint)xColor.ToArgb());
-            mSVGAIIDebugger.SendInternal($"Done drawing point");
-            /* No need to refresh all the screen to make the point appear on Screen! */
-            //xSVGAIIDriver.Update((uint)x, (uint)y, (uint)mode.Columns, (uint)mode.Rows);
-            _xSVGADriver.Update((uint)aX, (uint)aY, 1, 1);
+            _xSVGADriver.SetPixel((uint)aX, (uint)aY, (uint)aPen.ValueARGB);
+        }
+
+        public void DrawPointFast(uint aColor, uint aX, uint aY)
```
quajak, line 110: Why is this called `DrawPointFast` and not just `DrawPoint` with a different set of parameters?

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -155,7 +166,12 @@ public override void DrawPoint(Pen aPen, float aX, float aY)
         /// <exception cref="NotImplementedException">Thrown if VMWare SVGA 2 has no rectange copy capability</exception>
         public override void DrawFilledRectangle(Pen aPen, int aX_start, int aY_start, int aWidth, int aHeight)
         {
-            _xSVGADriver.Fill((uint)aX_start, (uint)aY_start, (uint)aWidth, (uint)aHeight, (uint)aPen.Color.ToArgb());
+            var color = aPen.Color.ToArgb();
+
+            for (int i = aY_start; i < aY_start + aHeight; i++)
+            {
+                _xSVGADriver.VideoMemory.Fill(GetPointOffset(aX_start, i) + (int)_xSVGADriver.FrameSize, aWidth, color);
```
quajak, line 173: Is this really quicker than just calling the `_xSVGADriver.Fill` which can use the VMWare driver call to fill the entire area with a single update? Might be better if you update the else case in Fill method with the VideoMemory method.

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -394,16 +421,93 @@ public override Color GetPointColor(int aX, int aY)
...
 
+        /// <summary>
+        /// Draw char.
+        /// </summary>
+        /// <param name="str">char to draw.</param>
+        /// <param name="aFont">Font used.</param>
+        /// <param name="pen">Color.</param>
+        /// <param name="x">X coordinate.</param>
+        /// <param name="y">Y coordinate.</param>
+        public override void DrawChar(char c, Font aFont, Pen pen, int x, int y)
```
quajak, line 481: Whats the difference between `DrawChar` and `DrawCharFast`?

quajak (changes requested):

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
valentinbreiz: Yes it is because the Fill method in the SVGAII driver can be slow due to some calls to WriteToFifo

quajak (approved):

terminal-cs:
Ik this is old but is accelerated cursor not available anymore with double buffered svgaii? It doesn't show anywhere for me. Both alpha and normal cursor defines seem broken

PR #2178, opened 2022-04-02
Added XS plug to make Math.Tan faster and safer

jwoff78:
Dependent on [XSharp PR](https://github.com/CosmosOS/XSharp/pull/58)

Fixed bug in assert also


quajak (changes requested):
Thanks for the PR. The new plug looks good. Can you please remove the old tan plug?

quajak (approved):

Issue #2166, opened 2022-03-20
Builder Pastebin integration


KM198912:
I think we should add pastebin or equivalent option to Builder so a failed build can be posted with a single Mouseclick, kind of like the "Copy log" button we have, the API seems pretty straight forward for that.

[PasteBin API Docs](https://pastebin.com/doc_api)

quajak:
Fixed by #2179

PR #2179, opened 2022-04-03
Added paste support to hastebin for the log

jwoff78:
Did the up for grabs for [Builder Pastebin integration](https://github.com/CosmosOS/Cosmos/issues/2166)
Had to use hastebin instead, pastebin requires dev key


quajak (approved):
Great work! Thanks for the PR!

MishaTY (approved):

PR #2243, opened 2022-05-23
Guid implementation 

MishaTY:
Reopens #942


MishaTY:
looks like I don't have write access there, so I will make a new branch

PR #2244, opened 2022-05-23
Guid Implementation

MishaTY:
Same as: #942


zarlo:
xD is this still not in it was started in 2017

zarlo:
and it looks like it still have te same issues with the guids been the same in the tests

quajak:
I think thats a problem with how we seed the random number generators currently. Making the random static should fix the issue.

zarlo:
mmm that a bit more of a hack then a fix

quajak:
Yeah, the correct fix would be to improve the seeding of the random number generator.

quajak:
@MishaTY The issue is that https://github.com/CosmosOS/Cosmos/blob/a1c4011cbfadcec5a8b61fac56d1d1a4de739462/source/Cosmos.System2_Plugs/System/RandomImpl.cs#L30-L33 does not use `GenerateGlobalSeed`.

quajak (approved):

PR #2246, opened 2022-05-25
Remove FTP server

valentinbreiz:
+ Make Tcp class and client StateMachine public for userland FTP implementation (coming soon as a nuget package)


IDE.sln
```diff
@@ -44,22 +44,18 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Spruce", "..\XSharp\source\
...
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IL2CPU.Debug.Symbols.Net48", "..\IL2CPU\source\IL2CPU.Debug.Symbols.Net48\IL2CPU.Debug.Symbols.Net48.csproj", "{AC3FEEDE-DA07-40BF-A1EA-56716EBC2818}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
 		Debug|x86 = Debug|x86
 		Release|Any CPU = Release|Any CPU
 		Release|x86 = Release|x86
+		TEST|Any CPU = TEST|Any CPU
```
KM198912, line 55: Is there a reason for a new configuration?

valentinbreiz: Visual Studio did it by itself

KM198912: I don't think we need it as only debug is supported right now anyway

valentinbreiz: What's the problem merging this?

valentinbreiz:
https://www.nuget.org/packages/CosmosFtpServer/

IDE.sln
KM198912: Will the extra Configuration end up on the user's vs? If yes I think it's better to remove it before merging

quajak: If think the Test configuration is needed for unit tests

quajak (approved):

PR #2258, opened 2022-05-29
Debug fixes

valentinbreiz:
- [x] Close VMWare when stop debugging
- [x] cdb already used problem

Needs https://github.com/CosmosOS/IL2CPU/pull/156

### DIFF
--- a/source/Cosmos.Debug.Hosts/VMware.cs
+++ b/source/Cosmos.Debug.Hosts/VMware.cs
@@ -88,12 +88,22 @@ namespace Cosmos.Debug.Hosts {
       }
       xPSI.UseShellExecute = false;  //must be true to allow elevate the process, sometimes needed if vmware only runs with admin rights
       mProcess.EnableRaisingEvents = true;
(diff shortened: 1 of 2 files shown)


MishaTY (approved):

source/Cosmos.Debug.Hosts/VMware.cs
```diff
@@ -88,22 +88,44 @@ public class VMware : Host {
...
+      {
+        try
+        {
+          //TODO: Close VMWare properly
+
+          //Force Kill VMWare
+          mProcess.Kill();
+
+          //kil vmware-vmx.exe
+          foreach (var process in Process.GetProcessesByName("vmware-vmx.exe"))
```
MishaTY, line 120: This should be changed in the future as it could close other vms that the user has open

valentinbreiz: Yes I thought about this, that's why I put a TODO

valentinbreiz: https://github.com/CosmosOS/Cosmos/issues/2259

PR #2277, opened 2022-06-02
[WIP] Add support for delegates with return values

Diamantino-Op:


quajak:
I dont think the correct code was pushed

Diamantino-Op:
Yeah, you're right, idk what happened, after i get home i'll push all the right code

PR #2278, opened 2022-06-03
 Add support delegates with support values

quajak:


quajak:
Should now be finished. Requires https://github.com/CosmosOS/IL2CPU/pull/160

MishaTY (approved):

PR #2288, opened 2022-06-10
Implement APIC

valentinbreiz:
- [ ] Fix PCI IRQs
-> Find a way to get IO APIC inputs 
ACPI AML? Too complicated. EDIT: Well..... Maybe (https://github.com/MishaTY/AMLParser) 
MP Tables? Deprecated.
MSI? Not supported on PCNETII..... but ~~should~~ will be implemented
- [x] Fix PIT IRQ
- [x] Use PIC if no APIC detected (Bochs fix)
-> not a bochs problem our memory map is broken so we overlap the ACPI tables and the APIC entry (btw idk why it works on vmware 😬)


AsertCreator:
does this acpi implementation require `ACPILib` namespace?

valentinbreiz:
Blocked by https://github.com/CosmosOS/Cosmos/issues/2546

Nik300:
@valentinbreiz how does the memory manager eat the APIC tables? Are you allocating them under the MM? If so, I'd argue that for something like that you'd actually need some free memory between the bootloader and cosmos to allocate tables of any sort (at least I always set up kernels like this).
Is someone working on the MM right now?

AsertCreator:
#2550

PR #2314, opened 2022-06-26
Improve DrawImageAlpha to not use a new Pen every time

GoldenretriverYT:
Currently, Canvas.DrawImageAlpha always creates a new Pen. For every pixel. This PR changes it so that there is one Pen, created at the beginning of the method. Each pixel just changes the color, which improves performance. My kernel with ~15 alpha images went from 15 fps to 24 fps


quajak (changes requested):
Nice optimisation! Could you please also add the trick to the normal DrawImage in Canvas.cs

GoldenretriverYT:
Done

quajak (approved):
Thanks!

Issue #164, opened 2015-07-24
Boxing is broken


charlesbetros:
There is a test kernel that shows the problem. This worked before the update to VS 2015.

charlesbetros:
The issue is due to some unimplemented plugs.

PR #2318, opened 2022-07-02
Cosmos Audio Infrastructure (CAI)

ascpixi:
This pull request implements an audio infrastructure that helps with audio-related operations and with interfacing to audio devices at a low-level. An AC97 driver is included.

The following features can be implemented in future pull requests:
- Surround channel support for the AC97
- 20-bit AC97 audio support by down-sampling 24-bit audio
- Sound Blaster 16 compatibility for VMWare

The AC97 driver is not supported by VMWare. It's recommended to use CAI with emulators/virtualizers such as QEMU or VirtualBox.

[You can download a demo kernel solution by clicking here.](https://github.com/CosmosOS/Cosmos/files/9033237/CAIDemo.zip)
> [This PR relies on IL2CPU PR #164](https://github.com/CosmosOS/IL2CPU/pull/164)


Kiirx:
This is a really interesting pr, but the problem is that in virtualbox the VFS doesnt work, and so we cant really use virtual box when we use the vfs, maybe ill try to give a preformatted vmdk file for virtualbox when I get home.

ascpixi:
> This is a really interesting pr, but the problem is that in virtualbox the VFS doesnt work, and so we cant really use virtual box when we use the vfs, maybe ill try to give a preformatted vmdk file for virtualbox when I get home.

There may be a way to get AC97 support in VMWare using the `sound.virtualDev` VMX property - I'll look into it

Kiirx:
> > This is a really interesting pr, but the problem is that in virtualbox the VFS doesnt work, and so we cant really use virtual box when we use the vfs, maybe ill try to give a preformatted vmdk file for virtualbox when I get home.
> 
> There may be a way to get AC97 support in VMWare using the `sound.virtualDev` VMX property - I'll look into it

This could maybe help you https://kb.vmware.com/s/article/2010400

source/Cosmos.HAL2/Audio/AudioBuffer.cs
```diff
@@ -0,0 +1,234 @@
...
+        /// </summary>
+        public void Flush()
+        {
+            int calculatedSize = format.size * Size;
+
+            if(buffer != null && buffer.Length == calculatedSize) {
+                // The buffer already exists and the size has not changed -
+                // set all of the bytes of the buffer to 0.
+                MemoryOperations.Fill(buffer, 0);
+            } else {
```
quajak, line 64: Can this case actually happen where the size is different? From the function name i would expect the size to already be correct.

source/Cosmos.HAL2/Audio/AudioBuffer.cs
```diff
@@ -0,0 +1,234 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Reads a sample from the audio buffer to the target pointer in memory.
+        /// </summary>
+        /// <param name="index">The index of the audio sample to read.</param>
+        /// <param name="dest">The part of memory to write the sample to.</param>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when attempting to write to a non-existent sample.</exception>
+        public unsafe void ReadSample(int index, byte* dest)
```
quajak, line 95: How is this different to MemoryOperation.Copy? If it's not please call it.

source/Cosmos.HAL2/Audio/AudioBuffer.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        /// <summary>
+        /// Reads a single channel of a sample to the target pointer in memory.
+        /// </summary>
+        /// <param name="index">The index of the audio sample to read the channel of.</param>
+        /// <param name="channel">The target channel to read.</param>
+        /// <param name="dest">The part of memory to write the sample to.</param>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when attempting to write to a non-existent channel or sample.</exception>
+        public unsafe void ReadSampleChannel(int index, int channel, byte* dest)
```
quajak, line 132: Same here.

source/Cosmos.HAL2/Audio/SampleFormat.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace Cosmos.HAL.Audio {
+    /// <summary>
+    /// Represents the format of an audio sample.
+    /// </summary>
+    public class SampleFormat {
+        /// <summary>
+        /// The audio bit depth, i.e. how many bits are used per a single
+        /// channel in a sample.
+        /// </summary>
+        public AudioBitDepth bitDepth;
```
quajak, line 12: Can you please capitalize the public fields?

source/Cosmos.HAL2/Cosmos.HAL2.csproj
```diff
@@ -5,10 +5,11 @@
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
         <RootNamespace>Cosmos.HAL</RootNamespace>
         <Configurations>Debug;Release;TEST</Configurations>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
quajak, line 8: Can you please clean up this csproj file? Unsafe code should be uniformly enabled.

source/Cosmos.HAL2/Drivers/PCI/Audio/AC97.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// <param name="mute">Whether to mute all output of the channel.</param>
+        private static ushort CreateMixerVolumeValue(byte right, byte left, bool mute)
+            => (ushort)((right & 0x3f) | ((left & 0x3f) << 8) | ((mute ? 1 : 0) & 1 << 15));
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="AC97"/> class, with the
+        /// given buffer size.
+        /// </summary>
+        /// <param name="bufferSize">The buffer size in samples to use. This value cannot be an odd number, as per the AC97 specification.</param>
+        /// <exception cref="ArgumentException">Thrown when the given buffer size is incorrect.</exception>
```
quajak, line 96: Incorrect=> invalid

source/Cosmos.HAL2/Drivers/PCI/Audio/AC97.cs
```diff
@@ -0,0 +1,301 @@
...
+
+            // Reset device
+            pGlobalControl.Byte = 0x2;
+            pResetRegister.DWord = 0xDEADBEEF; // any value will do here
+
+            // Reset PCM out
+            pTransferControl.Byte = (byte)(pTransferControl.Byte | TC_TRANSFER_RESET);
+            while ((pTransferControl.Byte & TC_TRANSFER_RESET) != 0)
+            {
+                // Wait until the byte is cleared
```
quajak, line 141: Will this always occur? Otherwise can you add some timeout.

source/Cosmos.HAL2/Drivers/PCI/Audio/AC97.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// Initializes the AC97 driver. This method will return
+        /// an existing instance if the driver is already initialized
+        /// and has a running instance.
+        /// </summary>
+        /// <param name="bufferSize">The buffer size in samples to use. This value cannot be an odd number, as per the AC97 specification.</param>
+        /// <exception cref="ArgumentException">Thrown when the given buffer size is incorrect.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when no AC97-compatible sound card is present.</exception>
+        public static AC97 Initialize(ushort bufferSize)
+        {
+            if (Instance != null)
```
quajak, line 170: Ensure that the current instance has the correct buffer size.

source/Cosmos.HAL2/Drivers/PCI/Audio/AC97.cs
```diff
@@ -0,0 +1,301 @@
...
+            }
+
+            // Tell BDL location
+            fixed (void* ptr = bufferDescriptorList)
+            {
+                pBufferDescriptors.DWord = (uint)ptr;
+            }
+
+            // Set last valid index
+            lastValidIdx = 2;
```
quajak, line 206: I assume the 2 is due to only two audio channels being supported? For clarity, could you please define an enum or const, which makes the meaning of the 2 clear.

source/Cosmos.HAL2/Drivers/PCI/Audio/AC97.cs
```diff
@@ -0,0 +1,301 @@
...
+            if (!sampleFormat.signed)
+                throw new NotSupportedException("The AC97 driver does not support unsigned audio.");
+
+            // TODO: The AC97 specification defines support 2/4/6 channel audio. Currently, only stereo audio output is supported.
+        }
+
+        public override bool Enabled =>
+            (pTransferControl.Byte & TC_RUN_OR_PAUSE) != 0;
+
+        public override void Enable()
```
quajak, line 275: I assume enabling an already enabled driver would cause it to reset?

source/Cosmos.System2/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AudioMixer"/> class,
+        /// with the specified initial internal mixing buffer size and sample format.
+        /// </summary>
+        /// <remarks>
+        /// This constructor should be used when the sizes and formats of the
+        /// buffers that will be given to the mixer are known. If there is a
+        /// mis-match between the mixing format and/or the buffer size and the given buffer,
+        /// the internal mixing buffer will be re-initialized with the right values.
```
quajak, line 27: Why not throw an exception in this case?

source/Cosmos.System2/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,177 @@
...
+        public override bool Depleted => false;
+
+        /// <summary>
+        /// The position, in samples, of the mixer audio stream. Attempting to set this
+        /// property to another value will result in an <see cref="InvalidOperationException"/>.
+        /// </summary>
+        public uint Position
+        {
+            get => pos;
+            set => throw new InvalidOperationException("Cannot set the position of an AudioMixer.");
```
quajak, line 77: Why not just remove the setter?

source/Cosmos.System2/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            pos += (uint)buffer.Size;
+
+            ApplyPostProcessing(buffer);
+        }
+
+        private static short SaturationAdd(short a, short b)
+        {
+            if(a > 0) {
+                if (b > short.MaxValue - a)
```
quajak, line 168: Please always include {}

ascpixi:
It looks like that while the ES1371 is a AC97-compatible sound card, [it still defines its own method to set it up.](https://wiki.osdev.org/Ensoniq_AudioPCI)... This explains why the AC97 driver does detect the sound card, but it doesn't want to reset. I feel like this will require another driver class :/

source/Cosmos.System2/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,177 @@
...
+                    return short.MaxValue;
+            }
+            else if (b < short.MinValue - a)
```
quajak, line 171: Does this not cause an underflow?

source/Cosmos.System2/Audio/Device/AudioManager.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <summary>
+        /// Disables this audio manager.
+        /// </summary>
+        public void Disable()
+        {
+            Output?.Disable();
+            enabled = false;
+        }
+
+        public void RequestBuffer(AudioBuffer buffer)
```
quajak, line 79: Please add method description

quajak (changes requested):
Thanks for the great PR! Overall the code looks very good with just a few tiny changes required. I already added a few questions/comments but will need to look through it again. 

Could you please add some documentation for both users to use it and other developers to better understand the design approach? I think it would also be good to add a few tests to ensure that the code does not break.

source/Cosmos.Core/PIC.cs
```diff
@@ -71,7 +71,7 @@ public PIC()
             //Init(Master, 0x20, 4, 0xFD | 0x08);
             //Init(Slave, 0x28, 2, 0xFF);
             //for now enable keyboard, mouse(ps2)
-            Remap(0x20, 0xF8 | 0x08, 0x28, 0xEB);
+            Remap(0x20, 0xF8 | 0x08, 0x28, 0xE3);
```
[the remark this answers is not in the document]
ascpixi: This is the PIC slave mask, and we're setting the 4th bit to unlock IRQ 11, which is used by most AC97-compatible cards (from testing, both QEMU and VBox use this IRQ)

source/Cosmos.HAL2/Audio/AudioBuffer.cs
ascpixi: Oops, yeah, sorry! This is a remnant of an old version of the API I was working on where the buffer was actually created within this method. I'll create a commit that removes this case ASAP!

source/Cosmos.HAL2/Audio/SampleFormat.cs
ascpixi: Wouldn't that be a violation of C#'s naming convention, though? IIRC, fields should always be `camelCase`, except when constant (`static readonly`), which would be `PascalCase`

source/Cosmos.HAL2/Drivers/PCI/Audio/AC97.cs
ascpixi: Yep, this won't occur on cards such as the Ensoniq AudioPCI. I'm working on a driver for it and the timeout is added in my local version, the driver should be out soon :3

source/Cosmos.HAL2/Drivers/PCI/Audio/AC97.cs
ascpixi: Honestly, the driver will work on any value that's in the range of <0; 32>... it's more of a remnant when I was testing the driver ^^'>
I'll change this to 0 in a future commit to this PR.

source/Cosmos.HAL2/Drivers/PCI/Audio/AC97.cs
ascpixi: Oops, yeah, a call to `Enable()` should be ignored if the driver is already enabled. Sorry, missed that one, will change that!

source/Cosmos.System2/Audio/AudioMixer.cs
ascpixi: Cannot remove the setter in an abstract class, unfortunately :(

source/Cosmos.System2/Audio/AudioMixer.cs
ascpixi: Honestly, this whole constructor can be deleted, it was made before the `AudioMixer` class supported writing to buffers of different sizes :P

source/Cosmos.System2/Audio/AudioMixer.cs
ascpixi: Yes, the method relies on detecting underflows/overflows in order to perform saturation arithmetic - you can test the code on a website like rextester.com

source/Cosmos.System2/Audio/Device/AudioManager.cs
ascpixi: Documentation is inherited from `IAudioBufferProvider`

ascpixi: Documentation is inherited from class `IAudioBufferProvider`... but to be fair, I should've put an `override` here so it's more clear, oops :P

source/Cosmos.System2/Audio/AudioMixer.cs
jp2masa: In this else, `a <= 0`, so it doesn't overflow.

ascpixi: We already check for this in the `if(a > 0)` statement

Docs/articles/Kernel/Audio.md
```diff
@@ -0,0 +1,81 @@
+# The Cosmos Audio Infrastructure (CAI)
```
quajak, line 1: Great article! It makes it very clear of how to use CAI.
Can you add it to https://github.com/CosmosOS/Cosmos/blob/master/Docs/articles/toc.md so that it can be found?

source/Cosmos.Core/PIC.cs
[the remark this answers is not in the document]
quajak: Thanks for the explanation.

source/Cosmos.HAL2/Audio/SampleFormat.cs
quajak: From https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions "When naming public members of types, such as fields, properties, events, methods, and local functions, use pascal casing." 

It is more important that its `public` than that it is a field.

source/Cosmos.System2/Audio/AudioMixer.cs
quajak: I thought `AudioMixer` only inherits from `AudioStream` which doesnt have `Position`. Only `SeekableAudioStream` has a `Position`.

source/Cosmos.System2/Audio/AudioMixer.cs
quajak: Thanks, missed the effect of the `a > 0` check.

source/Cosmos.System2/Audio/Device/AudioManager.cs
quajak: If the documentation is inherited, thats fine.

source/Cosmos.HAL2/Audio/AudioBuffer.cs
```diff
@@ -97,12 +88,15 @@ public unsafe void ReadSample(int index, byte* dest)
...
 
+
             int bufferOffset = index * format.size;
 
             fixed(byte* bufferPtr = buffer) {
-                for (int i = 0; i < format.size; i++) {
-                    *(dest + i) = *(bufferPtr + i);
-                }
+                MemoryOperations.Copy(
+                    dest,
```
quajak, line 96: Could you please write the call in a single line? Style guide is usually to have the call in a single line until it gets very long

source/Cosmos.HAL2/Cosmos.HAL2.csproj
```diff
@@ -9,15 +9,15 @@
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-      <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+      <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
quajak, line 12: I think we should be fine to just have a single PropertyGroup for all options, since the options shouldnt change. Please remove all but the first PropertyGroup

quajak (changes requested):

MEMESCOEP:
How do I use the example? I got 13 errors when I first opened the project :(

ascpixi:
> How do I use the example? I got 13 errors when I first opened the project :(
Right now, the CAI isn't merged into Cosmos - if you're the devkit built using [this branch](https://github.com/ascpixi/Cosmos/tree/cosmos-audio-infrastructure) and having errors, please post the errors you get here

Arawn-Davies:
SB16 in Cosmos, next up, DOOM!
👍🏽 looks exciting!

source/Cosmos.System2/Audio/AudioMixer.cs
ascpixi: Oops, yeah, you're right! I was thinking of an old version where all AudioStreams had that property, sorry ^^'>

quajak (approved):
Thanks for changing it. Looks good to me now!

valentinbreiz (approved):
Really great work thanks for your contribution, it looks fine to me. Plus since the AC97 card supports MSI it will be really helpful to test interrupts for PCI devices using APIC.

Edit: @quajak is it okay for you to have unsafe code under Cosmos.HAL2?

ascpixi:
Some minor adjustments, I feel like it's ready to be merged now!

valentinbreiz (approved):

quajak:
@valentinbreiz Good point, I think its fine with unsafe code in HAL. IMO The code definitely belongs in HAL. Not using pointers is slower and I think speed is quite important for CAI. Rewriting it all to use Spans would maybe work 🤷 but simply using pointers if fine for me.

quajak (approved):

PR #2322, opened 2022-07-07
fix building

terminal-cs:


Kiirx:
I tought I fixed this lol, ill try again since I just got on Debian 11

source/Cosmos.Debug.Hosts/Qemu.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System;
+﻿/*
+using System;
```
quajak, line 2: Can you find a fix which doesnt require commenting out all of Qemu? Or is the file no longer needed?

quajak (changes requested):

source/Cosmos.Debug.Hosts/Qemu.cs
zarlo: its used in the current master

terminal-cs: We can just un-comment the whole file when it comes time to merge

PR #2358, opened 2022-07-28
tags in VS templates

AsertCreator:
i just noticed there are no tags in VS templates. it looks bad. so i made this


quajak (approved):
Thanks!

PR #2378, opened 2022-08-08
Add kernel compression

JsphA:
This PR adds an option to compress the kernel binary using gzip, reducing the ISO size (by around 80%, on average.)

To use:
- Open a Cosmos project's properties.
- Set Compression to gzip.

Requires https://github.com/CosmosOS/Common/pull/9


AsertCreator:
i like it! almost all cosmos kernels are bigger than 30 mb

quajak (approved):
Great idea! Thanks for the PR

PR #2390, opened 2022-08-21
Fix Canvas "Wrap-Around" issue

MEMESCOEP:
This pull request fixes some of the canvas "Wrap-Around" issue (Where things drawn out of screen bounds get duplicated onto the opposing side of the screen). This currently only works with the VMWareSVGAII Canvas, and only fixes Filled Rectangles, Text, and Bitmaps

### DIFF
--- a/source/Cosmos.System2/Graphics/Bitmap.cs
+++ b/source/Cosmos.System2/Graphics/Bitmap.cs
@@ -518,5 +518,32 @@ namespace Cosmos.System.Graphics
             Array.Copy(imageData, 0, file, position, imageData.Length);
             stream.Write(file, 0, file.Length);
         }
+
+        /// <summary>
+        /// Resize a bitmap
+        /// </summary>
+        /// <param name="Width">New Bitmap Width.</param>
+        /// <param name="Height">New Bitmap Height.</param>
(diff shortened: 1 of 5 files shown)


MEMESCOEP:
I've just realized that I broke circle drawing in this pr, so I will fix it asap

GoldenretriverYT:
I think preventing wrap around should be handled by the OS, not Cosmos.
It decreases performance (on the already slow canvas) even if people don't want this feature.

MEMESCOEP:
> I think preventing wrap around should be handled by the OS, not Cosmos.
It decreases performance (on the already slow canvas) even if people don't want this feature.

From my testing it doesn't decrease performance? Would you tell me why it decreases performance, because the SVGAII Canvas doesn't seem all that slow to me

GoldenretriverYT:
Of course new, useless checks that should be up to the OS will cause performance decreases.
More code = more cycles

MEMESCOEP:
ah, I see. Thanks for explaining! Maybe I can make it faster!

Also, from what I've heard, these checks aren't useless because people want them. I think Loxy wanted it, and someone else wanted it but I can't remember their name

Issue #2391, opened 2022-08-21
Alpha blending doesn't work correctly if not displaying blank screen before


GoldenretriverYT:
#### Area of Cosmos - What area of Cosmos are we dealing with?
Graphics

#### Expected Behaviour - What do you think that should happen?
The alpha blending should blend with the color of the current image that hasn't been displayed yet.

#### Actual Behaviour - What unexpectedly happens?
Alpha blending blends with the color of the displayed image

#### Reproduction - How did you get this error to appear?
Setup canvas, draw image with alpha. Once without displaying directly after Canvas.Clear and once with displaying directly after Canvas.Clear

The one without call Canvas.Display() will cause the alpha image to look bolder every frame. Displaying the black screen inbetween will prevent it.

#### Version - Were you using the User Kit or Dev Kit? And what User Kit version or Dev Kit commit (Cosmos, IL2CPU, X#)?
DevKit https://github.com/CosmosOS/Cosmos/commit/545d9607fa0e6636fb9d20718d917f4bc0c70e35 (not up-to-date, but there aren't any changes to alpha drawing in the newer commits)

_Tested on VMWare_

#### Examples
Example with displaying black screen:
```
public void RenderFrame() {
            cv.Clear(Color.Black);
            cv.Display();
            HandleInputs();

            foreach (Form form in registeredForms) {
                form.Render(cv);
                Cosmos.Core.Memory.Heap.Collect();
            }

            cv.DrawImageAlpha(ResourceManager.CursorImage, (int)MouseManager.X, (int)MouseManager.Y);
            cv.Display();
        }
```
![image](https://user-images.githubusercontent.com/49924528/185789538-abf76562-55f2-4d39-af83-af1626a5b227.png)

Example without:
```
public void RenderFrame() {
            cv.Clear(Color.Black);
            HandleInputs();

            foreach (Form form in registeredForms) {
                form.Render(cv);
                Cosmos.Core.Memory.Heap.Collect();
            }

            cv.DrawImageAlpha(ResourceManager.CursorImage, (int)MouseManager.X, (int)MouseManager.Y);
            cv.Display();
        }
```
![image](https://user-images.githubusercontent.com/49924528/185789610-5a674bf0-1376-46a7-8da9-382e06d51b23.png)

JsphA:
This is due to GetPointColor using the screen buffer instead of the back buffer. That means this also applies to other functions that use it, like DrawPoint.

PR #2396, opened 2022-08-22
Fix SVGAII GetPixel

GoldenretriverYT:
This PR fixes following issue: https://github.com/CosmosOS/Cosmos/issues/2391

Every other canvas seems to always use the correct offset.


GoombaProgrammer:
"good fix"
"very needed"

AsertCreator (approved):
well im not reading SVGAII canvas data, but it sounds cool. one line fix, thats cool

quajak (approved):
Good find. Thanks for the fix!

PR #2400, opened 2022-08-22
Fix VMWare SVGAIICanvas "Wrap-Around" Issue

MEMESCOEP:
This pull request fixes VMWareSVGAII Canvas "Wrap-Around" issue (Where things drawn out of screen bounds get duplicated onto the opposing side of the screen). This currently only works with the VMWareSVGAII Canvas

### DIFF
--- a/source/Cosmos.System2/Graphics/Bitmap.cs
+++ b/source/Cosmos.System2/Graphics/Bitmap.cs
@@ -518,5 +518,32 @@ namespace Cosmos.System.Graphics
             Array.Copy(imageData, 0, file, position, imageData.Length);
             stream.Write(file, 0, file.Length);
         }
+
+        /// <summary>
+        /// Resize a bitmap
+        /// </summary>
+        /// <param name="Width">New Bitmap Width.</param>
+        /// <param name="Height">New Bitmap Height.</param>
+        /// <param name="Image">The bitmap to resize.</param>
+        /// <returns>The resized bitmap.</returns>
+        public static Cosmos.System.Graphics.Bitmap ResizeBitmap(uint Width, uint Height, Cosmos.System.Graphics.Bitmap Image)
+        {
+            if (Width <= 0 || Height <= 0 || Width == Image.Width || Height == Image.Height)
+            {
+                return Image;
+            }
+
+            Cosmos.System.Graphics.Bitmap B = new(Width, Height, Cosmos.System.Graphics.ColorDepth.ColorDepth32);
+            for (int IX = 0; IX < Image.Width; IX++)
+            {
+                for (int IY = 0; IY < Image.Height; IY++)
+                {
+                    long X = IX / (Image.Width / Width);
+                    long Y = IY / (Image.Height / Height);
+                    B.rawData[(B.Width * Y) + X] = Image.rawData[(Image.Width * IY) + IX];
+                }
+            }
+            return B;
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


GoldenretriverYT:
As I already said before, I think this should be handled by the host OS and not Cosmos. But lets see what the maintainers think.

MEMESCOEP:
yeah ik, but I just fixed some stuff

MEMESCOEP:
@GoldenretriverYT why should this be handled by the os?

quajak (commented):
I am not convinced that we should be "fixing" user input in this case since it makes bugs harder to find and comes with performance costs. If we are fine with the performance costs, imo we should just raise an exception.  

Do you want the Bitmap changes and Crop function to be part of this PR?

MEMESCOEP:
@quajak yeah the bitmap stuff is fine for this pr

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -138,6 +143,21 @@ public override void DrawPoint(Pen aPen, float aX, float aY)
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Crop the width of Shapes / Images / Etc
+        /// </summary>
+        public byte[] Crop(byte[] data, uint width, uint height, uint cropWidth, uint pixelSize)
```
quajak, line 149: Why is this method in SVGAIICanvas? If we want to have such a method, it should be part of Bitmap/Image imo.

source/Cosmos.System2/Graphics/Bitmap.cs
```diff
@@ -518,5 +518,32 @@ public void Save(Stream stream, ImageFormat imageFormat)
...
         }
+
+        /// <summary>
+        /// Resize a bitmap
+        /// </summary>
+        /// <param name="Width">New Bitmap Width.</param>
+        /// <param name="Height">New Bitmap Height.</param>
+        /// <param name="Image">The bitmap to resize.</param>
+        /// <returns>The resized bitmap.</returns>
+        public static Cosmos.System.Graphics.Bitmap ResizeBitmap(uint Width, uint Height, Cosmos.System.Graphics.Bitmap Image)
```
quajak, line 529: We already have this method https://github.com/CosmosOS/Cosmos/blob/master/source/Cosmos.System2/Graphics/Canvas.cs#L894
I think its a valid use case to also have a public interface in the Bitmap class itself but we should not be implementing the same thing twice

quajak (changes requested):
I added my thoughts about the Bitmap changes. I am still not convinced about the proposed changes to the position checks.

AsertCreator:
why not add a check to the drawpoint that won't draw the point if the point is out of bounds

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
MEMESCOEP: good question, and idk lol. it can be moved

source/Cosmos.System2/Graphics/Bitmap.cs
MEMESCOEP: alr, I'll remove it

MEMESCOEP:
> 

I believe that's what I did, although I don't fully remember

AsertCreator:
well its not

MEMESCOEP:
Alr lemme fix it

MEMESCOEP:
I'm closing this pr until i get the motivation to fix this again lol

PR #2406, opened 2022-09-06
fix directory not found when using .net 6

Ninjastorm380:
found a bug in your makefile in regards to .net 6. the output folder was changed to net6.0, so I updated your makefile to use any folder name, as well as clean beforehand so that there will be only one output folder.


Kiirx:
duplicatas of #2322

quajak (approved):
Thanks!

PR #2417, opened 2022-09-24
Minor PIT improvements

ascpixi:
This PR adds minor improvements to the PIT:
- minor formatting changes
- added documentation
- ability to read the CPU state when the timer interrupt has occurred, allowing for things like basic preemptive multitasking
- timeouts changed from signed 32-bit integers to unsigned 64-bit integers, allowing for higher delays between each timer cycle

### DIFF
--- a/source/Cosmos.HAL2/PIT.cs
+++ b/source/Cosmos.HAL2/PIT.cs
@@ -282,10 +282,10 @@ namespace Cosmos.HAL
                         handler.ID = -1;
                         activeHandlers.RemoveAt(i);
                     }
-
+                    
                     handler.HandleTrigger(aContext);
                 } else {
-					handler.NSRemaining -= T0Delay;
+                    handler.NSRemaining -= T0Delay;
 				}
             }
 
--- a/source/Cosmos.HAL2/PIT.cs
+++ b/source/Cosmos.HAL2/PIT.cs
@@ -217,6 +217,7 @@ namespace Cosmos.HAL
             T2Frequency = (uint)aFreq;
         }
 
+        [Obsolete("This method has been deprecated and is equivalent to a no-op.")]
         public void MuteSound()
         {
             DisableSound();
--- a/source/Cosmos.HAL2/PIT.cs
+++ b/source/Cosmos.HAL2/PIT.cs
@@ -4,18 +4,18 @@ using Cosmos.Core;
 
 namespace Cosmos.HAL
 {
-	/// <summary>
-	/// Handles the Programmable Interval Timer (PIT).
-	/// </summary>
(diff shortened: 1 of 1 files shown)


terminal-cs:
Could you change all nanosecond fields to ulong instead of int? Int limits the max time to ~2.3 seconds and we don't need values less than 0

AsertCreator:
will pc speaker work?

ascpixi:
> will pc speaker work?

This PR only affects channel 0 of the PIT, so the PC speaker shouldn't be affected, as it runs on channel 2 (I haven't tested it though...)

source/Cosmos.HAL2/PIT.cs
```diff
@@ -1,280 +1,341 @@
...
+            {
+                if (value > 54918330) {
+                    throw new ArgumentException("Delay must be no greater than 54918330");
+                }
+
+                T2Countdown = (ushort)(value / PITDelayNS);
+            }
+        }
+
+        public void EnableSound()
```
quajak, line 202: Can we also deprecate all the sound methods?

source/Cosmos.HAL2/PIT.cs
```diff
@@ -1,280 +1,341 @@
...
+                    }
+                    else
+                    {
+                        handler.ID = -1;
+                        activeHandlers.RemoveAt(i);
+                    }
+
+                    handler.HandleTrigger(aContext);
+                } else {
+					handler.NSRemaining -= T0Delay;
```
quajak, line 288: Formatting looks off.

quajak (changes requested):
Thanks for the great contribution! Just some minor points

source/Cosmos.HAL2/PIT.cs
ascpixi: Sure, will mark with `[Obsolete]` as to not break any old kernels

source/Cosmos.HAL2/PIT.cs
ascpixi: woops

AsertCreator:
wait, i noticed `PIT.PITTimer.OnTrigger` does not use reference parameter. therefore i cant, for example, implement premptive multitasking. i can only read registers (maybe)

ascpixi:
> wait, i noticed `PIT.PITTimer.OnTrigger` does not use reference parameter. therefore i cant, for example, implement premptive multitasking. i can only read registers (maybe)

Writing data to the structure won't affect registers... only actually writing to the registers using assembly will. The IRQContext structure is intended to be only read from, CMIIW

quajak (approved):
Thanks!

PR #2421, opened 2022-09-29
Add arc drawing to canvas

### LINKED ISSUE #1639: Add method to draw arc
Thanks to the discussion https://github.com/CosmosOS/Cosmos/discussions/1638, I noticed that a method to draw an arc between two points is currently missing from canvas and should be added(the algorithm should be very similar to how we draw circles or ellipses).

terminal-cs:
[+] Added arc drawing (Resolves #1639)

### DIFF
--- a/Tests/Kernels/GraphicTest/Kernel.cs
+++ b/Tests/Kernels/GraphicTest/Kernel.cs
@@ -79,6 +79,9 @@ namespace GraphicTest
             pen.Color = Color.Chartreuse;
             aCanvas.DrawCircle(pen, 69, 69, 10);
 
+            pen.Color = Color.CadetBlue;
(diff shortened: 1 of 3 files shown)


quajak (changes requested):
Looks good. Before I merge, can you please add a test usage to the CGS test kernel?

terminal-cs:
> Looks good. Before I merge, can you please add a test usage to the CGS test kernel?

May i ask where that is?

MishaTY:
> > Looks good. Before I merge, can you please add a test usage to the CGS test kernel?
> 
> May i ask where that is?

https://github.com/CosmosOS/Cosmos/blob/master/Tests/Kernels/GraphicTest/Kernel.cs

quajak (approved):

PR #2423, opened 2022-10-01
Combine QEMU power functions with standard functions

RaphMar2021:
Combine the QEMU power functions ( QemuShutdown() and QemuReboot() ) with standard power functions ( Shutdown(), Reboot() )

### DIFF
--- a/source/Cosmos.System2/Power.cs
+++ b/source/Cosmos.System2/Power.cs
@@ -13,6 +13,10 @@ namespace Cosmos.System
         /// </summary>
         public static void Reboot()
         {
+            // Try QEMU method
+            new IOPort(0x64).Byte = 0xFE;
+            
+            // Try normal method
             HAL.Power.CPUReboot();
         }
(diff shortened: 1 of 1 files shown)


terminal-cs:
Not a great idea to brute force it because this could invoke some unwanted function in a device on the system

RaphMar2021:
wdym?

quajak:
is there anyway to detect if we are running in qemu?

zarlo:
from memory there are CPU flags we can check (this is still not 100%)

zarlo:
yep the CPU vendor id should be "TCGTCGTCGTCG"

quajak:
Then we should either let the user do the check, or do it in the method.

terminal-cs:
the pr was merged and this no longer is needed, you can delete it

RaphMar2021:
bruhhhhhhhhhhhhhhhhhhhh

PR #2426, opened 2022-10-04
Driver Optimize And added Fast Boost Option

TheFocusMan:
Improved the "Partitions" in Disk,
Now Able to get Max LBA in ATAPI,
Added the option "Fast boost"

### DIFF
--- a/source/Cosmos.Core/CPU.cs
+++ b/source/Cosmos.Core/CPU.cs
@@ -38,6 +38,16 @@ namespace Cosmos.Core
         public static uint GetEBPValue() => throw null;
 
         /// <summary>
+        /// Get position of current ESP register
+        /// </summary>
+        /// <returns></returns>
+        [PlugMethod(PlugRequired = true)]
+        public static uint GetESPValue() => throw null;
+
+        [PlugMethod(PlugRequired = true)]
+        public static void SetESPValue(uint val) => throw null;
+
+        /// <summary>
         /// Get the address at which the stack starts
         /// </summary>
         /// <returns></returns>
--- a/source/Cosmos.System2/FileSystem/Disk.cs
+++ b/source/Cosmos.System2/FileSystem/Disk.cs
@@ -11,6 +11,7 @@ namespace Cosmos.System.FileSystem
     public class Disk
     {
         private List<ManagedPartition> parts = new List<ManagedPartition>();
+        private static List<FileSystemFactory> registeredFileSystems = new List<FileSystemFactory>() { new FatFileSystemFactory(), new ISO9660.ISO9660FileSystemFactory()};
         public bool IsMBR { get { return !GPT.IsGPTPartition(Host); } }
         /// <summary>
         /// The size of the disk in bytes.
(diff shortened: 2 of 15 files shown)


AsertCreator:
what the fast boost does? why drives are not working now? how driver is optimized. why you added `registeredFileSystems` if they're been moved to separate class? what is LBA?

source/Cosmos.System2/FileSystem/Disk.cs
```diff
@@ -23,50 +24,49 @@ public List<ManagedPartition> Partitions
...
+                //}
 
-                return converted;
+                return parts;
             }
         }
         /// <summary>
         /// List of file systems.
         /// </summary>
-        [Obsolete("use FileSystemManager.RegisteredFileSystems")]
```
zarlo, line 68: this change here will been to be reverted

source/Cosmos.Core/CPU.cs
```diff
@@ -37,6 +37,16 @@ public static uint GetAmountOfRAM()
         [PlugMethod(PlugRequired = true)]
         public static uint GetEBPValue() => throw null;
 
+        /// <summary>
+        /// Get position of current ESP register
+        /// </summary>
+        /// <returns></returns>
+        [PlugMethod(PlugRequired = true)]
```
zarlo, line 44: i feel this should not be exposed to c#

source/Cosmos.System2/Kernel.cs
```diff
@@ -3,14 +3,15 @@
 using Cosmos.Debug.Kernel;
 using Cosmos.HAL;
 using Cosmos.System.ScanMaps;
+using Cosmos.Core;
```
zarlo, line 6: you can not access the core ring in the system ring this will break when the ring master comes out

zarlo (changes requested):

source/Cosmos.Core/CPU.cs
TheFocusMan: yes u right it was a bug test

source/Cosmos.System2/FileSystem/Disk.cs
TheFocusMan: the "registeredFileSystems" in the disk.cs was not updated

source/Cosmos.System2/Kernel.cs
TheFocusMan: You can remove the "AdvancedObject" class in the core ring

source/Cosmos.System2/FileSystem/Disk.cs
TheFocusMan: The Change was reveted

source/Cosmos.Core/CPU.cs
TheFocusMan: The Change is set the methods from public to internel

source/Cosmos.System2/Kernel.cs
TheFocusMan: Change is reveted

source/Cosmos.System2/FileSystem/Disk.cs
```diff
@@ -23,43 +23,43 @@ public List<ManagedPartition> Partitions
         {
             get
             {
-                var converted = new List<ManagedPartition>();
```
MishaTY, line 26: The point of this property is to get all of the latest partitions on the disk. If a partition gets deleted, this should also reflect that.

MishaTY (changes requested):

TheFocusMan:
> what does 'FastBoost' even do? also to me that doesn't sound like proper english.

The fast boost is just remove by choice the "WriteDebugVideo" in the compiler, its helped me to solve heap problems

PR #2430, opened 2022-10-06
Minor heap changes, Realloc function

terminal-cs:


zarlo:
have you been reading my mind?

well at lest i can remove this from what im working on

source/Cosmos.Core/Memory/Heap.cs
```diff
@@ -32,6 +32,40 @@ public static unsafe void Init()
             HeapLarge.Init();
         }
 
+        /// <summary>
+		/// Re-allocates or "re-sizes" data asigned to a pointer.
```
quajak, line 36: Maybe add that this must be a pointer to the start of the allocated memory region

source/Cosmos.Core/Memory/Heap.cs
```diff
@@ -32,6 +32,40 @@ public static unsafe void Init()
...
+		/// </summary>
+		/// <param name="aPtr">Existing pointer</param>
+		/// <param name="NewSize">Size to extend to</param>
+		/// <returns>New pointer with specified size while maintaining old data.</returns>
+        public static byte* Realloc(byte* aPtr, uint NewSize)
```
quajak, line 41: Typo: newSize not NewSize

terminal-cs:
> have you been reading my mind?
> 
> well at lest i can remove this from what im working on

I guess so lol

source/Cosmos.Core/Memory/Heap.cs
```diff
@@ -32,6 +32,40 @@ public static unsafe void Init()
...
+                Size -= (NewSize - Size);
+			}
+
+            // Allocate a new buffer to use
+			byte* ToReturn = Alloc(NewSize);
+
+            // Copy the old buffer to the new one
+            MemoryOperations.Copy(ToReturn, aPtr, (int)Size);
+            // Copy the GC state
+            ((ushort*)ToReturn)[-1] = ((ushort*)aPtr)[-1];
```
quajak, line 62: I have thought about this more and it seems like we should just set the GC state to 0 for the new object, since the old references no longer count.

quajak (changes requested):
Good idea! Could you please add a test somewhere to ensure that this does what its supposed to do and does not mess with the garbage collector? 

Regarding the garbage collecting, this change does not really work well with C# objects since all reference to the original object would be broken. This means that either we dont allow this for objects with references pointing to it or its the users fault.  Since we cant accurately detect if something is being referenced (if the reference is not static), I would recommend we go with option 2 but include a large warning.

Would be great if you can also add a sentence to the documentation.

quajak (approved):

quajak (approved):

quajak (approved):

PR #2436, opened 2022-10-11
Mini update of README-FR.md

RaphMar2021:
I have just updated the "veuillez d'abord vérifier les issues" by "veuillez d'abord vérifier les problèmes".


AnErrupTion:
This isn't correct. A problem is different to an issue (at least, software speaking). An issue on GitHub is a bug report most of the time, while a problem is a bug, but not necessarily a report.

Kiirx:
The readme shouldnt even be there and your changes are wrong, if I didnt traducr this its for a reason, this is about the github **issues**

PR #2440, opened 2022-10-15
Decrease VS debugging crashes

quajak:
More work can be done on this but this should improve the current state slightly.


quajak:
@MishaTY  we need to merge this since the il2cpu changes broke a few things which are fixed in this pr

MishaTY (approved):

MishaTY (approved):

PR #2449, opened 2022-10-23
Bug fix: PDB files was locked by the debugger

TheFocusMan:
In the debugger was bugs and they fix, the bugs was:

-  Pdb files was locked by the debugger see [https://github.com/CosmosOS/IL2CPU/pull/174](url)
- Fix crash in DbgCmdCoreDump method


quajak (approved):
Thank you for the PR!

PR #2473, opened 2022-11-18
Add compiler optimizations

MEMESCOEP:
I've added the ability to enable compiler optimizations and choose optimization level for a slight increase in performance (Around 3 fps for me).

### DIFF
--- a/Docs/articles/Express/Compile.md
+++ b/Docs/articles/Express/Compile.md
@@ -1,0 +1,13 @@
+For creating your own Cosmos in Express 2013 you need the following:
+
+ * Visual Studio Express for C# or VB.NET
+ * installed Cosmos Kit(User Kit or compiled from checkout (Dev Kit))
+ * Visual Studio 2013 Shell (Isolated Shell)
+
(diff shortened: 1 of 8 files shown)


terminal-cs:
i wonder if this will make any major differences for other code, as it can vary

MEMESCOEP:
Currently working on making this optional (via a check box in the properties menu)

MEMESCOEP:
@terminal-cs It might, and @quajak suggested that I make it optional (which is a good idea)

quajak (commented):
I think you accidently added a few incorrect files to the PR. Could you please remove them?

The actual changes look good although I think we can simplify the UI a bit and remove the boolean flag for `UseCompilerOptimizations` since we can just include 0 as a valid optimization level. Thanks for the work!

quajak (changes requested):
Pressed the wrong button, please see my previous comment for the review.

MEMESCOEP:
Alright, I'll change the code

MEMESCOEP:
gonna close this pr and make a new one because I can't figure out how to remove some of the files that I accidentally added

PR #2483, opened 2022-11-22
Add xharp checkout back

charlesbetros:
todo:
- [x] Fix kernel tests
- [x] Remove appveyor CI
- [ ] Update RestoreSources.props


.github/workflows/pr.yml
```diff
@@ -71,26 +64,102 @@ jobs:
...
+      #     "testkernels=[$($testKernelsEnv)]" >> $env:GITHUB_OUTPUT
+
+  test:
+    if: ${{ success() }}
+    needs: [build]
+    runs-on: self-hosted
+    
+    strategy:
+      matrix:
+        kernel: [
```
zarlo, line 103: we should add a note somewhere that when you make new tests you need to add them here

PR #2485, opened 2022-11-23
Add option to set GRUB boot timeout

MEMESCOEP:
I've added the option to change the GRUB boot timeout because I think it could be useful during debugging. If the value isn't set it will default to 0 seconds.

### DIFF
--- a/source/Cosmos.Build.Tasks/build/Cosmos.Build.targets
+++ b/source/Cosmos.Build.Tasks/build/Cosmos.Build.targets
@@ -383,12 +383,12 @@
         <CreateGrubConfig TargetDirectory="$(IntermediateIsoDirectory)"
                               BinName="$([System.IO.Path]::GetFileName('$(BinGzFile)'))"
                               Modules="gzio"
-                              timeout="$(Timeout)"
+                              Timeout="$(Timeout)"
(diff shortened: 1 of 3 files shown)


quajak (changes requested):
Thanks for the contribution! Looks good except one small nitpick.

MEMESCOEP:
What needs to be changed?

source/Cosmos.Build.Tasks/build/Cosmos.Build.targets
```diff
@@ -383,10 +383,12 @@
         <CreateGrubConfig TargetDirectory="$(IntermediateIsoDirectory)"
                               BinName="$([System.IO.Path]::GetFileName('$(BinGzFile)'))"
                               Modules="gzio"
+                              timeout="$(Timeout)"
```
quajak, line 386: Just to confirm it ok not to capitalize the timeout in the call here?

MEMESCOEP: Wait I did capitalize that in my code? I'll check it out and see what's going on

MEMESCOEP: Problem has been fixed

quajak (approved):
Thank you!

PR #2488, opened 2022-11-24
Decreased Heap Crashes

TheFocusMan:
Fixed NativeMemory Compile error
Deacresed crashes on small heap
+ Added Asset and Fail methods in the debuger and the user can ignore the Asserts With Debugger.IgnoreAsserts but is not works on the fails

### DIFF
--- a/source/Cosmos.Debug.Hosts/VMware.cs
+++ b/source/Cosmos.Debug.Hosts/VMware.cs
@@ -2,7 +2,9 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Text;
+using System.Threading;
 using Microsoft.Win32;
 
 //using Vestris.VMWareLib;
@@ -14,6 +16,7 @@ namespace Cosmos.Debug.Hosts {
     protected string mDir;
     protected string mVmxPath;
     protected Process mProcess;
+    protected Process[] mProcessesBefore, mProcessesAfter, mProcesses;
     protected string mWorkstationPath;
     protected string mPlayerPath;
     protected string mHarddisk;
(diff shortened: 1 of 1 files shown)


MishaTY (dismissed):

source/Cosmos.Core/Memory/HeapSmall.cs
```diff
@@ -449,9 +471,10 @@ public static void Free(void* aPtr)
             if (size == 0)
             {
                 // double free, this object has already been freed
-                Debugger.DoBochsBreak();
-                Debugger.DoSendNumber((uint)heapObject);
-                Debugger.SendKernelPanic(0x99);
+                //Debugger.DoBochsBreak();
```
quajak, line 474: Can you please undo all the commenting out of the code?

quajak (changes requested):
Thanks for the PR! The main changes themselves look very good, just some small details to change.

source/Cosmos.Core/Memory/HeapSmall.cs
TheFocusMan: Why is need to crash the OS If you accidentaly freeing something not allocated?

TheFocusMan:
> excelent, i've been needing to fix nativememory but i never got around to it, as for the heap, that will be very useful as crashing the system by accidentaly freeing something not allocated is very annoying.

thats why i created the PR

MishaTY (approved):

source/Cosmos.Core/Memory/HeapSmall.cs
quajak: My use case is that if we fail hard it's easier to figure out what's going wrong. This prevents us having difficult to find bugs in the garbage collector for example. For a more user-friendly of the function, i would suggest checking if it's not already freed.

TheFocusMan: I changed it but in way that the user can ignore the kernel panic is is not fail

quajak (dismissed):
Thanks for the changes. The Debugger.Assert approach is interesting and should work. If this does lead to more difficult to detect bugs in the future, we can revert it then but lets try it for now.

source/Cosmos.Core/Memory/HeapLarge.cs
```diff
@@ -35,8 +35,7 @@ public static void Init()
             var xPtr = (uint*)RAT.AllocPages(aType, xPages);
             if (xPtr == null)
```
mukundrd, line 36: Shouldn't the size of `xPtr` to be checked along with `null` check?

quajak: What do you mean with size of `xPtr`?

mukundrd: I mean `xPtr.Length()` to be 2

quajak: Length is not defined for a pointer?

mukundrd: since there's attempt to access the array positions 0 and 1, would be part of sanity to check the length of array along with null check (which is being done already).

jp2masa:
Sorry, disabled auto-merge with a misclick on my phone.

quajak:
Looking at the appveyor test results, it looks like there is still a bug in the changes.

source/Cosmos.Core/Memory/HeapLarge.cs
ascpixi: This is not possible for pointers. Pointers point to raw areas of memory; bound checking is not possible.

quajak:
Why did you close it?

quajak:
Very weird, it looks like all the changes are gone...

PR #2494, opened 2022-11-28
Remove Pen Class

MEMESCOEP:
This is a pr that removes the Pen class and replaces it with System.Drawing.Color (Since the Pen class was mostly just a wrapper anyway)


Jspa2 (changes requested):
ColorEx should probably be removed if it's empty

AsertCreator:
the weird thing was that pen's width property didnt work :P

MEMESCOEP:
Alright, I'll change the code and then do some checks to make sure everything still works.

MEMESCOEP:
> the weird thing was that pen's width property didnt work :P

Before or after my changes? Cause the Pen class doesn't exist anymore (in my code)

AsertCreator:
> > the weird thing was that pen's width property didnt work :P
> 
> Before or after my changes? Cause the Pen class doesn't exist anymore (in my code)

i meant that pen class you just removed was weird, cus it didnt fully work

ascpixi:
> the weird thing was that pen's width property didnt work :P

The width property of all `Pen` instances wasignored by all CGS rendering methods. Maybe it was planned to support different stroke widths, but ultimately it was scrapped... this is the way the default .NET `Graphics` class handles drawing too; with "pens" and "brushes"; however, in standard .NET, it made more sense, since you had things like gradient/texture/custom brushes.

source/Cosmos.System2/Graphics/Extensions/ColorEx.cs
```diff
@@ -8,10 +8,10 @@ namespace Cosmos.System.Graphics.Extensions
     public static class ColorEx
```
ascpixi, line 8: This class can be removed.

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -89,34 +89,34 @@ public override void Disable()
         /// <summary>
         /// Draw point.
         /// </summary>
-        /// <param name="pen">Pen to draw with.</param>
+        /// <param name="acolor">Color to draw with.</param>
         /// <param name="x">X coordinate.</param>
         /// <param name="y">Y coordinate.</param>
         /// <exception cref="Exception">Thrown on memory access violation.</exception>
-        public override void DrawPoint(Pen aPen, int aX, int aY)
+        public override void DrawPoint(Color acolor, int aX, int aY)
```
ascpixi, line 96: The name of the parameter should be `aColor`.

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -128,11 +128,11 @@ public override void DrawArray(Color[] aColors, int aX, int aY, int aWidth, int
         /// Draw point.
         /// Not implemented.
         /// </summary>
-        /// <param name="pen">Pen to draw with.</param>
+        /// <param name="acolor">Color to draw with.</param>
         /// <param name="x">X coordinate.</param>
         /// <param name="y">Y coordinate.</param>
         /// <exception cref="NotImplementedException">Thrown always (only int coordinates supported).</exception>
-        public override void DrawPoint(Pen aPen, float aX, float aY)
+        public override void DrawPoint(Color acolor, float aX, float aY)
```
ascpixi, line 135: Same here

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -141,16 +141,16 @@ public override void DrawPoint(Pen aPen, float aX, float aY)
...
-        /// <param name="aPen">Pen to draw with.</param>
+        /// <param name="acolor">color to draw with.</param>
         /// <param name="aX_start">starting X coordinate.</param>
         /// <param name="aY_start">starting Y coordinate.</param>
         /// <param name="aWidth">Width.</param>
         /// <param name="aHeight">Height.</param>
         /// <exception cref="Exception">Thrown on memory access violation.</exception>
         /// <exception cref="NotImplementedException">Thrown if VMWare SVGA 2 has no rectange copy capability</exception>
-        public override void DrawFilledRectangle(Pen aPen, int aX_start, int aY_start, int aWidth, int aHeight)
+        public override void DrawFilledRectangle(Color acolor, int aX_start, int aY_start, int aWidth, int aHeight)
```
ascpixi, line 151: `acolor` -> `aColor`

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
```diff
@@ -419,14 +419,14 @@ public override void Display()
         /// </summary>
         /// <param name="str">string to draw.</param>
         /// <param name="aFont">Font used.</param>
-        /// <param name="pen">Color.</param>
+        /// <param name="color">Color.</param>
         /// <param name="x">X coordinate.</param>
         /// <param name="y">Y coordinate.</param>
-        public override void DrawString(string str, Font aFont, Pen pen, int x, int y)
+        public override void DrawString(string str, Font aFont, Color color, int x, int y)
```
ascpixi, line 425: While this method signature was like this from the start, can we change all the parameters to adhere to the naming convention defined by the .editorconfig? That way, if we ever will want to change it in the future, we can use a refactoring tool.

source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -263,32 +261,32 @@ internal void DrawDiagonalLine(Pen pen, int dx, int dy, int x1, int y1)
...
         /// <item>Coordinates invalid.</item>
         /// </list>
         /// </exception>
         /// <exception cref="Exception">Thrown on memory access violation.</exception>
         /// <exception cref="OverflowException">Thrown if x1-x2 or y1-y2 equal to Int32.MinValue.</exception>
-        public virtual void DrawLine(Pen pen, int x1, int y1, int x2, int y2)
+        public virtual void DrawLine(Color color, int x1, int y1, int x2, int y2)
         {
-            if (pen == null)
+            if (color == null)
```
ascpixi, line 287: A value type can never be `null`; a null-check is unnecessary here (unless we're not using the `Color` type from `System.Drawing`)

source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -301,50 +299,50 @@ public virtual void DrawLine(Pen pen, int x1, int y1, int x2, int y2)
...
         /// Draw line.
         /// </summary>
-        /// <param name="pen">Pen to draw with.</param>
+        /// <param name="color">Color to draw with.</param>
         /// <param name="p1">Staring point.</param>
         /// <param name="p2">End point.</param>
         /// <exception cref="ArgumentOutOfRangeException">
         /// <list type="bullet">
-        /// <item>Thrown if pen is null.</item>
+        /// <item>Thrown if color is null.</item>
```
ascpixi, line 324: To be removed alongside the null-check

**NOTE:** There are a lot of instances in the documentation where this exception is documented to be thrown, but with the removal of the null-check, these would no longer apply. I don't want to duplicate the same comment over and over again, so I'm only going to leave this one, but there's more lines to update

source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -353,18 +351,18 @@ public void DrawLine(Pen pen, float x1, float y1, float x2, float y2)
...
         /// <exception cref="ArgumentOutOfRangeException">Thrown if coorinates invalid.</exception>
         /// <exception cref="Exception">Thrown on memory access violation.</exception>
-        public virtual void DrawCircle(Pen pen, int x_center, int y_center, int radius)
+        public virtual void DrawCircle(Color color, int x_center, int y_center, int radius)
         {
-            if (pen == null)
+            if (color == null)
             {
-                throw new ArgumentNullException(nameof(pen));
+                throw new ArgumentNullException(nameof(color));
```
ascpixi, line 365: Unnecessary null-check; condition will never be met

source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -971,15 +969,15 @@ public virtual void DrawImage(Image image, int x, int y, int w, int h)
         /// <exception cref="ArgumentOutOfRangeException">Thrown on fatal error.</exception>
         public void DrawImageAlpha(Image image, int x, int y)
         {
-            var _pen = new Pen(Color.White);
+            var _color = Color.White;
```
ascpixi, line 972: Unnecessary assignment to `_color`; the value of `Color.White` will always be replaced with `(Color.FromArgb(image.rawData[_x + _y * image.Width]));`

source/Cosmos.System2/Graphics/Canvas.cs
```diff
@@ -904,15 +902,15 @@ public virtual void DrawRectangle(Pen pen, float x_start, float y_start, float w
         /// <exception cref="ArgumentOutOfRangeException">Thrown on fatal error.</exception>
         public virtual void DrawImage(Image image, int x, int y)
         {
-            Pen _pen = new Pen(Color.White);
+            Color _color = Color.White;
```
ascpixi, line 905: Unnecessary assignment to _color; the value of `Color.White` will always be replaced with `(Color.FromArgb(image.rawData[_x + _y * image.Width]));`

ascpixi (changes requested):
Seems good, but some minor changes could be made to improve the quality of the code! Thanks!

MEMESCOEP:
Alright, I'll change the code in a it!

AsertCreator:
a bit

MEMESCOEP:
I forgor lol

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
valentinbreiz: what do you mean?

valentinbreiz (approved):

source/Cosmos.System2/Graphics/SVGAIICanvas.cs
ascpixi: When following the settings defined in `.editorconfig`, every parameter name should be prefixed with "a". I guess this doesn't really matter here, so on second thought, this can be ignored XD

PR #2505, opened 2022-12-15
Static IOPort class, Mode refactoring and more!

AnErrupTion:
This PR changes the following:

- Made IOPort class static
- Changed Mode.Columns and Mode.Rows to Mode.Width and Mode.Height respectively
- Deleted the now unused Pen class
- Deleted the Point class in Cosmos.System, also removed functions that used Point in Canvas except for DrawPolygons (I've added a comment that explains why)
- Made some minor refactoring like removing unused parentheses or converting fields to auto-properties
- Possibly fixed a bug in Canvas.Clear()


AsertCreator:
auto properties are slower than fields. i think we need opposite

AnErrupTion:
> auto properties are slower than fields. i think we need opposite

When the IL level optimizer for IL2CPU is done, it won't matter much.

AsertCreator:
oh okay

ascpixi:
Great job! Is this still a WIP, or is this ready for review?

AnErrupTion:
> Great job! Is this still a WIP, or is this ready for review?

It's ready for review!

source/Cosmos.HAL2/PCIDevice.cs
```diff
@@ -875,23 +874,17 @@ public PCIBaseAddressBar(uint raw)
...
-        {
-            get { return baseAddress; }
-        }
+        public uint BaseAddress { get; } = 0;
 
-        public bool IsIO
-        {
-            get { return isIO; }
-        }
+        public bool IsIO { get; } = false;
```
quajak, line 888: Do we need the default value? If not, please remove

source/Cosmos.System2/Graphics/FullScreenCanvas.cs
```diff
@@ -30,7 +30,7 @@ public static void Disable()
         }
 
         /// <summary>
-        /// List of all video drivers (BGA, 
+        /// List of all video drivers (BGA,
```
quajak, line 33: Please add the rest of list

source/Cosmos.System2/Graphics/Mode.cs
```diff
@@ -145,11 +132,9 @@ public int CompareTo(Mode other)
         /// <returns>bool value.</returns>
         public static bool operator >=(Mode mode_a, Mode mode_b)
         {
-            int result;
+            var result = mode_a.CompareTo(mode_b);
 
-            result = mode_a.CompareTo(mode_b);
-
-            return (result == 0 || result > 0) ? true : false;
+            return result == 0 || result > 0;
```
quajak, line 137: Simplify to `mode_a.CompareTo(mode_b) >= 0`;

source/Cosmos.System2/Graphics/Mode.cs
```diff
@@ -160,58 +145,33 @@ public int CompareTo(Mode other)
...
-            result = mode_a.CompareTo(mode_b);
-
-            if (result == 0 || result < 0)
-            {
-                return true;
-            }
+            var result = mode_a.CompareTo(mode_b);
 
-            return false;
+            return result == 0 || result < 0;
```
quajak, line 150: Same change here

source/Cosmos.System2/Graphics/VGACanvas.cs
```diff
@@ -87,18 +87,6 @@ public override void Disable()
             }
```
quajak, line 87: Please remove all other methods in this file which just call the base method

quajak (changes requested):
Thank you for cleaning up the code and refactoring it. There are a few things which would increase the code quality a lot further.

Can you please merge all the data in the IOGroup directory classes with the drivers which use the values? Since they are just constants now, I dont see any reason to keep them separate.

Regarding the code style, since most of the code uses the explicit type and not `var`, it would be better if we can continue to use that trend.

source/Cosmos.Core/IOPort.cs
```diff
@@ -52,7 +15,7 @@ protected IOPortBase(ushort aBase, ushort aOffset)
         /// <param name="aPort">A port to write to.</param>
         /// <param name="aData">A data.</param>
         [PlugMethod(PlugRequired = true)]
-        static protected void Write8(ushort aPort, byte aData) => throw null;
+        public static void Write8(ushort aPort, byte aData) => throw null;
```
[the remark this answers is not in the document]
AnErrupTion: The number of I/O ports, on most machines, never exceeds ushort.MaxValue. In fact, x86 even limits you to that. However I don't see how making everything int instead of ushort will save us casting.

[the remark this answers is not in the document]
quajak: Internally in .Net almost everything happens using int32 and anytime a calculation requires something else we need to convert it to that format. This PR itself contains quite a lot of casting to ushorts. Invalid ports can still happen with uint16 so I dont think that they lead to better code but come at a small code quality and runtime cost.

quajak (commented):
The changes look good. Thank you for changing it so quickly. Could you also merge the files in Cosmos.Core/IOGroup with the rest of their implementation?

source/Cosmos.Core/IOPort.cs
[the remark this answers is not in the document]
AnErrupTion: Alright, I'll try to convert as much to Int32 as possible!

source/Cosmos.Core/BaseIOGroups.cs
```diff
@@ -1,48 +1,24 @@
...
+    /// </summary>
+    public class BaseIOGroups {
+        // These are common/fixed pieces of hardware. PCI, USB etc should be self discovering
+        // and not hardcoded like this.
+        // Further more some kind of security needs to be applied to these, but even now
+        // at least we have isolation between the consumers that use these.
+        /// <summary>
+        /// Primary ATA.
+        /// </summary>
+        public readonly IOGroup.ATA ATA1 = new IOGroup.ATA(false);
```
quajak, line 18: If we only use this for ATA we should call this differently. Maybe we could even store the two values in IDE.cs as static values

quajak (changes requested):
Thank you for changing the IOGroups! Just one small thing and I think we can merge

AnErrupTion:
I've made a small change: now the runner checks if QEMU is present in Program Files as well, since that's where QEMU installs by default on Windows.

source/Cosmos.Debug.Hosts/Qemu.cs
```diff
@@ -228,7 +228,12 @@ private void SetDefaultPath()
 
       if (!File.Exists(defaultPath))
       {
-        throw new Exception("The Qemu emulator doesn't seem to be installed on this machine.");
+        defaultPath = @"C:\Program Files\qemu\qemu-system-i386.exe";
+        
+        if (!File.Exists(defaultPath))
+        {
+          throw new Exception("The Qemu emulator doesn't seem to be installed on this machine.");
```
quajak, line 235: Maybe state what paths were checked

quajak (changes requested):

quajak (approved):

MishaTY (approved):

quajak (approved):

PR #2513, opened 2022-12-18
Fix QEMU launching

AnErrupTion:
This PR fixes launching a Cosmos project via QEMU. It adds additional checks for empty options and removes options that should have not been there in the first place (like port forwarding on the guest network).


Kiirx:
maybe replace all the IFs by a switch

Kiirx (changes requested):
Maybe replacing the ifs by a switch

valentinbreiz (approved):

Kiirx:
oh ok

PR #2516, opened 2022-12-18
Allow un and re-masking of IRQs.

terminal-cs:
Title explains it.


AnErrupTion (changes requested):
You'll have to pull from master and change the IOPort read/writes.

source/Cosmos.Core/INTs.cs
```diff
@@ -239,6 +239,40 @@ public struct IRQContext {
...
+
+            if (aIRQLine < 8)
+            {
+                // Read PIC1 Data
+                Port = IOPort.Read8(0x20);
+            }
+            else
+            {
+                // Read PIC2 Data
+                Port = IOPort.Read8(0xA0);
```
AnErrupTion, line 260: According to https://wiki.osdev.org/8259_PIC, if you want to read PIC1 data then Port should be 0x21, and if you want to read PIC2 data it should be 0xA1. See section "ISR and IRR", where PIC1_DATA and PIC2_DATA are defined.

AnErrupTion (changes requested):

source/Cosmos.Core/INTs.cs
terminal-cs: mybad

AnErrupTion: @terminal-cs You shouldn't read from those ports. ``Port`` should directly equal to 0x21 for PIC1 and 0xA1 for PIC2.

quajak:
Is there any way we could test if the enabling/disabling works in TestRunner?

AsertCreator:
pit test? we can disable interrupts but unmask pit irq

source/Cosmos.Core/INTs.cs
terminal-cs: > @terminal-cs You shouldn't read from those ports. `Port` should directly equal to 0x21 for PIC1 and 0xA1 for PIC2.

the port variable is a completely different var, not related to pic1 or 2 addresses

terminal-cs:
> Is there any way we could test if the enabling/disabling works in TestRunner?

i don't know

quajak (dismissed):

TheFocusMan:
> > Is there any way we could test if the enabling/disabling works in TestRunner?
> 
> i don't know

The same problem is in all the PR`s

quajak (approved):

PR #2521, opened 2022-12-19
Change bootloader to Limine

Kiirx:
- [x] Change mkisofs to xorriso
- [x] Create Limine config file
- [x] Add to tasks
- [x] Create Limine Deploy system
- [x] Add custom wallpaper and backdrop system
- [x] Update documentation
- [x] Support for compressed kernels


Kiirx:
the Limine Deploy is to clarify I don't know if it does work, the Cosmos.Build.targets are also a bit confusing for me, there's no sign of life of the LimineDeploy task that I put

MEMESCOEP:
Iirc limine is 64-bit only, and Cosmos builds 32-bit operating systems. Would that cause issues? Also I don't think it's a good idea to only allow booting on x64 based systems

MEMESCOEP:
> > Iirc limine is 64-bit only, and Cosmos builds 32-bit operating systems. Would that cause issues? Also I don't think it's a good idea to only allow booting on x64 based systems
> 
> as I know it does it is just x64 by default
> image
> 

Ahh ok well then if it supports x86 then lets go for it!

MEMESCOEP:
Iirc limine is 64-bit only, and Cosmos builds 32-bit operating systems. Would that cause issues? Also I don't think it's a good idea to only allow booting on x64 based systems

Kiirx:
> Oh BTW if we want to merge this then some of the grub configuration needs to be removed

check the changes

AsertCreator:
will cosmos support limine boot protocol?

Kiirx:
> will cosmos support limine boot protocol?

This would need a complete rework of IL2CPU, some of Cosmos.Core, change consoleimpl plugs, use request features from the protocol, and handling a .h file in C# would be very hard

AsertCreator:
perhaps you need to enable interrupts before `BeforeRun` because it may read keyboard, access timer in tests or in another projects

Kiirx:
> perhaps you need to enable interrupts before `BeforeRun` because it may read keyboard, access timer in tests or in another projects

It already was like that, there must be a reason

source/Cosmos.System2/Kernel.cs
```diff
@@ -1,204 +1,178 @@
 using System;
+using System.Drawing;
```
quajak, line 2: I dont think is needed

source/Cosmos.System2/Kernel.cs
```diff
@@ -1,204 +1,178 @@
...
 using sysIO = System.IO;
 using Cosmos.Debug.Kernel;
 using Cosmos.HAL;
 using Cosmos.System.ScanMaps;
 
-namespace Cosmos.System
+namespace Cosmos.System;
```
quajak, line 8: Are these changes just whitespace changes? I can't tell because of all the formatting changes.

source/Cosmos.Build.Tasks/CreateLimineConfig.cs
```diff
@@ -0,0 +1,80 @@
...
+    [Required]
+    public string BinName { get; set; }
+
+    public string[] Modules { get; set; }
+
+    private string Indentation = "    ";
+
+    public string Timeout { get; set; }
+
+    // public string LimineBackDrop { get; set; }
```
quajak, line 23: Could you please remove all the commented out code in this file?

quajak (changes requested):
Thank you for the great work! Regarding limine, do we need to include all the source code or can we just include the executables?

Tests/Cosmos.TestRunner.Full/TestKernelSets.cs
```diff
@@ -33,7 +33,7 @@ public static IEnumerable<Type> GetStableKernelTypes()
             yield return typeof(NetworkTest.Kernel);
             yield return typeof(AudioTests.Kernel);
             // Please see the notes on the kernel itself before enabling it
-            //yield return typeof(ConsoleTest.Kernel);
+            yield return typeof(ConsoleTest.Kernel);
             // This is a bit slow and works only because ring check is disabled to decide if leave it enabled
```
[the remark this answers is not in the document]
Kiirx: ooohh so thats why it fails

source/Cosmos.Build.Tasks/CreateLimineConfig.cs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private void WriteIndentedLine(TextWriter aWriter, string aText)
+    {
+        aWriter.WriteLineAsync(Indentation + aText);
+    }
+
+    private bool IsNull(string s)
+    {
+        return String.IsNullOrWhiteSpace(s) && String.IsNullOrEmpty(s);
```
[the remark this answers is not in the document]
Kiirx: This isnt my code but maybe

source/Cosmos.System2/Kernel.cs
Kiirx: Somehow it got there

source/Cosmos.System2/Kernel.cs
Kiirx: yes Rider reformatted the file

source/Cosmos.System2/Keyboard/ScanMaps/US_Standard.cs
```diff
@@ -60,8 +60,8 @@ protected override void InitKeys()
             _keys.Add(new KeyMapping(0x1A, '[', '{', '[', '{', '[', '{', ConsoleKeyEx.LBracket));
             _keys.Add(new KeyMapping(0x1B, ']', '}', ']', '}', ']', '}', ConsoleKeyEx.RBracket));
             /* ENTER, CTRL */
-            _keys.Add(new KeyMapping(0x1C, ConsoleKeyEx.Enter));
-            _keys.Add(new KeyMapping(0x1D, ConsoleKeyEx.LCtrl));
+            _keys.Add(new KeyMapping(0x9C, ConsoleKeyEx.Enter));
```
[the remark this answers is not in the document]
Kiirx: It was using keypad scancodes

Kiirx:
> Thank you for the great work! Regarding limine, do we need to include all the source code or can we just include the executables?

I put all of them first but yeah if needed I can remove the unused ones

quajak:
> I put all of them first but yeah if needed I can remove the unused ones

I would prefer if we only include what we need

Kiirx:
@quajak I removed the unused ones except for limine-pxe in case it is going to be used later

quajak (approved):

AsertCreator:
there is a stack corruption. do we give cosmos enough stack memory? or limine gives its own stack?

Kiirx:
> there is a stack corruption. do we give cosmos enough stack memory? or limine gives its own stack?

it doesnt uses the protocol so there's no reason to give its own stack, the tests gives a timeout exception somehow

terminal-cs:
> > there is a stack corruption. do we give cosmos enough stack memory? or limine gives its own stack?
> 
> it doesnt uses the protocol so there's no reason to give its own stack, the tests gives a timeout exception somehow

Recently all PRs have been doing this, all of them time out

quajak:
> Recently all PRs have been doing this, all of them time out

I dont think so, for this one, one specific test is failing with a timeout issue. That makes me believe that the issue is specific to the PR. @Kiirx are you able to reproduce the failure locally?

Kiirx:
I dont get any timeout errors but this `Error: Error occurred: System.IO.FileLoadException: Could not load file or assembly 'Cosmos.Build.Common, Version=10.0.0.0, Culture=neutral, PublicKeyToken=5ae71220097
cb983'. The located assembly's manifest definition does not match the assembly reference. (0x80131040)
` 
on my side, a normal project runs fine @quajak

quajak:
Can you clear the nuget cache and try again?

Kiirx:
> Can you clear the nuget cache and try again?

I get the same error

quajak:
Are you getting this when running the TestRunner? During building of the TestRunner or when it tries to compile the test kernel?

Kiirx:
> Are you getting this when running the TestRunner? During building of the TestRunner or when it tries to compile the test kernel?

when it tries to compile the test kernel if I remember

quajak:
Try do a dotnet clean and try again. It sounds like you have an old version of one of the dependency nugets still lying around somewhere.

Kiirx:
I still cant get it to work could tou try on your end?

source/Cosmos.System2/Keyboard/ScanMaps/US_Standard.cs
quajak, line 63: I think this is causing the test to fail since this PR breaks pressing enter in the console no longer works. See https://github.com/CosmosOS/Cosmos/blob/9d59ffa7bc06dcbe3268cd5906ce9e69c500a438/Tests/Kernels/Cosmos.Compiler.Tests.SingleEchoTest/Kernel.cs#L30 for the reason the test is failing.

quajak (changes requested):
I was able to reproduce the issue locally

Kiirx:
Ok so this should be the final fix

quajak (approved):

PR #2537, opened 2022-12-27
turn demos into examples

AsertCreator:
why not. i have no idea why not. plus i added one more example and updated CGS docs


quajak (approved):
I think the organisation of all examples in one sln makes sense. The new demo also fills a nice gap. Thank you for updating the documentation

PR #2539, opened 2022-12-28
expose serial ports

AsertCreator:
why not?


quajak (approved):
Looks better now. Thanks for improving the design

PR #2541, opened 2022-12-30
implement ramdisk and refactor

AsertCreator:
why it doesn't work? so i made this rAmDisK thing, but it just crashes.


terminal-cs:
> why it doesn't work? so i made this rAmDisK thing, but it just crashes.

i assume this is for tar files?

AsertCreator:
no, this is just for ramdisk, but it is possible to read tar files and convert them into ramdisks

quajak:
Where does it crash?

AsertCreator:
it crashes when i create new ramdisk

AsertCreator:
and for some reason test is not starting

AsertCreator:
oh there is actually syntx error :P

AsertCreator:
ok, i think this is more user feature and developers should implement that themselves. im changing concept of pr and going to implement ustar filesystem

PR #2548, opened 2023-01-04
Add support for dotnet new project creation

quajak:
Feedback welcome!
This is a MVP and I am not sure what additional features are useful.

1. This currently uses path based installation, does it make sense to package the template as a nuget package? If yes, would it only be part of the local nuget store or do we publish it to nuget.org?

2. Not sure how to best integrate this automatically in the installation process since the syntax for `dotnet new` has changed between .Net 6 and .Net 7 (https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-new-install). Should this be included in the inno setup and the makefile? If yes we would have to detect which version the user has installed.


valentinbreiz:
Could dotnet list be added too?

valentinbreiz (approved):

terminal-cs:
why not have it install the template automaticaly when the make command is ran?

quajak:
> "dotnet new list" will show a cosmos project too?

Yes, thats the goal

> why not have it install the template automaticaly when the make command is ran?

As I stated in the comment, the syntax for the command changed from `dotnet new install` in .net 6 to `dotnet new --install` .net 7 and I don't know a good way of handling both cases. 

> if you include it in the inno setup i think that will make it show up 2 times in VS

Yeah, we can get rid of vs templates if we include in inno setup.

MishaTY:
Will fix all makefile issues in a separate PR

PR #2563, opened 2023-01-22
Add paging support

MishaTy:
Very WIP - will not work yet


Kiirx:
if it doesnt work yet you should put it as a draft

ascpixi:
I'm not sure if this is still active, but looking at the code, this will not work, since in its current state, it's using PDPTs, while nowhere the `CR4.PAE` bit is set.

MishaProductions:
Closing this PR as the correct place to implement this is in IL2CPU

MishaTy:
will make a new PR

PR #2568, opened 2023-01-25
build xsharp as nuget package linux only

zarlo:


zarlo:
i tried to make to work on the windows builder too but it did not work and i am unable to test it

MishaTY:
doesn't seem to build it
https://hastebin.com/johaxigoli.sql

zarlo:
> doesn't seem to build it https://hastebin.com/johaxigoli.sql

are you looking at an old commit as all i did was edit the make file

PR #2593, opened 2023-03-05
Minor refactoring & Implement swap for memoryblock

terminal-cs:
This allows for faster video operations in graphics.


terminal-cs:
Will not work

PR #2603, opened 2023-03-13
Implement string[] to the debugger

husaral2:
I don't know how to test it so it might not work at all

### DIFF
--- a/source/Cosmos.Debug.Kernel.Plugs.Asm/DebuggerAsm.cs
+++ b/source/Cosmos.Debug.Kernel.Plugs.Asm/DebuggerAsm.cs
@@ -16,9 +16,6 @@ namespace Cosmos.Debug.Kernel.Plugs.Asm
     [PlugMethod(Assembler = typeof(DebugDoSend))]
     public static void DoSend(string aText) { }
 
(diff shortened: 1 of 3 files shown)


quajak:
Thanks for the PR. To test the changes you can use the debugger in your kernel. You can either create an instance of https://cosmosos.github.io/api/Cosmos.Debug.Kernel.Debugger.html and then call the Send method. Alternatively you can all the Debugger.DoSend static method directly. If you are using Cosmos with VS the debugger output will be shown in the output window (This requires the debugger to be enabled). Alternatively the TestRunner will show any debug messages sent as well. 

Please test the features, since I do not know how cosmos will react to the dosend plug getting passed an array. If it doesnt work, it might be easier to convert the string array to a string before using the plug.

husaral2:
It didn't work at first because I used a plug which took only string and when I tested, it neither worked nor crashed.
So I wrote a function that takes a string array and prints all of the strings in the array and it works so far (by making a new instance of Debugger class and using DoSend function).

quajak (approved):
Looks good now! Thanks for the work. Since you said, that you tested it I think we can merge this.

MishaTy (approved):

PR #2606, opened 2023-03-14
Change syntax standard to .NET's standard.

terminal-cs:


ascpixi (changes requested):
.editorconfig changes look good to me, but shouldn't we run Code Cleanup to refactor the existing methods?

terminal-cs:
> .editorconfig changes look good to me, but shouldn't we run Code Cleanup to refactor the existing methods?

I will do this in a separate PR manually.

.editorconfig
```diff
@@ -9,8 +9,8 @@ root = true
 # Use tabs as indentation
 # Trim trailing whitespace
 [*]
-insert_final_newline = true
-indent_style = space
+insert_final_newline = false
+indent_style = tabs
```
ascpixi, line 13: I know it's a bit of a heated debate in the programming world, but is there a reason to switch this to tabs instead of spaces? ^^'>

ascpixi (changes requested):
Well, alright, I'd do this in one PR though, as on its own, this PR is extremely small and updates one part of the project without reflecting the changes in the code-base. But, having said that, updating the project's used naming convention to be the C# standard is good.

There are some minor nitpicks I have, but otherwise, a step in the good direction! :)

.editorconfig
```diff
@@ -66,8 +66,7 @@ dotnet_style_qualification_for_method = false:warning
 
 dotnet_naming_style.pascal_case.capitalization = pascal_case
 
-dotnet_naming_style.parameters.capitalization = pascal_case
-dotnet_naming_style.parameters.required_prefix = a
+dotnet_naming_style.parameters.capitalization = camel_case
 
 dotnet_naming_style.private_fields.capitalization = pascal_case
 dotnet_naming_style.private_fields.required_prefix = _
```
ascpixi, line 72: In the standard C# naming convention, private fields are either camelCase or _camelCase.

terminal-cs:
> Well, alright, I'd do this in one PR though, as on its own, this PR is extremely small and updates one part of the project without reflecting the changes in the code-base. But, having said that, updating the project's used naming convention to be the C# standard is good.
> 
> There are some minor nitpicks I have, but otherwise, a step in the good direction! :)

The core devs said to do it separately

.editorconfig
terminal-cs: > I know it's a bit of a heated debate in the programming world, but is there a reason to switch this to tabs instead of spaces? ^^'>

Aren't tabs the standard? And yes, managing whitespace using tabs is easier

ascpixi: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#layout-conventions
> Use the default Code Editor settings (smart indenting, four-character indents, **tabs saved as spaces**). For more information, see [Options, Text Editor, C#, Formatting](https://learn.microsoft.com/en-us/visualstudio/ide/reference/options-text-editor-csharp-formatting).

Although, I'm not sure myself. I personally use spaces because it's easier to indent comments after variable/constant declarations, like so:
```cs
const int SomeConstant = 3;        // A constant value of "3".
const int SomeOtherConstant = 5;   // A constant value of "5".
const int NinePlusTen = 21;        // The result of the arithmetic operation "9 + 10".
const int AValueMaybe3 = 6;        // A value that's possibly 3, with no guarantee.
```
The IDE can help with automatically indenting this when using tabs. However, tabs are also good, so I'm gonna dismiss this comment.

ascpixi (dismissed):

zarlo:
why don't we use the .editor config from https://github.com/dotnet/runtime/blob/main/.editorconfig then we would be using the same as the standard library for .net

terminal-cs:
> why don't we use the .editor config from https://github.com/dotnet/runtime/blob/main/.editorconfig then we would be using the same as the standard library for .net

Cuz some parts of it aren't that great
This PR is a combination of (some of the) better syntax options, such as using tabs and not requiring '_' for private & internal fields

.editorconfig
quajak: I talked with a few other coredevs and we would prefer if we stay with spaces.

quajak (approved):

Issue #2607, opened 2023-03-14
Improvements to the XMLDoc documentation comments


ascpixi:
In a lot of areas of Cosmos, the XMLDoc documentation comments are either lackluster (e.g. being a transliteration of the member name) or non-existent. After improving the documentation comments, a more complete API documentation reference can be created automatically with tools like Doxygen or the current one used for cosmosos.github.io.

quajak:
Since Cosmos is quite large, its impossible to do this all at once. It would be helpful if we track a list of areas, where the changes are especially impactful/important.

valentinbreiz:
Cosmos Gen3 has replaced the IL2CPU-based Gen2 codebase, and `gen3` is now the default branch of this repository. This issue was filed against Gen2, which is no longer developed, so it is being closed as part of the Gen3 cutover (#3088).

If you still hit this on Gen3, please open a fresh issue. The compiler, the build pipeline and the debugging story are all different now, so a new report will get a far more useful answer than reviving this one. Start here: https://cosmosos.github.io/

Gen2 remains on the `gen2` branch, and this issue stays readable.

PR #2613, opened 2023-03-22
Code cleanup phase 1: Cosmos.System

ascpixi:
This pull request aims to improve the code quality of the Cosmos.System namespace, along with adding/updating XMLDoc documentation comments (partially resolving #2607). This PR does not update plugs.

- [ ] Cosmos.System.Network
- [ ] Cosmos.System.Helpers
- [ ] Cosmos.System.Keyboard
- [x] Cosmos.System.Text
- [x] Cosmos.System

As a file system overhaul is planned, the file system namespace has been excluded.

### DIFF
--- a/source/Cosmos.HAL2/Power.cs
+++ b/source/Cosmos.HAL2/Power.cs
@@ -1,34 +1,34 @@
+using System;
 using Cosmos.Core;
 
 namespace Cosmos.HAL
 {
     /// <summary>
-    /// Power class. Used to reboot and shutdown the PC.
+    /// Manages the power state of the system by directly manipulating
+    /// hardware devices.
     /// </summary>
     public class Power
     {
-        //Reboot with CPU
         /// <summary>
-        /// Reboot the CPU.
+        /// Triggers a direct, raw CPU reboot.
         /// </summary>
         public static void CPUReboot()
         {
-            Core.CPU.Reboot();
+            CPU.Reboot();
         }
 
-        //Reboot with ACPI
         /// <summary>
-        /// Reboot using ACPI.
+        /// Reboot the system using ACPI. Currently not implemented.
         /// </summary>
         /// <exception cref="NotImplementedException">Thrown always.</exception>
+        [Obsolete("This method is not yet implemented.", error: true)]
         public static void ACPIReboot()
         {
-            ACPI.Reboot(); //TODO
+            ACPI.Reboot();
         }
 
-        //Shutdown with ACPI
         /// <summary>
-        /// Shutdown the ACPI.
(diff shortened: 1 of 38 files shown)


ascpixi:
Note: I'm following the 80 column rule for the XMLDoc comments - if we have a different limit for the amount of characters in a single line (excluding leading whitespace), please let me know.

quajak:
Greatly appreciate the effort! I will try to review the changes in detail but skimming through it looks good for now.

> Note: I'm following the 80 column rule for the XMLDoc comments - if we have a different limit for the amount of characters in a single line (excluding leading whitespace), please let me know.

That sounds fine as long as it is consistent.

ascpixi:
I've finished the network stack cleanup, however, I have changed the debugger from the global one to the already existing network stack one. I'm not sure how this will affect the execution of end-user kernels, and I don't really have time to set up a testing environment, so before this is merged, this will need to be tested.

ascpixi:
@quajak The PR should be done! I'll launch a test build in a second to verify that this does indeed compile, but the PR is ready for review.

source/Cosmos.System2/Graphics/Mode.cs
```diff
@@ -82,89 +76,13 @@ public int CompareTo(Mode other)
             return 0;
         }
 
-        /// <summary>
```
zarlo, line 85: do we want to remove the operator overrides?

ascpixi: The overrides aren't removed - I only removed the unneeded XMLDoc comments and reformatted the overrides to fit on a single line. You can see the overrides below the line deletions.

zarlo: o i see them now

quajak: Is there any reason to do that?

ascpixi: Usually, the XMLDoc comments aren't visible for operator overloads (and they should be self-documenting). This is mainly for code clarity, as these operators simply use the `CompareTo` methods.

quajak (changes requested):
Thanks for the work! The cleanup looks good except for the changes to the bracket placement. I would prefer if we stay with 
```
if(x)
{
    y
}
else
{
    z
}
```
The changes sometimes changed this.

quajak (dismissed):
Thank you!

ascpixi:
Oops, looks like the GB keyboard layout PR used the pre-refactor field names... I've updated the files in question, the PR should be ready to go now!

quajak (approved):

PR #2617, opened 2023-03-26
remove the need to run make install as root after first run

zarlo:


zarlo:
ill come back to this

Issue #2608, opened 2023-03-16
Publish cosmos libraries to nuget.org


terminal-cs:
#### Area of Cosmos - What area of Cosmos are we dealing with?
Cosmos/General

#### Expected Behaviour - What do you think that should happen?
n/a

#### Actual Behaviour - What unexpectedly happens?
n/a

#### Reproduction - How did you get this error to appear?
n/a

#### Version - Were you using the User Kit or Dev Kit? And what User Kit version or Dev Kit commit (Cosmos, IL2CPU, X#)?
n/a

#### Reason/Explanation

We should be using nuget.org to publish cosmos packages. Ideally, we should have a GitHub action run every time a PR is merged to update the libraries on nuget.org. This would both allow users to get instant updates to the core libraries without having to update cosmos every time, And it would also greatly ease using cosmos projects in testing/documentation software via GitHub actions, so that cosmos doesn't need to be set up again every time an action runs. Optionally we could have 2 channels (stable & developer), however, this may be a bit more complicated. If it is done, we would need to make sure that existing projects switch to and from the stable and dev channels correctly when loaded when either the user-kit or the dev-kit is installed.

Edit: This would also improve the usability of cosmos on linux because we won't have to worry about local-only nuget packages being installed properly, which has issues currently.

zarlo:
we will not likely do this any time soon as we use some hard coded paths and it would not work as cosmos still needs to be installed as not every part of it can be made into a nuget package

quajak:
Is it possible to include everything which is currently stored in the User kit directory in a nuget package?

charlesbetros:
We will most likely start with publishing packages to our feed in GitHub rather than nuget.org. The kernel libraries are already nuget packages and are installed to a local nuget feed. They are hood candidates to move out of the user kit and in to our GitHub feed.

terminal-cs:
> Is it possible to include everything which is currently stored in the User kit directory in a nuget package?

All of the DLLs yes

quajak:
Yeah thinking about it more, there is no real advantage to the local package source. If i remember correctly, the GitHub nuget package source is very difficult to use since you cannot make it publicly available (you need access rights) but they might have finally changed that.

zarlo:
[...]

PR #2618, opened 2023-03-26
more nuget

zarlo:
work for #2608 

this will be a breaking change we may want to make a userkit after this


zarlo:
will need someone to update the windows installer

zarlo:
okay can confirm this works on windows

quajak:
> this will be a breaking change we may want to make a userkit after this

Do users have to update existing projects to use nuget?

zarlo:
> > this will be a breaking change we may want to make a userkit after this
> 
> Do users have to update existing projects to use nuget?

yes they will need to add the nuget package ```cosmos.plugs```

quajak:
Can you update the documentation somewhere to reflect that?

zarlo:
yep ill add a breaking change log to the articles and add a link to it in the wiki and read me

quajak:
Thanks, could you please fix the spelling and capitalisation in the text?

zarlo:
okay i think its all fixed, if there are others they will need to be pointed out

quajak (approved):

PR #2623, opened 2023-03-27
Added GB keyboard layout

Owen2k6:
altGR keys such as € á é ú í ó e.t.c have been excluded and i plan to work on at a better time than 9PM on a monday


Owen2k6:
As a note, i ran the build sucessfully and tested it in a VM. on 2 separate keyboards it indeed works as intended. however i will leave review up to the reviewers ofc.

9xbt (commented):
All seems fine to me, can't test it out because I don't have a british keyboard but it seems like nothing is going to break.

9xbt (approved):
All seems fine to me, can't test it out because I don't have a british keyboard but it seems like nothing is going to break.

MishaTy (approved):

MishaTy:
Thanks for the contribution!

PR #2626, opened 2023-03-29
Fix CGS documentation

RaphProductions:
Fix the old documentation that was still using Pen (even Brush 😧)


Docs/articles/Kernel/CGS.md
```diff
@@ -78,7 +80,7 @@ namespace GraphicTest
         {
             try
             {
-                Pen pen = new Pen();
+                Color col = new Color();
```
MishaTy, line 83: You can't create an instance of Color

AsertCreator: better remove it, because that variable is not used

Blackneeed: Yeah, just rm it

PR #2629, opened 2023-04-05
fixed call to change encoder in ConsoleTest Kernel.cs

Owen2k6:


quajak (approved):
Thank you for the fix! We should investigate at some point why we no longer plug the Encoding correctly but this is good for now.

PR #2631, opened 2023-04-06
Some refactoring

terminal-cs:


source/Cosmos.HAL2/PCSpeaker.cs
```diff
@@ -85,9 +85,13 @@ public static void Beep(uint frequency)
         /// <exception cref="ArgumentOutOfRangeException">Thrown if duration or frequency invalid.</exception>
         public static void Beep(uint frequency, uint duration)
         {
+            if (frequency < 37 || frequency > 32767)
+            {
+                throw new ArgumentOutOfRangeException("'frequency' must be between 37 and 32767Hz.");
```
GoldenretriverYT, line 90: The correct way to do that would be `throw new ArgumentOutOfRangeException("frequency", "Value must be between 37 and 32767");` as the first constructor argument is paramName and the second one is message

terminal-cs: > The correct way to do that would be `throw new ArgumentOutOfRangeException("frequency", "Value must be between 37 and 32767");` as the first constructor argument is paramName and the second one is message

Opposite of that, but thanks

ascpixi: > Opposite of that, but thanks

The signature for the `ArgumentOutOfRangeException` constructor is `public ArgumentOutOfRangeException (string? paramName, string? message);`. I'd recommend using `throw new ArgumentOutOfRangeException(nameof(frequency), "Value must be between 37 and 32767");`, since that's the standard way to throw such exceptions, though.

jwoff78:
I don't think values like get_ set_ can be changed such as `get_CursorSize()` because its referencing IL fields from things like .property show in this  with ildecompiler, `get_` `set_` in plugs correlate to IL

getset

quajak (approved):

quajak:
> I don't think values like get_ set_ can be changed such as `get_CursorSize()` because its referencing IL fields from things like .property show in this with ildecompiler, `get_` `set_` in plugs correlate to IL
> 
> getset

I dont fully understand your point. I think the properties get compiled to the corresponding get_, set_ methods, so what terminal did was more or less just a "rename".

terminal-cs:
> > I don't think values like get_ set_ can be changed such as `get_CursorSize()` because its referencing IL fields from things like .property show in this with ildecompiler, `get_` `set_` in plugs correlate to IL
> > getset
> 
> I dont fully understand your point. I think the properties get compiled to the corresponding get_, set_ methods, so what terminal did was more or less just a "rename".

Yep, console colors still work and they were originaly get_ and set_ methods

quajak (approved):

PR #2632, opened 2023-04-07
Refactoring 2

terminal-cs:
Also plugs Environment.CurrentDirectory


source/Cosmos.System2_Plugs/System/IO/DirectoryInfoImpl.cs
```diff
@@ -16,13 +13,13 @@ namespace Cosmos.System_Plugs.System.IO
     public static class DirectoryInfoImpl
     {
         /* The real implementation uses IEnumerable and do a conversion ToArray that crashes IL2CPU */
-        public static DirectoryInfo[] GetDirectories(DirectoryInfo aThis)
+        public static DirectoryInfo[] GetDirectories(DirectoryInfo thisContext)
```
quajak, line 16: I don't like this naming scheme, can we just call it context or directory?

quajak (changes requested):
Please change that one naming thing. Regarding CurrentDirectory, I think we should only support it if Cosmos works well with it since otherwise people will just get confused and we get more bugs. If that is the case, we can go ahead and add the plug.

source/Cosmos.System2_Plugs/System/IO/DirectoryInfoImpl.cs
terminal-cs: K

quajak (approved):

PR #2653, opened 2023-04-23
Implement MBR creating and partition information adding

husaral2:
Partition tables and new partitions can be created now.


quajak (changes requested):
Thank you the the PR!

Would it possible to add some tests to the test runner (for example [Diskmanager tests](https://github.com/CosmosOS/Cosmos/blob/master/Tests/Kernels/Cosmos.Kernel.Tests.DiskManager/Kernel.cs) to ensure that the code works as expected?
Additionally, it would be great if you could add a small section to the documentation to explain to other users when and how to use the functions.

husaral2:
Added the docstrings, made the changes and changed the test kernel to create MBR and write partition information at first when the disk starts to format the disk.

quajak (dismissed):
Thanks for the changes!

quajak:
Looks like its causing the DiskManager tests to fail. Have you been able to reproduce it locally?

husaral2:
> Looks like its causing the DiskManager tests to fail. Have you been able to reproduce it locally?

It was because I tried to access the partition after I recreated the partition table. It should work now.

quajak (approved):

PR #2659, opened 2023-04-28
Make the plug documentation clearer

GoldenretriverYT:
People may miss that part considering its at the very bottom, also made it stand out a bit more


9xbt:
make it bigger so even blind people can see it (well maybe not as much but u get it)

GoldenretriverYT:
eh idk I think thats visible enough hopefully

quajak (approved):

quajak:
Good idea!

PR #2662, opened 2023-05-03
Fix GRUB reference in CosmosPropertyPage.xaml

MEMESCOEP:
I've changed "GRUB" to "Limine" because we aren't using grub anymore.


Kiirx:
maybe instead we should let the person choose between grub and limine

Kiirx:
then instead modify this not to limine but only "bootloader" to make a more general thing

source/Cosmos.VS.ProjectSystem/BuildSystem/Rules/CosmosPropertyPage.xaml
```diff
@@ -85,7 +85,7 @@
     </EnumProperty>
 
     <StringProperty Name="Timeout"
-                    DisplayName="GRUB Menu Timeout"
+                    DisplayName="Limine Menu Timeout"
```
Kiirx, line 88: Display="Bootloader Menu Timeout" instead

Kiirx (changes requested):

MEMESCOEP:
> then instead modify this not to limine but only "bootloader" to make a more general thing

Good Idea, I'll get on that now

Kiirx (approved):
Yep good

PR #2670, opened 2023-05-21
Made Startup.md a bit more formal in the String.Empty section

Winksplorer:
All this PR does is change this section in Startup.md:

>  First, it checks if `System.String.Empty` was not... Uhh, thrown away by the compiler.

Into this:

> First, it checks if `System.String.Empty` is null. If it is null, then it will just throw an exception. If it isn't, it just continues.

This is to make the documentation a bit more formal.


quajak (approved):
Thanks!

PR #2675, opened 2023-05-22
Serial communication related changes

9xbt:
Made it possible to change the baud rate of a serial port when enabled and moved COM ports to an enum.


9xbt:
Crap why does it fail all the time

zarlo:
Constant value '384' cannot be converted to a 'byte'  source\Cosmos.HAL2\BaudRate.cs
Line: 63
Column: 23
Cosmos.HAL2

9xbt:
Yes i know i figured that out a while ago

9xbt:
It should be fixed now

PR #2677, opened 2023-05-23
Serial port communication related changes

9xbt:
Added support for changing baud rate in serial port communication, move COM ports to an enum


9xbt:
Why is it taking so long, why did the build process freeze

quajak (approved):
Thanks!

PR #2686, opened 2023-06-04
Enum.HasFlag

pogrammerX:
Implement Enum.HasFlag.

Sample:
```cs
public static bool EnumHasFlag(this Enum enum, Enum value) {
     return (enum & value) == value;
}
```


quajak:
The PR seems to be missing the implementation code.

quajak:
Please reopen this PR with the code change

PR #2689, opened 2023-06-07
Fixes for the PIT

GoldenretriverYT:
The pit can now run at higher frequencies. Making it too high will still cause a stack overflow (because it doesnt even disable the interrupts in time, lmao) and other bogus errors.

It could very well still be improved further, but its definitely an improvement to the like 20 calls/second previously, lol

### DIFF
--- a/Tests/Kernels/ProcessorTests/Kernel.cs
+++ b/Tests/Kernels/ProcessorTests/Kernel.cs
@@ -30,6 +30,7 @@ namespace ProcessorTests
                 TestCycleCount();
                 TestCycleRateIsNotZero();
                 TestMultiboot();
+                TestPit();
 
                 TestController.Completed();
             }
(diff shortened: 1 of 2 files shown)


terminal-cs:
Have you been able to record what the max stable frequency is?

terminal-cs:
> > Have you been able to record what the max stable frequency is?
> 
> For the test I added (which basically just is incremeting i by 1) I got to 6500 times per second, but it probably could still go a bit higher if I would have tested other values more instead of just going slower by factors of 10 each time
> 
> I still _think_ it depends on how long the handlers itself take because even with the stackoverflows being rarer now, it still likes to do weird stuff (I got random NullReferenceExceptions which were fixed after decreasing the frequency)

Ok so probably just as long as whatever the target is finishes before it calls again, it should be fine

quajak (approved):

quajak:
Thank you for the PR! 

That's a very weird appveyor failure, let us see if it is consistent. If it is, then there might be some mistake with how the GC enables/disables IRQ.

terminal-cs:
> Thank you for the PR! 
> 
> That's a very weird appveyor failure, let us see if it is consistent. If it is, then there might be some mistake with how the GC enables/disables IRQ. 

I've always had issues with the GC and pit, e.g. trying to collect when a pit runs, it crashes

zarlo:
o yes when i was making the threading PR into library when ever the GC would run that would kill PIT

terminal-cs:
I tried your PR and it still seems to be limited to 19 hz

GoldenretriverYT:
> I tried your PR and it still seems to be limited to 19 hz

you have to change the frequency too

terminal-cs:
> > I tried your PR and it still seems to be limited to 19 hz
> 
> you have to change the frequency too

How do I do that?/what do I set it to?

GoldenretriverYT:
> > > I tried your PR and it still seems to be limited to 19 hz
> > 
> > 
> > you have to change the frequency too
> 
> How do I do that?/what do I set it to?

PIT.T0Frequency and to, well, the frequency you want it to run at lol

terminal-cs:
> > > > I tried your PR and it still seems to be limited to 19 hz
> > > 
> > > 
> > > you have to change the frequency too
> > 
> > 
> > How do I do that?/what do I set it to?
> 
> PIT.T0Frequency and to, well, the frequency you want it to run at lol

T0 isn't public, you should make it public

GoldenretriverYT:
T0Frequency is public.

GoldenretriverYT:
oopsie doopsie i am little silly 🤪

well it didnt pass the tests anyways lol

PR #2694, opened 2023-06-14
Simplify makefile

terminal-cs:
Simplifies the makefile to just ``make`` instead of multiple commands, and prompts the user for a password when needed.

also fixes https://github.com/CosmosOS/Cosmos/issues/2697


quajak (approved):

zarlo:
* you should check if sudo is installed it does not come installed on all disros 
* you should run the whole install process as root because it would make it a pain to work on as bin and obj will be owned as root and its well not safe

zarlo:
the only line that always needs root is ```@echo $(DESTDIR) > /etc/CosmosUserKit.cfg```

terminal-cs:
> the only line that always needs root is `@echo $(DESTDIR) > /etc/CosmosUserKit.cfg`

On my distro /opt/* is read only by non-root users, so that's not always the case

zarlo:
> > the only line that always needs root is `@echo $(DESTDIR) > /etc/CosmosUserKit.cfg`
> 
> On my distro /opt/* is read only by non-root users, so that's not always the case

that is true but its not uncommon to change the owner of folders in /opt/ on single user systems

in any case this build commands should not be ran as root the mkdir and cp can be to keep it simple

zarlo (approved):

valentinbreiz (approved):

PR #2695, opened 2023-06-14
Return used memory in MBs than bytes

RaphMar2021:
Since GetAvailableRAM is returned in MBs, why not returning used memory in MBs?


terminal-cs:
Because then there's less info, it's better to have granular info on small data

AsertCreator:
i think we need to make `GetAvailableRAM` to also return value in bytes, becuase mbs are too big

valentinbreiz:
Thanks for the PR but we are losing information using MB. We should stay in bytes and let the user convert as he wants. I close this

PR #2698, opened 2023-06-15
Convert object to string in ConsoleImpl.Write(object) and ConsoleImpl.WriteLine(object)

Z3RYX:
Before, if you would pass an object to Console.Write or Console.WriteLine without converting it to a string, it would hang the system and output a mess of a color fest (at least in my case).

This was easily fixed by calling ToString on the object, which makes the Write (and WriteLine) method behave as it should.


quajak (approved):
Thanks, I think previously we would have infinite recursion leading to a crash.

PR #2703, opened 2023-06-17
Fix ISO9660

GoldenretriverYT:
Fixes ISO9660

It was caused by the VFSManager looking for partitions, but on a CD there usually are no partitions. So I just made CDs act like one big partition and this fixed the problem

image


GoldenretriverYT:
do not merge yet i have found one small issue i wanna fix first

GoldenretriverYT:
ok can be merged now, the old implementation had a bug where directory names were incomplete

quajak (approved):
Thank you for the fixes!

RaphMar2021:
yey i can finally install a cosmos os on my pc

PR #2714, opened 2023-06-21
Add Object.ToString

GoldenretriverYT:
Adds a small Object.ToString implementation that shows
`<object of TYPENAME>`

so like
`<object of EventArgs>`


jvyden (changes requested):
.NET itself just returns the name directly in this scenario, no surrounding text. I think it would be much better if you modified it to be consistent with .NET here.

quajak:
> .NET itself just returns the name directly in this scenario, no surrounding text. I think it would be much better if you modified it to be consistent with .NET here.

I think thats a good idea

jvyden (approved):

quajak (approved):

PR #2720, opened 2023-06-25
Fix Encodings

9xbt:
Revert the console to `ed9f12ae0c1487988f6eafc89fdd8ab47807092f`, which fixes the encoding issue.


9xbt:
I could have just fixed the area that was broken, but I didnt know the area that was broken so I just replaced the entire console. Some newer console edits may be lost, which sucks, but I can't do anything about it.

9xbt:
So it was just some CIL stuff, I'll leave the fix for @terminal-cs

PR #2721, opened 2023-06-25
Verify length when constructing a string

### LINKED ISSUE #2717: Making a string with a negative length should just cause an exception
#### Area of Cosmos - What area of Cosmos are we dealing with?
System.String

#### Expected Behaviour - What do you think that should happen?
`new string(' ', -1);` should just throw `System.ArgumentOutOfRangeException`

#### Actual Behaviour - What unexpectedly happens?
Causes a CPU Exception

#### Reproduction - How did you get this error to appear?
`new string(' ', -1);`

#### Version - Were you using the User Kit or Dev Kit? And what User Kit version or Dev Kit commit (Cosmos, IL2CPU, X#)?
Latest devkit

jvyden:
Closes #2717

### DIFF
--- a/source/Cosmos.Core_Plugs/System/StringImpl.cs
+++ b/source/Cosmos.Core_Plugs/System/StringImpl.cs
@@ -36,6 +36,11 @@ namespace Cosmos.Core_Plugs.System
             [FieldAccess(Name = "System.Int32 System.String._stringLength")] ref int aStringLength,
             [FieldAccess(Name = "System.Char System.String._firstChar")] char* aFirstChar)
         {
+            if (length < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(length), "Cannot initialize a string with a negative length");
+            }
+
             aStringEmpty = "";
             aStringLength = length;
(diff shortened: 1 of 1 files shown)


zarlo:
when length is 0 
we should return String.Empty
like it is in .net https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/String.cs

quajak:
> when length is 0

does this have any actual difference execution wise? "" will be replaced by loading the same constant string each time

quajak (approved):
Thank you for the fix!

Will wait for Zarlo to get back to me before merging

zarlo:
> > when length is 0
> 
> does this have any actual difference execution wise? "" will be replaced by loading the same constant string each time

nope it will just use more memory

jvyden:
Agreed, but I'm not quite sure how to implement this. We don't return strings directly like it's done in mscorlib.

quajak:
I don't think it actually does because we load a constant string, which does not actually use memory on the heap.

PR #2727, opened 2023-07-16
Add more options for global init

MEMESCOEP:
I've added the following options to `HAL2.Global.Init`:
* InitSerial
* InitPCI
* InitACPI
* Init AHCI
<br/>
This gives more control to the developer because they can enable/disable certain drivers and initialization steps, which may be helpful when certain drivers/devices/features aren't needed, or the developer is writing their own.


MishaProductions:
I think that the better approach would be to move everything from Global.Init to the project template.

MEMESCOEP:
> I think that the better approach would be to move everything from Global.Init to the project template.

That could work, but then we'd have a bunch of checkboxes everywhere. Unless we find a different way of course

MEMESCOEP:
> I meant to move it to Kernel.BeforeRun()

Oh alright

quajak:
I think the idea is that we provide a initialise function, which anyone can then implement themselves if they want to change it. Rather than using this flag based approach.

MEMESCOEP:
> I think the idea is that we provide a initialise function, which anyone can then implement themselves if they want to change it. Rather than using this flag based approach.

how would we implement this though?

quajak:
> how would we implement this though? Maybe a function that can be overridden?

Yeah, that is what I was thinking

Kiirx:
PS/2 driver, ac97 (dont really remember the name)... you get the point

Kiirx:
Or we can make it a List and then use foreach to execute all of them (so the type would be Action) https://learn.microsoft.com/en-us/dotnet/api/system.action?view=net-7.0

zarlo:
> Would be better like that but please add so that we can do this with every part of cosmos, even the **vga screen** so that we can really just initialise what is only needed for the user

that can already be done you can override the virtual method in your Kernel (OnBoot, Start or GetTextScreen in the case of vga screen) or you can make a plug to override what cosmos does

Kiirx:
Well then ok we can close the pr then

quajak:
> that can already be done you can override the virtual method in your Kernel (OnBoot, Start or GetTextScreen in the case of vga screen) or you can make a plug to override what cosmos does

Then is there any reason to add more configuration options?

PR #2742, opened 2023-07-28
Update documentation

ptobuon:
Let me know if there are any requests / problems
Added some examples on VFS and fixed Installation docs


quajak (approved):
Thanks for updating the documentation!

PR #2748, opened 2023-08-11
code clearups

PratyushKing:
cosmos is beyond savable with its 10 million comments,... so i cleaned it a bit


Jspa2:
A tab is the same width as a space in VGA text mode.

PratyushKing:
i guess? but isn't this the normal console.write code?

AsertCreator:
by doing this, driver will just put tab into the memory. vga doesn't tabs, also font may glitch. thats why driver uses 4 spaces.

PR #2769, opened 2023-09-11
start opening up the netstack

zarlo:
i plan to open up the netstack to allow network drivers to made with out needing to edit cosmos source code and a few QOL things


source/Cosmos.System2/Network/IPv4/IPPacket.cs
```diff
@@ -38,7 +38,8 @@ internal static void IPv4Handler(byte[] packetData)
             ARPCache.Update(ipPacket.SourceIP, ipPacket.SourceMAC);
 
             if (NetworkStack.AddressMap.ContainsKey(ipPacket.DestinationIP.Hash) == true ||
-                ipPacket.DestinationIP.Parts[3] == 255)
+                ipPacket.DestinationIP.Parts[3] == 255 // this is wrong x.x.x.255 is not always broadcast
```
valentinbreiz, line 41: In which case?

zarlo: when subnets are not /24 in size

like with 172.16.0.0/12 172.16.0.255 is not a broadcast address

like with 192.168.0.0/16 192.168.0.255 is not a broadcast address

like with 172.16.0.0/24 172.16.0.255 is the broadcast address

like with 172.16.0.0/30 172.16.0.3 is the broadcast address

source/Cosmos.System2/Network/IPv4/Address.cs
```diff
@@ -195,5 +198,47 @@ public int CompareTo(object obj)
...
+
+            if (obj == null)
+            {
+                return false;
+            }
+
+            if (this == null)
+            {
+                return false;
+            }
```
EliaSulimanov, line 218: I think this can be single "if".

source/Cosmos.System2/Network/IPv4/IPPacket.cs
EliaSulimanov: I'm not sure why this check is even there, what does it gives us?

source/Cosmos.System2/Network/IPv4/TCP/Tcp.cs
```diff
@@ -141,6 +141,40 @@ public class TransmissionControlBlock
...
+
+        public static ushort DynamicPortStart = 49152;
+
+        private static Random dynamicPortStartRandom = new Random();
+
+        /// <summary>
+        /// gets a random port
+        /// </summary>
+        /// <param name="tries"></param>
+        /// <returns></returns>
```
EliaSulimanov, line 153: Random.Next can throw exception.
Maybe it's better to just take (the last used port +1)? you won't need tries parameter any more and it would improve the speed of the process + time complexity.

I also think this function should use mutex to prevent race conditions. In multithread system the CPU can do task switch on the last "if" and the kernel will give the same port to 2 different process at the same time.

source/Cosmos.System2/Network/IPv4/UDP/UdpClient.cs
```diff
@@ -9,6 +10,27 @@ namespace Cosmos.System.Network.IPv4.UDP
...
+        {
+            for (int i = 0; i < tries; i++)
+            {
+
+                var port = (ushort)dynamicPortStartRandom.Next(DynamicPortStart, ushort.MaxValue);
+                if (!clients.ContainsKey(port)) return port;
+
+            }
+
+            return 0;
```
EliaSulimanov, line 32: Same race condition as with tcp.

source/Cosmos.System2/Network/IPv4/TCP/Tcp.cs
zarlo: cosmos does not have at form of mutex right now

source/Cosmos.System2/Network/IPv4/IPPacket.cs
zarlo: its needed so we dont just drop broadcast packets

source/Cosmos.System2/Network/IPv4/Address.cs
zarlo: it could be but speed wise its the same

MishaProductions (approved):

Issue #2700, opened 2023-06-16
Memory Manager memory leaks


valentinbreiz:
Adding strings together causes memory leaks, RAM is growing more and more when looping in graphic mode.

Example: 
`canvas.DrawString("Used RAM  = " + GCImplementation.GetUsedRAM() + "B" , font, BlackColor, (int)x, (int)(y + font.Height));`

quajak:
I cannot reproduce the issue with the following code.

```
            Canvas canvas = FullScreenCanvas.GetFullScreenCanvas();
            while (true)
            {
                Heap.Collect();
                canvas.Clear(Color.White);
                canvas.DrawString("Used RAM  = " + GCImplementation.GetUsedRAM() + "B", PCScreenFont.Default, Color.Black, 10, 40);
                canvas.Display();
            }
```

@terminal-cs also seems to be having issues on linux. Are you using linux?

valentinbreiz:
I'm on Linux yes

valentinbreiz:
```

Save
New
Duplicate & Edit
Just Text
using Cosmos.Core.Memory;
using Cosmos.Core;
using System;
using System.Collections.Generic;
using System.Text;
using Sys = Cosmos.System;
using Cosmos.Debug.Kernel;
using Cosmos.System.Graphics.Fonts;
using Cosmos.System.Graphics;
using System.Drawing;

namespace CosmosKernel2
{
    public class Kernel : Sys.Kernel
    {

        protected override void BeforeRun()
        {
            Canvas canvas = FullScreenCanvas.GetFullScreenCanvas();
            while (true)
            {
                var count = Heap.Collect();
                canvas.Clear(Color.White);
                canvas.DrawString("Available RAM                = " + GCImplementation.GetAvailableRAM() + "MB", PCScreenFont.Default, Color.Black, 10, 50);
                canvas.DrawString("Used RAM                     = " + GCImplementation.GetUsedRAM() + "B", PCScreenFont.Default, Color.Black, 10, 60);
                canvas.DrawString("Small Allocated Object Count = " + HeapSmall.GetAllocatedObjectCount(), PCScreenFont.Default, Color.Black, 10, 70);
                canvas.DrawString("Small Page Count             = " + RAT.GetPageCount((byte)RAT.PageType.HeapSmall), PCScreenFont.Default, Color.Black, 10, 80);
[...]

PR #2770, opened 2023-09-12
Fix issue when SMT Table expands to larger than one page

quajak:


quajak:
Fixes https://github.com/CosmosOS/Cosmos/issues/2700

valentinbreiz (approved):

PR #2774, opened 2023-09-17
Update Binutils to version 2.41 from cygwin

MishaProductions:
The version of binutils cosmos currently uses is from 2005 (v2.16.1), which needs to be updated to be able to use a linker script. Compiled in cygwin using following commands:
```
wget https://ftp.gnu.org/gnu/binutils/binutils-2.41.tar.xz
cd binutils-2.41
mkdir build; cd build;
../configure --target=i386-elf
make -j16; make install
```


MishaProductions:
Ready for review

zarlo:
looks good

quajak (approved):

PR #2792, opened 2023-10-11
Improove the readme a bit

9xbt:


PratyushKing:
but i would still refer as VFS than FS so change that

9xbt:
> but i would still refer as VFS than FS so change that

its not really meant to run on real hardware yet because of how buggy it is so its gonna stay

PR #2793, opened 2023-10-13
Add ability to automatically trigger Heap.Collect

quajak:


zarlo:
it would be nice if you could disable this like have it so you can oped out

PratyushKing:
what i would suggest is to use a bool variable to toggle auto-collect just in case the user wants to modify that. what do you think @quajak

quajak:
The way it is configured in the code, it is disabled by default and must be explicitly enabled.

MishaProductions (approved):

PR #2803, opened 2023-10-22
Add HeapSmall.PruneSMT

quajak:
Add new HeapSmall.PruneSMT method with tests and documentation
This allows the user to prune useless pages from the SMT, if it has grown too much

### DIFF
--- a/source/Cosmos.Core/Memory/RAT.cs
+++ b/source/Cosmos.Core/Memory/RAT.cs
@@ -297,6 +297,11 @@ namespace Cosmos.Core.Memory
             throw new Exception("Page type not found. Likely RAT is rotten.");
         }
 
+        /// <summary>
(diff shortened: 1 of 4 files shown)


source/Cosmos.Core/Memory/RAT.cs
```diff
@@ -297,6 +297,11 @@ public static uint GetFirstRATIndex(void* aPtr)
             throw new Exception("Page type not found. Likely RAT is rotten.");
         }
 
+        /// <summary>
+        /// Get the pointer pointing to the start of the position to which the current pointer is pointing
```
zarlo, line 301: if im understanding this right, would this not be better "get the pointer of the start of the page of this pointer"

quajak: Updated it

zarlo (approved):

PR #2808, opened 2023-10-30
DHCP improvements + TCP fix

valentinbreiz:


quajak:
Sounds good to me. A bit unfortunate but disabling the network tests seems reasonable.

MishaProductions (approved):

Issue #2779, opened 2023-09-25
Allow to skip Visual Studio Extensions in devkit installer


valentinbreiz:

Kiirx:
It would also be great to not use Visual Studio's msbuild, removing completely the requirement

Issue #1626, opened 2021-01-07
Visual Studio Workloads not found. Install .NET Core cross-platform development, Visual Studio Extension development


emanueleriboldi:
========================================
    Builder Log
========================================

========================================
    Checking Dependencies
========================================

Repos: IL2CPU, XSharp and Common is installed.
Visual Studio 15.9+ is installed.
Visual Studio Workloads not found. Install .NET Core cross-platform development, Visual Studio Extension development, 

========================================
    Error
========================================

Visual Studio Workloads not found.

But i installed .NET Core cross-platform development,.NET Core dev kit 
![Immagine 2021-01-07 075600](https://user-images.githubusercontent.com/63238570/103861589-f686c300-50bd-11eb-8257-c75d90272ded.png)

![Immagine 2021-01-07 075633](https://user-images.githubusercontent.com/63238570/103861593-f7b7f000-50bd-11eb-9575-3cc956cd8f55.png)

and Visual Studio Extension development!

quajak:
This seems to be connected to https://github.com/CosmosOS/Cosmos/issues/1603 in that, we as soon as multiple vs installations the check breaks. If you are installing the devkit, the easiest way would be to simply disable to check in the local installation, for an actual fix, the two solutions I can think of are 1) detect if there is more than 1 vs installation, disable check but warn user or 2) always warn the user if the check did not succeed but let the installation continue, even if they arnt installed, as it will break at some later point.

jp2masa:
Maybe the check for multiple VS instances is not working, it should show a dialog to select the target instance:

https://github.com/CosmosOS/Cosmos/blob/86dfa0879f882e6a9c09557ad080cb8a694d6961/source/Cosmos.Build.Builder/App.xaml.cs#L56-L76

PythonVbaUser:
> 
> 
> This seems to be connected to #1603 in that, we as soon as multiple vs installations the check breaks. If you are installing the devkit, the easiest way would be to simply disable to check in the local installation, for an actual fix, the two solutions I can think of are 1) detect if there is more than 1 vs installation, disable check but warn user or 2) always warn the user if the check did not succeed but let the installation continue, even if they arnt installed, as it will break at some later point.

Of course! I had 2 visual studio! Thanks very much, now is working!

quajak:
This issue should stay open until we fix the underlying issue. 

> Maybe the check for multiple VS instances is not working, it should show a dialog to select the target instance:
> 
[...]

PR #2826, opened 2023-11-28
Builder improvements

MishaProductions:
- Fix DPI support
 - Disable showing exit dialog when rebuild button is clicked
 - Updated nuget packages
 - Added option dialog (and removed command line options)
 - Fix "install install" string
 - Solves #2779 (and probably also solves #1603 #1626)


quajak (approved):
These are some nice improvements

PR #2827, opened 2023-12-03
[WIP] x64

MishaProductions:
This is still very far from booting (or compiling).
Changes:
 - Updated linker to support x64
 - Updated limine
 
 Depends on: https://github.com/CosmosOS/IL2CPU/pull/211

 Current state: Empty cosmos projects compile, but will not run as IL2CPU needs to be modified further

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ The following is a non-exhaustive list of features that Cosmos offers:
 - A basic network interface
 - A basic audio interface
 
-> **Note**
+> [!NOTE]
 > Use [embeded resources](https://cosmosos.github.io/articles/Kernel/ManifestResouceStream.html) instead of the VFS for now for assets.
 
 ## Setting it up


zarlo:
should also up the the default VM configs RAM to like 512 as we will be using a bit more RAM when using 64bit address

ADev531:
x64 Cosmos C# kernel, x86 Cosmos C# legacy kernel, x64 Cosmos C# UEFI Kernel.

if can, please make legacy kernel can convert to 64 kernel.

zarlo:
> x64 Cosmos C# kernel, x86 Cosmos C# legacy kernel, x64 Cosmos C# UEFI Kernel.
> 
> if can, please make legacy kernel can convert to 64 kernel.

with x64 we will be dropping x86 support so for the most part nothing will be needed to be done by some one using cosmos

ADev531:
will there is x86 kernel -> x64 kernel converter?

zarlo:
no as it will not likely be needed / lack of man power

ADev531:
any updates?

Kiirx:
This could be made as a draft

PR #2828, opened 2023-12-03
Add support for UEFI

MishaProductions:


PratyushKing:
"Latest News says that the world has changed since source/Cosmos.Build.Tasks/IL2CPU.cs line 79 has been changed from 'String' to 'string'."

MishaProductions:
I am not sure why that change even happened

valentinbreiz (approved):

PR #2829, opened 2023-12-04
Cosmos Linux Fix

PratyushKing:
To be short, this debug enabled thing will allow new projects to have debug screens disabled, which is actually helpful because on linux you have to do it manually before dotnet build to get the project to actually run so it will be really nice if this was pre-included!

### DIFF
--- a/source/templates/csharp/CSharpProject.csproj
+++ b/source/templates/csharp/CSharpProject.csproj
@@ -8,6 +8,7 @@
     </PropertyGroup>
 
     <PropertyGroup>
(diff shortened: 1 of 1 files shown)


MishaProductions (approved):

pleasenoban:
wouldnt this also disable debugging by default on windows?

PratyushKing:
@pleasenoban I think yes, but most people disable it anyways so its gonna be probably fine, anyways newbies dont need it.

pleasenoban:
it would be better if we use a conditional property group so we could get the best of both worlds

PR #2836, opened 2023-12-05
Cosmos on linux, improved better as per a request

PratyushKing:
As per a requet from @pleasenoban, I have made it such that the debug enabled is only disabled on a specific new template called `csharp-linux`, and yes I have also updated the docs for linux users who want to install the templates. If anyone has further queries please do tell.


valentinbreiz:
Shouldn't we just disable by default instead of adding the same code?

PratyushKing:
> Shouldn't we just disable debugging by default instead of adding the same code?

I did it but @pleasenoban saw that on windows it somehow made it worse so I did this. Don't worry I tested it in both OS

valentinbreiz:
> > Shouldn't we just disable debugging by default instead of adding the same code?
> 
> I did it but @pleasenoban saw that on windows it somehow made it worse so I did this. Don't worry I tested it in both OS

what do you think @quajak?

quajak:
> I did it but @pleasenoban saw that on windows it somehow made it worse so I did this. Don't worry I tested it in both OS

How is it worse? I think it's better if we disable it by default

Jspa2:
Adding a separate template for Linux is overcomplicating things. I think debugging should be switched off by default for all platforms.

valentinbreiz:
Agree

PR #2841, opened 2023-12-09
Improve disk IO speeds, optimize FAT Implementation memory allocation & speeds

GoldenretriverYT:
The current IO was heavily limited by the overhead added by IL2CPU.

My benchmark, which certainly shouldnt fall out this heavy due to sounding more like a write operation, creating as many files as possible within 30 seconds had following results:

Pre optimization: 9 files within 30 seconds created on Fat32 using File.Create
After: 73 files within 30 seconds created on Fat32 using File.Create

I also ran all the FAT Tests:
Pre: Took 157s
After: Took 45s

After added the fixes to reduce memory allocation, writing a 32kb file (using FileStream, not including the creation of the FileStream) had following results:
From 27586560 bytes to 102400 bytes. Not perfect, but certainly an improvement!


GoldenretriverYT:
Draft for now cause I will look into speeding up writes too

GoldenretriverYT:
image
Even a 1mb file took 77s before (under my testing after patching the memory leaks)

Samma2009 (approved):

valentinbreiz (approved):
Great optimizations thanks for this

quajak:
Awesome work!

PR #2843, opened 2023-12-10
Fix ISO9660 support

GoldenretriverYT:
- ISO9660 somehow got detected as FAT FS since UEFI support was added
- There is no block count on ATAPI devices (? not sure why that is), instead we just put a large number there. Its read only anyways.
- Xorriso somehow changed and output all file names as UPPERCASE. Adding -allow-lowercase made it go back to what it did before.


MishaProductions:
Looks good to me!

MishaProductions (approved):

Issue #2546, opened 2023-01-02
ACPI Tables overriden by Memory Manager


valentinbreiz:
Cause bochs errors, for some reasons works on VMWare

danyzimo:
can I ask you? I got an error called CS0234, would you mind helping me?

zarlo:
> can I ask you? I got an error called CS0234, would you mind helping me?

open a new issue for that

AsertCreator:
apparently, any compliable multiboot2 bootloader can provide rsdp address as tag. maybe we could parse that?

PR #2853, opened 2023-12-21
Revert Misha's commit for GCImpl

RaphMar2021:
The commit made by Misha 2 weeks ago broke the 'Cosmos.Core.Memory.Heap.Collect()' function. This commit aims to revert these changes

### DIFF
--- a/source/Cosmos.Core/GCImplementation.cs
+++ b/source/Cosmos.Core/GCImplementation.cs
@@ -84,6 +84,13 @@ namespace Cosmos.Core
             {
                 memPtr = (byte*)largestBlock->Address;
                 memLength = largestBlock->Length;
+                if ((uint)memPtr < CPU.GetEndOfKernel() + 1024)
+                {
+                    memPtr = (byte*)CPU.GetEndOfKernel() + 1024;
+                    memPtr += RAT.PageSize - (uint)memPtr % RAT.PageSize;
+                    memLength = largestBlock->Length - ((uint)memPtr - (uint)largestBlock->Address);
+                    memLength += RAT.PageSize - memLength % RAT.PageSize;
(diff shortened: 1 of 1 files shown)


valentinbreiz:
What did it broke? It seems to work for me

RaphMar2021:
Heap.Collect

RaphMar2021:
When i tried Mirage with the latest commit, it froze on the setup screen.

GoldenretriverYT:
Yeah there are some issues with Heap.Collect on the latest commit, but were you able to actually verify that it was that commit? Just because it changed the GC code doesn't meant it has to be that commit, so please make sure it was actually this commit.

RaphMar2021:
I reverted back the GCimpl code on my local version, and it works again

ADev531:
it, works for me and i can use gui os over than 10 minutes

9xbt:
don't blame misha, it's just that the gc is shit but i mean if you optimize your os which i really did with mango it will run just fine, no fps drop

RaphMar2021:
> don't blame misha, it's just that the gc is shit but i mean if you optimize your os which i really did with mango it will run just fine, no fps drop

I didn't blame Misha, they are doing really good work, but Heap.Collect now crashes the OS

RaphMar2021:
I mean, everyone can break everything

quajak:
We might need to readd some code like the original one to prevent the heap from being on top of kernel code. But I think we removed this code to fix https://github.com/CosmosOS/Cosmos/issues/2546. Therefore, we should try find a solution which fixes both issues, rather than just reintroduce the old one

pleasenoban:
i dont think this stops it from overriding the acpi tables though?

PR #2861, opened 2023-12-28
Updates on int.ToString(`format`)

Guillermo-Santos:
Improved Hexadecimal format code as it was changing the value of the original variable (tested only on [SharpLab](https://sharplab.io/), since I did the change before testing on Cosmos, but I assume that if it was changing on normal C# context, then it should have been changing on Cosmos too), also added support for the decimal format and padding.

### DIFF
--- a/Tests/Kernels/Cosmos.Compiler.Tests.Bcl.System/System/Int32Test.cs
+++ b/Tests/Kernels/Cosmos.Compiler.Tests.Bcl.System/System/Int32Test.cs
@@ -10,7 +10,7 @@ namespace Cosmos.Compiler.Tests.Bcl.System
         public static void Execute()
         {
             bool efuse;
-            int value, tmp;
+            int value;
             string result;
             string expectedResult;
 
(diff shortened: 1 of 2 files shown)


Guillermo-Santos:
Code used to do the test:
``` csharp
using System;

var a = 255;
Console.WriteLine(ToString(ref a,"X"));
Console.WriteLine(a);
Console.WriteLine(ToString2(ref a,"X"));
Console.WriteLine(a);

/* Changed Version */
static string ToString(ref int aThis, string format)
{
    if (string.IsNullOrEmpty(format))
    {
        return "0";
    }

    string result = aThis > 0 ? string.Empty : "0";

    switch (format[0])
    {
        case 'X':
            int value = aThis;

            while (value != 0)
            {
                int remainder = value % 16;
                if (remainder < 10)
                {
                    result = remainder + result;
                }
                else
                {
                    char temp = (char)('A' + (remainder - 10));
                    result = temp + result;
                }
                
                value /= 16;
            }
            break;
        case 'D':
            result = aThis.ToString();
            break;
        default:
            return aThis.ToString();
    }

    if (format.Length > 1)
    {
        return int.TryParse(format.AsSpan(1), out int number) ? result.PadLeft(number, '0') : aThis.ToString();
    }
    else
    {
        return result;
    }
}

// Original
static string ToString2(ref int aThis, string format)
{
    if (format.Equals("X"))
    {
        string result = "";
        if(aThis == 0)
        {
            result = "0";
        }
        while (aThis != 0)
        {
            if (aThis % 16 < 10)
            {
                result = aThis % 16 + result;
            }
            else
            {
                string temp = "";
                switch (aThis % 16)
                {
                    case 10: temp = "A"; break;
                    case 11: temp = "B"; break;
                    case 12: temp = "C"; break;
                    case 13: temp = "D"; break;
                    case 14: temp = "E"; break;
                    case 15: temp = "F"; break;
                }
                result = temp + result;
            }
            aThis /= 16;
        }
        return result;
    }
    else
    {
        return aThis.ToString();
    }
}
```

zarlo:
since this is been worked on right now 

could you make result in to a string builder so less strings are made

Guillermo-Santos:
> since this is been worked on right now
> 
> could you make result in to a string builder so less strings are made

Done

Guillermo-Santos:
What does this have to do with audio?

quajak:
Seems like there was an issue building the iso. This might be a random error. Could you please add some test cases to Cosmos to ensure that the new version now handles the broken cases correctly?

Guillermo-Santos:
> Seems like there was an issue building the iso. This might be a random error. Could you please add some test cases to Cosmos to ensure that the new version now handles the broken cases correctly?

Should I create a new project or do I add it to IO test?

quajak:
Please add them to the existing Int32 tests in the BclSystem test kernel

Guillermo-Santos:
Done

quajak (changes requested):
Thanks for adding the test. Sorry for the miscommunication, we only need tests to ensure that the new version works :) not that the old one does not work

Guillermo-Santos:
Done

quajak (approved):

PR #2864, opened 2023-12-29
Fix a bug

Kayi-KuzinAlexandr25456:
Mixed color channels on loading Bitmap from BMP file


quajak:
Can this not be fixed by specifying the color order?

PR #2867, opened 2023-12-30
Support scrolling properly

GoldenretriverYT:
Scrolling wasnt supported in the MouseManager and the PS/2 implementation didnt implement it properly either

### DIFF
--- a/source/Cosmos.HAL2/PS2Mouse.cs
+++ b/source/Cosmos.HAL2/PS2Mouse.cs
@@ -141,7 +141,7 @@ namespace Cosmos.HAL
             var xTest1 = mouseCycle == 2 && !HasScrollWheel;
             var xTest2 = mouseCycle == 4 && HasScrollWheel;
 
-            if (xTest1 || xTest2)
+            if ((mouseCycle == 2 && !HasScrollWheel) || (mouseCycle == 4 && HasScrollWheel))
             {
                 int xDeltaX = 0;
                 int xDeltaY = 0;
--- a/source/Cosmos.HAL2/PS2Mouse.cs
+++ b/source/Cosmos.HAL2/PS2Mouse.cs
@@ -131,7 +131,9 @@ namespace Cosmos.HAL
                 {
                     mouseCycle++;
                 }
-            }else if (mouseCycle == 3) {
+            }
+            else if (mouseCycle == 3) 
+            {
                 mouseByte[3] = IOPort.Read8(Core.IOGroup.PS2Controller.Data);
                 mouseCycle++;
             }
--- a/source/Cosmos.HAL2/PS2Mouse.cs
+++ b/source/Cosmos.HAL2/PS2Mouse.cs
@@ -36,7 +36,7 @@ namespace Cosmos.HAL
 
         #region Properties
 
-        private bool HasScrollWheel => mouseID == 3 || mouseID == 4;
+        public bool HasScrollWheel => mouseID == 3 || mouseID == 4;
 
         public byte PS2Port { get; }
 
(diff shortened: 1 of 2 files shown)


source/Cosmos.HAL2/PS2Mouse.cs
```diff
@@ -166,8 +169,15 @@ public void HandleMouse(ref INTs.IRQContext context)
...
-                    xScrollWheel = (xScrollWheelByte & 0b1000) == 0 ? xScrollWheelByte : xScrollWheelByte | ~0x0F;
+                    var xScrollWheelByte = mouseByte[3];
+
+                    xScrollWheel = xScrollWheelByte & 0b0000_0111;
+
+                    // If the 4th bit is set, the scroll wheel is negative
+                    if ((xScrollWheelByte & (1 << 4)) == 1 << 4) {
+                        // if we just negate the value it starts at -8 and goes to 0, but we want it to start at 0 and go to -8
+                        xScrollWheel = -(-xScrollWheel + 8);
+                    }
```
jp2masa, line 180: This doesn't look right. The 4th bit can be the 4th button state, in 5 buttons mode ([OSDev ref](https://wiki.osdev.org/PS/2_Mouse#5_buttons)). Also, you wrote a very similar logic in a more complicated way, this is just a sign extension.

GoldenretriverYT: You are right, I tried a bunch of things until I found out that it just didnt read the 4th byte at all, I am gonna revert that change.

source/Cosmos.HAL2/PS2Mouse.cs
```diff
@@ -131,12 +131,15 @@ public void HandleMouse(ref INTs.IRQContext context)
...
+                mouseCycle++;
             }
 
             // TODO: move conditions to the if statement when stack corruption detection
             //       works better for complex conditions
             var xTest1 = mouseCycle == 2 && !HasScrollWheel;
-            var xTest2 = mouseCycle == 3 && HasScrollWheel;
+            var xTest2 = mouseCycle == 4 && HasScrollWheel;
 
             if (xTest1 || xTest2)
```
quajak, line 144: We can get rid of these helper bools
```suggestion
            if ((mouseCycle == 2 && !HasScrollWheel) || (mouseCycle == 4 && HasScrollWheel))
```

source/Cosmos.HAL2/PS2Mouse.cs
```diff
@@ -131,12 +131,15 @@ public void HandleMouse(ref INTs.IRQContext context)
                 {
                     mouseCycle++;
                 }
+            }else if (mouseCycle == 3) {
```
quajak, line 134:
```suggestion
            }
            else if (mouseCycle == 3) 
            {
```

source/Cosmos.HAL2/PS2Mouse.cs
```diff
@@ -132,13 +132,18 @@ public void HandleMouse(ref INTs.IRQContext context)
...
             }
+            else if (mouseCycle == 3) 
+            {
+                mouseByte[3] = IOPort.Read8(Core.IOGroup.PS2Controller.Data);
+                mouseCycle++;
+            }
 
             // TODO: move conditions to the if statement when stack corruption detection
             //       works better for complex conditions
             var xTest1 = mouseCycle == 2 && !HasScrollWheel;
```
quajak, line 143: Please remove these two lines

quajak (changes requested):

quajak (approved):
Thanks!

PR #2891, opened 2024-01-13
Update Console Plugs

Guillermo-Santos:
### Plugged Properties

- Out: Standard Output TextWriter.
- Error: Standard Error TextWriter.
- In: StandardInput TextReader.
- IsInputRedirected: Flag to check if 'In' does not point to the System Standard Input.
- IsOutputRedirected: Flag to check if 'Out' does not point to the System Standard Output.
- IsErrorRedirected: Flag to check if 'Error' does not point to the System Standard Error.

### Plugged Methods

- OpenStandardInput, Output and Error (and it's variants).
- SetIn: to set the 'In' property.
- SetOut: to set the 'Out' property.
- SetError: to set the 'Error' property.
Each of these three changes the value of its respective 'Is*Redirecded' flag.

### Other changes

- Support for the 'Delete' key, now it remove the character infront of the cursor when pressed.
- A 'Tab' change it's size from 1 to 4 spaces depending of cursor position.
- Cursor position now is 'Cached', this to make Console Writes only change it's position at the end of the operation.
- The console write now checks for '\0' chars since StreamWriters use a byte[1024] array by default to write to console (even when input is less than that). This resulted on long blank writes when using encodings different from ASCII and UTF8 (really strange). This behavior was brought into light by this implementation.
- Removed a lot of the Console Plugs since they were not needed anymore.

### Bugs Found

- When user input a tab, it is not writhed (was this like that before?).


valentinbreiz:
Really cool changes! Could you make tests for this?

Guillermo-Santos:
> Really cool changes! Could you make tests for this?

I'll make them. Also, what do you think about making the 'Console class' and interface or BaseClass so that people could change the behavior of it? (Like changing the methods that decide what things are the standard input/output/error). So that, for example, you could make that the standard error is a file stream pointing to some file ln the FS.

Guillermo-Santos:
Or simply make those methods static and use a make static Func<TextWriter/Reader, bool> properties for each so you can change them.

source/Cosmos.System2/Console.cs
```diff
@@ -239,6 +367,84 @@ public int CursorSize
             }
         }
 
+        public Stream OpenStandardInput()
+        {
+            return new CosmosConsoleStream(FileAccess.Read);
```
zarlo, line 372: can we use the same instance and not make a new one each time this is called? same for the other 2 methods

source/Cosmos.System2/Global.cs
```diff
@@ -61,10 +61,7 @@ public static void Init(TextScreenBase textScreen, bool initScrollWheel = true,
             // We must init Console before calling Inits.
             // This is part of the "minimal" boot to allow output.
             Debugger.Send("Creating the global console...");
-            if (textScreen != null)
```
zarlo, line 64: the reason for this was so the dev couls pass their own backend for the console (it has been "broken" for some time but)

zarlo (changes requested):

source/Cosmos.System2/Console.cs
Guillermo-Santos: Could be make, since Error and oytput are a stream with only write access, should they both be the same stream or different ones?

zarlo: i was thinking of something like this, unless we want a new stream for each time its called
```cs
Stream? stdOut;
public Stream OpenStandardInput()
{
    if(stdOut == null)
        stdOut = new CosmosConsoleStream(FileAccess.Read);
    return stdOut;
}
```

Guillermo-Santos: Can also be:
```csharp
Stream? stdOut;
public Stream OpenStandardInput()
{
    return stdOut ?? = new CosmosConsoleStream(FileAccess.Read);
}
```

it really does not matter to much if it is the same string or a completly new one, if you see the code of the `Cosmos.Console.cs` you will see that all it does is just point directly to de Cosmos.System.Console methods. Different instances would be a case if we want it to work as a normal Stream(Do not really write content till you call the Flush method), but since the content is not saved on a buffer it really does not matter.

zarlo: o if it does not matter then i would say reuse the same one better for GC

Guillermo-Santos: Will do

Guillermo-Santos: > Could be make, since Error and output are a stream with only write access, should they both be the same stream or different ones?

Wich one do you think is a better choice? the same one to have less objects?

zarlo: i think not as they are not the same so 2 instances

Guillermo-Santos: Done

Guillermo-Santos: Now i will look into why the `Console.WriteLine(format, string)` is not working. I think it have to do something with how StreamWriter get's a StringBuilder object, but i do not know. maybe there is a method to plug.

valentinbreiz (approved):

zarlo (approved):

PR #2897, opened 2024-01-14
🩹 MakeFile various tweaks

Kiirx:
- IL2CPU was using Release config which broke the setup, so now every project will use Debug config.
- Added build time
- Added some coloring for easier reading because of the huge amount of warnings

image


Kiirx:
it seems like the github actions aren't using the makefile, which is probably why the config problem passed through the test, the github action should be moved to a linux one instead to test both of the platforms

Makefile
```diff
@@ -17,38 +17,47 @@ GIT = git
 DOTNET = dotnet
 
 GITFLAGS = clone --depth=1
-DOTNETFLAGS = -v:q -nologo
+DOTNETFLAGS = -nologo -v:q -c:Debug
```
zarlo, line 20: i dont think we use debug on the windows install @valentinbreiz do you know?

Kiirx: This is for Linux anyways

zarlo: yes but we should do the same as windows when building

zarlo: for build something with cosmos yes but if the windows build uses release to build cosmos we should do that on linux too

Kiirx: We could make a setup for that, but for everything to be working for sure we will have to add this makefile to the github build action

Kiirx:
so?

Makefile
quajak: I think we use release mode on windows as well. We only use debug when compiling the kernel using cosmos

valentinbreiz (changes requested):
Good work @Kiirx but could you enable release mode again?

Kiirx:
@valentinbreiz could you please review this PR? As it is ready to merge

valentinbreiz (approved):

PR #2936, opened 2024-01-31
Make ATAPI a tiny bit faster

GoldenretriverYT:
Converted reading the IOPort in a for loop to an optimized asm loop. Made copying the liminewp.bmp (~1.3mb) from the ATAPI ISO CDROM to the hard drive go from 14s to 9s.

Obviously still not super fast, but hey, its an improvement. If we want to go even further the next step for both ATA and ATAPI is DMA probably


valentinbreiz (dismissed):

GoldenretriverYT:
@valentinbreiz lol i found another part of code that made it super slow, its now 1.7s instead of 14s, can you pls review again?

valentinbreiz (approved):

PR #2944, opened 2024-02-06
🐛 Fix linux build

valentinbreiz:
Fix the `cp: cannot stat '/home/runner/cosmos/Cosmos/../IL2CPU/artifacts/Debug/nupkg/*.nupkg': No such file or directory` problem

But we still have https://github.com/CosmosOS/CosmosFtp/issues/5


MishaProductions (approved):

Kiirx:
bruh I did that and its still on hold and he does the same thing and its directly merged...

zarlo:
@valentinbreiz  i think we should set the output paths when we build/publish it should make it so this issue does not happen again

PR #2959, opened 2024-02-21
Fixing a patch

PratyushKing:
There was a recent commit by @valentinbreiz, that commit changed the copy paths of IL2CPU artifacts from Debug to Release which is not working as of right now on any current linux device and this pull request will revert it back to when it worked, This makes sure that there are no errors during the general build process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,9 +109,9 @@ install:
 	@mkdir -p $(DESTDIR)/Build/VMware/Workstation
 	@mkdir -p $(DESTDIR)/Packages
 	@mkdir -p $(DESTDIR)/Kernel
-	@cp -r $(IL2CPU_DIR)/artifacts/Release/nupkg/*.nupkg $(DESTDIR)/Packages/
-	@cp -r $(THISDIR)/artifacts/Release/nupkg/*.nupkg $(DESTDIR)/Packages/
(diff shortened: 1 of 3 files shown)


9xbt:
I would discard any changes to any `.lock.json` file before making the PR and I would also add an option to switch between debug and release instead of undoing the changes to a coredev

PratyushKing:
> I would discard any changes to any `.lock.json` file before making the PR and I would also add an option to switch between debug and release instead of undoing the changes to a coredev

Well, I wasn't sure about it either so I have enabled to allow edits according to coredevs so its entirely modifyable

Kiirx:
Already done in #2944 
I will finish this when I have time in the week

Kiirx:
consider this PR useless

Kiirx:
and also we don't use Debug but Release so it's even more useless

zarlo:
okay this is some thing that keeps breaking i thing we should just set the out put paths just to be done with it

PratyushKing:
> consider this PR useless also you should continue your minecraft gameplay it was so interesting and also the python tutorial damn your voice makes me wanna eat spicy strawberries

there was no voice.....

PratyushKing:
also yeah ill just keep it closed until theres something fixed, i anyways just update my own repo for changes and use fork

PR #2973, opened 2024-03-02
Fix IL2CPU docs

### LINKED ISSUE #2969: IL2CPU Documentation references Syslinux when Limine is being used
#### URL of Documentation - What page of documentation are we dealing with?
https://cosmosos.github.io/articles/Compiler/il2cpu.html

#### Expected Behaviour - What do you think should happen?
The IL2CPU documentation states that Cosmos uses Syslinux

#### Actual Behaviour - What is the issue?
The IL2CPU documentation should be updated to say Limine

9xbt:
Fixes #2969


PratyushKing:
Hold up, doesnt cosmos still use limine

quajak (approved):

PR #2975, opened 2024-03-02
Add extra args to Global.Init()

9xbt:
I had to do this because I need to initialize just PCI and then I have to initialize something in my kernel and then I can initialize the rest


9xbt:
Oh crap I messed it up, forgot to update some files

PR #2988, opened 2024-03-13
Add UEFI toggle to project properties

MEMESCOEP:
Users can choose if they want to compile their OS with UEFI support.

### DIFF
--- a/source/Cosmos.Build.Tasks/MakeIso.cs
+++ b/source/Cosmos.Build.Tasks/MakeIso.cs
@@ -14,6 +14,9 @@ namespace Cosmos.Build.Tasks
         [Required]
         public string OutputFile { get; set; }
 
+        [Required]
+        public bool UseUEFI { get; set; }
+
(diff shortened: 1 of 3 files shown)


quajak (approved):
Thank you for the contribution!

9xbt:
Would it actually work? If it would it would be awesome

MEMESCOEP:
@9xbt I've tested it and it seems to work quite well

9xbt:
> @9xbt I've tested it and it seems to work quite well

Hmm does VGA text mode work? i dont think that does

MEMESCOEP:
> > @9xbt I've tested it and it seems to work quite well
> 
> Hmm does VGA text mode work? i dont think that does

All my commit does is apply Misha's UEFI changes if the checkbox is checked. If it isn't, it's essentially the same thing as compiling a project before that PR was merged

9xbt:
Ahhh okay

9xbt:
> If it doesn't work, then I've got some work to do

I thought it wouldn't work but i guess it should

PR #2997, opened 2024-04-10
Plugged `System.Activator.CreateInstance` methods for types with parameterless constructors.

Guillermo-Santos:
This PR add the plugs for `System.Activator.CreateInstance<T>()` and `System.Activator.CreateInstance(Type T, bool, bool)` two methods that are used to create generic types.

With  `System.Activator.CreateInstance<T>()` the following start to work on cosmos:
```csharp
// A generic method to get an instance of type T
public static T Get<T>() where T : class, new()
{
    return new T();
}
```

# Limitations

1. **Constructors may not be on the compilation result**:At the moment Constructors, as any other method, are not included on the compilation result if never called, this could cause stack corruption or null reference exceptions when calling one of these methods or when using its result.

To avoid this, you would need to add a `dummy` variable somewhere (`BeforeRun` recommended) to make IL2CPU include the ctor on the compilation.
```csharp
var dummy = new MyClass();
var myVar = Activator.CreateInstance<MyClass>();//This will work as ctor is included on the compilation.
```

2. **`ref struct` should not be supported:** If the type you are trying to instantiate is a `ref struct` these methods should throw a `NotSupportedException` (as far as I know, is because `ref struct` should never be boxed), but since, at the moment, we cannot check if a type is a `ref struct` (`System.Type.IsByRefLike` field not plugged) we just process the `ref struct` and return it as a normal struct or class.

### DIFF
--- a/source/Cosmos.Core_Plugs/System/ActivatorImpl.cs
+++ b/source/Cosmos.Core_Plugs/System/ActivatorImpl.cs
@@ -1,0 +1,81 @@
+﻿using System;
+using System.Runtime.CompilerServices;
+using Cosmos.Core;
+using IL2CPU.API.Attribs;
+using IL2CPU.API;
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+
+namespace Cosmos.Core_Plugs.System
+{
+    [Plug(typeof(global::System.Activator))]
+    public static class ActivatorImpl
+    {
+        public static T CreateInstance<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] T>()
+        {
+            return (T)Activator.CreateInstance(typeof(T))!;
+        }
+
+        // nonPublic is ignored since, at the moment, we cannot know if a ctor is public or not
+        // and we would get a Stack Corruption or Null reference exception if the ctor is not present either way.
+        public unsafe static object CreateInstance(CosmosRuntimeType ctr, bool nonPublic, bool wrapExceptions)
+        {
+            ArgumentNullException.ThrowIfNull(ctr, "Type");
+
+            // get the Type's VTable entry
+            var mType = VTablesImpl.mTypes[ctr.mTypeId];
+
+            // Calculate Object Size
+            uint dSize = 0;
+            if (ctr.IsValueType)
+            {
+                // For value types this property holds the correct size, so we can avoid the iteration
+                dSize = mType.Size;
+            }
+            else
+            {
+                // Calculate Object Actual Size.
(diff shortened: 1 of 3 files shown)


zarlo:
> ref struct should not be supported: If the type you are trying to instantiate is a ref struct these methods should throw a NotSupportedException (as far as I know, is because ref struct should never be boxed), but since, at the moment, we cannot check if a type is a ref struct (System.Type.IsByRefLike field not plugged) we just process the ref struct and return it as a normal struct or class.

@Guillermo-Santos if it has not already been done could you make a issue for this so we can track it

source/Cosmos.Core_Plugs/System/ActivatorImpl.cs
```diff
@@ -0,0 +1,81 @@
...
+    [Plug(typeof(global::System.Activator))]
+    public static class ActivatorImpl
+    {
+        public static T CreateInstance<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] T>()
+        {
+            return (T)Activator.CreateInstance(typeof(T))!;
+        }
+
+        // nonPublic is ignored since, at the moment, we cannot know if a ctor is public or not
+        // and we would get a Stack Corruption or Null reference exception if the ctor is not present either way.
```
zarlo, line 20: is there a way to prevent the Stack Corruption and just throw an exception (we should try and match what .net does whti what exception is thrown)

Guillermo-Santos: The problem is, we do not have method info, we do not know what method is a ctor and what method is not (and their parameters) at runtime. To make it all work as intended the Vtable should be reworked.

Guillermo-Santos:
> > ref struct should not be supported: If the type you are trying to instantiate is a ref struct these methods should throw a NotSupportedException (as far as I know, is because ref struct should never be boxed), but since, at the moment, we cannot check if a type is a ref struct (System.Type.IsByRefLike field not plugged) we just process the ref struct and return it as a normal struct or class.
> 
> @Guillermo-Santos if it has not already been done could you make a issue for this so we can track it

Will do

source/Cosmos.Core_Plugs/System/ActivatorImpl.cs
Guillermo-Santos: What i mean for stack corruption or null reference, is that a method other than the ctor was called or for some reason the null check of the ctor throws (happened while testing but should not happend now). This could happend either because there is no parameterless constructor or that ctor is not at address 0 (till now, it always have been at address 0, but the possibility of it changing is still there).

quajak (approved):
Great work!

PR #3039, opened 2024-07-10
fixed dotnet command in makefile

GreenData17:


Makefile
```diff
@@ -154,7 +154,7 @@ nuget-install:
 .PHONY: template-install
 template-install:
 	@printf "Installing ${GREEN}C# Template packages${DEFAULT}\n"
-	@-dotnet new uninstall $(THISDIR)/source/templates/csharp/
+	@-dotnet new --uninstall $(THISDIR)/source/templates/csharp/
```
MishaProductions, line 157: When running dotnet new --uninstall on my system, I get this: `Warning: use of 'dotnet new --uninstall' is deprecated. Use 'dotnet new uninstall' instead.`

selkij: @GreenData17 what version of dotnet are you using? You might have an older dotnet version where using only `uninstall` isn't possible

GreenData17: I am using the dotnet-sdk-6.0 from the AUR. (its the version mentioned in the documentation)

selkij: weird I am also using the same one, you should leave it as is, what error were you getting?

GreenData17: Those three:

Screenshot from 2024-07-10 00-50-48

After adding the double dashes, all of them disappeared.

GreenData17: Some more information for finding the problem: 
I use arch and install applications mostly via yay. 
To install the dotnet sdk I executed `yay -S dotnet-sdk-6.0`. 
When executing `dotnet --version` I get the version "6.0.131". 
Hope this helps.

benni347: For dotnet6 and older, `--uninstall` should be used according to the docs. https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-new#description The change to subcommands was introduced in .net 7.

GreenData17: Ok, so for users like me, who only have the sdk 6.0 installed the command should have `--uninstall`.
And the warning probably only appears for users who have newer versions or multiple sdk's installed.
@selkij how is it for you, do you have a newer version installed? and what do you think should happen with this problem?

selkij: I also use the net6.0 sdk

PR #3050, opened 2024-07-25
Update Canvas.cs (Fixed the AlphaBlend function.)

Epiczhul:
I fixed the AlphaBlend function because it was wrong.
please merge.


aomine001 (commented):
Bitwise operator shift right by 8 does the same thing as division by 256 but you used 255 here which would break the AlphaBlend function.

PR #3051, opened 2024-07-28
Plug Random.NextBytes(Span<byte> buffer)

Guillermo-Santos:


Guillermo-Santos:
Since this is basically a copy of `Random.NextBytes(bytes[] buffer)` i wonder if i should also change that one to make a call to this one.

zarlo (approved):

zarlo:
> Since this is basically a copy of `Random.NextBytes(bytes[] buffer)` i wonder if i should also change that one to make a call to this one.

if you want to do this i can wait to merge it in

PR #3065, opened 2024-08-11
Multicore implementation

Epiczhul:
We now finaly have MADT and it works! It detects all CPU cores.
image

### DIFF
--- a/source/Cosmos.Core/ACPI.cs
+++ b/source/Cosmos.Core/ACPI.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Runtime.InteropServices;
 
 namespace Cosmos.Core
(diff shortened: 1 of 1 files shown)


9xbt:
What the fuck is this. You removed the shutdown code and most of the code is ChatGPT generated

source/Cosmos.Core/ACPI.cs
```diff
@@ -188,6 +263,44 @@ static bool Check_RSD(uint address)
...
+                            // Here you can add this APIC ID to a list or process it as needed.
+                        }
+                        break;
+
+                    case MADTEntryType.IoApic:
+                        IoApicEntry* ioApic = (IoApicEntry*)entryPtr;
+                        Console.WriteLine($"Found IO APIC with ID: {ioApic->IoApicId} at address: {ioApic->IoApicAddress:X}");
+                        break;
+
+                        // Handle other MADT entries if needed
```
winksplorer, line 297: You said "I MADE MADT WORK"
But these types of comments are usually from ChatGPT. Not judging, I use ChatGPT code too sometimes, but it's still not your code, stop pretending it is.

Epiczhul: No I Added the Comments because Last Time I comitted something they hated me because I removed all comments

Epiczhul: Only the Comments are Copilot Not the Code

winksplorer: you didn't ask copilot to add comments, you asked it something like "add madt" and gave it the ACPI.cs. "// Handle other MADT entries if needed" does not come from an AI adding comments separately.

PR #3094, opened 2024-10-22
🎨 Add Alpha Color support to DrawFilledRectangle

cosmicdaman:
So I made a discovery that If you use DrawPoint for DrawFilledRectangle, it enables you to use a System.Drawing.Color with an alpha value less than 255. Thought I might share it.


9xbt (changes requested):
For the love of god, make DrawLine support colors with the alpha channel instead of using DrawPoint

MishaProductions (changes requested):
These changes will probably make the function significantly slower.

cosmicdaman:
> For the love of god, make DrawLine support colors with the alpha channel instead of using DrawPoint

on it

PR #3217, opened 2026-01-01
Fix formatting bug and remove aDescription in HandleException

cosmicorsmthidk:
HandleException previously had a formatting bug where aName and the exception's error code (which is for some reason the vector?) would print in the same row. And since the aDescription argument was usually the same as aName, I see no problem in just removing it.


cosmicorsmthidk:
Closing due to coredevs no longer merging gen 2 PRs

PR #3235, opened 2026-06-15
Update README.md

nikitaosx:
I added more comprehensive descriptions for various pointers, and replaced not needed pointers. The reason why I removed the "CPU/FPU accelerated math library" and "low level assembly access and pointer memory control" parts is because they are rarely used in Cosmos OSes as far as I know.


nikitaosx:
Is anyone here to merge this?

Demiomad:
I don't think anyone is going to merge this, **they're now working on Gen3**

clarence112:
oh
gen 3 is slopcoded
that's disappointing :/

PR #3239, opened 2026-08-04
Scope console present() to the changed rect, not the whole framebuffer

Arawn-Davies:
## Summary
- `KeyboardTextReader.ReadLine()` and `ConsolePlug.ReadKey()` (i.e. `Console.ReadLine()`/`Console.ReadKey()`) each call `Canvas.Display()` on every keystroke, which goes through `GopDriver.Swap()` and copies the **entire** framebuffer (`LinearFrameBuffer.Copy(lastbuffer)`, several MB at 1024x768x32) to real VRAM — even though only one character cell actually changed.
- Under a virtualized/emulated display device that per-keystroke full-buffer copy is slow enough that normal typing visibly tears on screen (confirmed the display's own drawing model is already correctly double-buffered — `GopDriver.lastbuffer` is a software back buffer, `Swap()` is the only point that touches the real VRAM — so the fix is scoping *that* present, not adding buffering).
- Adds `GraphicDevice.SwapRect(x,y,w,h)` / `Canvas.Display(x,y,w,h)`, backed by a new `MemoryBlock.CopyNonTemporal(offset, span)` for partial MMIO writes, and has `GopDriver.SwapRect` present only the given rect (row by row) instead of the whole buffer.
- `KernelConsole` gains `CursorRowRect` (covers the common no-wrap case — full row width, one character tall); `KeyboardTextReader.ReadLine()` and `ConsolePlug.ReadKey()` now present just that row per keystroke instead of the full screen.
- Existing full-screen `Display()` callers (e.g. a compositor doing one blit per frame) are untouched — this only adds a scoped variant, the default full-screen path still works exactly as before.

## Test plan
- [x] Reproduced with a minimal from-scratch kernel (`Console.Clear()` + `Console.WriteLine()` banner + a `Console.ReadLine()` echo loop, no application code) booted under QEMU (`-M q35 -vga std -display cocoa,zoom-to-fit=on`): idle text was stable, but typing produced visible full-screen tearing.
- [x] Applied this patch, rebuilt the same minimal kernel against the patched local packages, reproduced clean — no exceptions, and the tearing was gone under active typing.
- [ ] Would appreciate a check on real hardware / a non-virtualized GOP framebuffer, since the bug's root cause (a slow-relative-to-refresh full-buffer MMIO copy) is specific to virtualized display devices — this is unlikely to have been visible on real hardware where the copy is fast.

### DIFF
--- a/src/Cosmos.Kernel.Core/Memory/MemoryBlock.cs
+++ b/src/Cosmos.Kernel.Core/Memory/MemoryBlock.cs
@@ -267,6 +267,22 @@ public class MemoryBlock
     }
 
     /// <summary>
+    /// Copy a byte span into this memory block at the given offset. Same
+    /// non-temporal (cache-bypassing) write as <see cref="Copy(ManagedMemoryBlock)"/>,
+    /// but scoped to a sub-range — for presenting only a damaged rect of an MMIO
+    /// framebuffer instead of paying for a full-buffer copy on every small update.
+    /// </summary>
+    /// <param name="aByteOffset">Destination byte offset within this memory block.</param>
+    /// <param name="aData">Source bytes to write.</param>
+    public unsafe void CopyNonTemporal(int aByteOffset, ReadOnlySpan<byte> aData)
+    {
+        fixed (byte* src = aData)
+        {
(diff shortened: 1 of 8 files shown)


Arawn-Davies:
Not sure if reproducible, will reopen when testing on Windows

PR #3247, opened 2026-09-20
Added commands to disable loading message on Limine configuration

josue-beauregard:


valentinbreiz:
Hello, we are in early stages of developement so there is actually no point on hidding logs. Logs are good for debugging.

